PR #4, opened 2017-04-25
Backport patches from GNOME Shell 3.24 for GJS 1.48

mariospr:
Do this in preparation for the GJS upgrade, even before the rebase.

https://phabricator.endlessm.com/T16757


ptomato:
**CR pass**, looks like all the fixes that were required upstream!

PR #6, opened 2017-04-27
Update Debian's GSettings overrides for the Endless use case

mariospr:
We will ship our own list of favourite apps for Endless, and tweak
some other keys, so we repurpose this file from Debian for us.

https://phabricator.endlessm.com/T2212


mariospr:
This is no longer valid, as it has been superseeded by https://github.com/endlessm/gnome-shell/pull/94

PR #18, opened 2017-06-05
watermak: Protect against null background managers

GeorgesStavracas:
When running under a GDM session, Shell has no background
manager instances available, and the watermark code would
make it crash.

Fix that by checking if there are background managers.

https://phabricator.endlessm.com/T15839

### DIFF
--- a/js/ui/watermark.js
+++ b/js/ui/watermark.js
@@ -216,6 +216,9 @@ const WatermarkManager = new Lang.Class({
 });
 
 function _forEachBackgroundManager(func) {
-    Main.overview._bgManagers.forEach(func);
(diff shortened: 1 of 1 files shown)


js/ui/watermark.js
```diff
@@ -216,6 +216,9 @@ const WatermarkManager = new Lang.Class({
 });
 
 function _forEachBackgroundManager(func) {
-    Main.overview._bgManagers.forEach(func);
-    Main.layoutManager._bgManagers.forEach(func);
+    if (Main.overview._bgManagers != null)
```
cosimoc: You can omit the `!= null`, so that the code will not risk running when `Main.overview._bgManagers` is e.g. `undefined`.

PR #33, opened 2017-06-09
window-manager: implement Endless animations

GeorgesStavracas:
The new animations make windows minimize and unminimize from
the bottom of the screen, as well as making the Speedwagon
windows animate from the bottom of the screen when they're
mapped.

https://phabricator.endlessm.com/T17526


js/ui/windowManager.js
```diff
@@ -1909,26 +1884,54 @@ const WindowManager = new Lang.Class({
...
         switch (actor._windowType) {
         case Meta.WindowType.NORMAL:
-            actor.set_pivot_point(0.5, 1.0);
-            actor.scale_x = 0.01;
-            actor.scale_y = 0.05;
-            actor.opacity = 0;
-            actor.show();
-            this._mapping.push(actor);
+            // Speedwagon windows slide from the bottom, while regular
+            // windows just face in. Regular windows with a Speedwagon
```
smspillaz: 'windows just fade in'

PR #37, opened 2017-06-12
T17240

smspillaz:
Bring back coverage reporting.

In order to get this to work, we needed to carry forward patches to port the tests to jasmine and refactor some of the shell code so as to avoid cross-dependencies. Some of this work is a good candidate for upstreaming.

https://phabricator.endlessm.com/T17240


js/misc/messageListMarkup.js
```diff
@@ -0,0 +1,25 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
+
+const GLib = imports.gi.GLib;
+const Pango = imports.gi.Pango;
+
+function fixMarkup(text, allowMarkup) {
+    log("Should allow markup " + allowMarkup);
```
cosimoc: Remove debug leftover

js/misc/coreEnvironment.js
```diff
@@ -0,0 +1,21 @@
...
+function coreInit() {
+    window.log = _makeLoggingFunc(window.log);
+
+    window._ = Gettext.gettext;
+    window.C_ = Gettext.pgettext;
+    window.ngettext = Gettext.ngettext;
+    window.N_ = function(s) { return s; };
+
+    const Format = imports.format;
+    String.prototype.format = Format.format;
```
cosimoc: Should you not remove these from the non-core environment then?

js/ui/sessionMode.js
```diff
@@ -106,8 +106,8 @@ const _modes = {
         panel: {
             left: ['endlessButton', 'appIcons'],
             center: [],
-            right: ['a11y', 'keyboard', 'aggregateMenu', 'userMenu',
-                    'dateMenu', 'hotCorner']
+            right: ['dateMenu', 'a11y', 'keyboard', 'aggregateMenu', 'userMenu',
```
cosimoc: I don't think this change is right. The clock should stay where it currently is.

smspillaz: Yep, @rshuler confirmed that too. I'll update the test, in that case.

js/misc/coreEnvironment.js
smspillaz: I think they should be removed ... unless I missed them.

smspillaz:
Comments addressed. As for the test naming - I did try and revert that commit, but it looks like it breaks Jasmine. If I remember correctly, Jasmine required 'test' to be in the name of the file before it would execute it.

js/misc/messageListMarkup.js
cosimoc: Still should be removed

smspillaz:
> Still should be removed

Oof, I thought I removed that :( I'll remove it now.

smspillaz:
Upstream bug: https://bugzilla.gnome.org/show_bug.cgi?id=783738

smspillaz:
There's a bunch of comments on the upstream patchset which I'd like to fix and port back downstream before merging this, see the linked bug report.

mariospr:
I'm closing this for now, as it's been open since June and does not really apply anymore.

@smspillaz if you want to rebase it and work on it after the upgrade to 3.26, then it's fine but for now it's just cluttering the list of open PRs.

PR #38, opened 2017-06-12
Add integration with the old App Store 

mariospr:
We use the old App Store to add folders and websites to the desktop, so add the relevant component back.

https://phabricator.endlessm.com/T17236

### DIFF
--- a/js/ui/shellDBus.js
+++ b/js/ui/shellDBus.js
@@ -11,6 +11,7 @@ const Config = imports.misc.config;
 const ExtensionSystem = imports.ui.extensionSystem;
 const ExtensionDownloader = imports.ui.extensionDownloader;
 const ExtensionUtils = imports.misc.extensionUtils;
+const IconGridLayout = imports.ui.iconGridLayout;
 const Main = imports.ui.main;
 const Screenshot = imports.ui.screenshot;
 const ViewSelector = imports.ui.viewSelector;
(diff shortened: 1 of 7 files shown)


js/ui/windowManager.js
```diff
@@ -2136,6 +2136,12 @@ const WindowManager = new Lang.Class({
             }
 
             if (SideComponent.isSideComponentWindow(actor.meta_window) && this._showDesktopOnDestroyDone) {
+                if (!Main.appStore.appLaunched) {
+                    Main.overview.showApps();
+                }
```
joaquimrocha: No curly braces on single line bodies. Same above on line 66.

js/ui/shellDBus.js
```diff
@@ -16,6 +17,11 @@ const Main = imports.ui.main;
 const Screenshot = imports.ui.screenshot;
 const ViewSelector = imports.ui.viewSelector;
 
+/* Occurs when an application is installed, meaning added to the desktop's app
```
joaquimrocha: Maybe just saying "Occurs when an application is added to the app grid" is enough?

js/ui/windowManager.js
mariospr: Ooops! I removed all those from appStore.js, but I clearly forget about this place. I'll fix that

js/ui/shellDBus.js
mariospr: This is not my comment (port from older code), but I'm happy to change it

PR #40, opened 2017-06-13
T16975

smspillaz:
Essentially this wraps a D-Bus method call by forwarding all of
its in-arguments and back-proxying its return value or error
to the caller, with the side effect of showing a splash screen
for the application if the application wasn't open yet. We'll
use this in DiscoveryFeed to ensure that we show a splash screen
for discovery-feed applications when they are opened to a given
page from the feed itself.

https://phabricator.endlessm.com/T16975

### DIFF
--- a/js/ui/appActivation.js
+++ b/js/ui/appActivation.js
@@ -90,6 +90,48 @@ const AppActivationContext = new Lang.Class({
             this.showSplash();
     },
 
+    // Activate an app via a D-Bus call and keep track of its splash screen
+    // if we had to open a new window for it. Otherwise, if there were
+    // windows already open, just do the D-Bus call
+    activateViaDBusCall: function(busName, path, interfaceName, method, args, done) {
+        Gio.bus_get(Gio.BusType.SESSION, null, Lang.bind(this, function(source, result) {
+            let bus = null;
+            try {
+                bus = Gio.bus_get_finish(result);
+            } catch (error) {
+                done(error, null);
+                return;
+            }
+
+            let proxy = new Gio.DBusProxy({
+                g_bus_type: Gio.BusType.SESSION,
+                g_connection: bus,
+                g_default_timeout: GLib.MAXINT32,
+                g_flags: Gio.DBusProxyFlags.NONE,
+                g_interface_name: interfaceName,
+                g_name: busName,
+                g_object_path: path
+            });
+
+            let invoke = Lang.bind(this, function() {
+                proxy.call(method, args, Gio.DBusCallFlags.NONE, -1, null, function(source, result) {
+                    try {
+                        done(null, proxy.call_finish(result));
+                    } catch (error) {
+                        done(error, null);
+                    }
+                });
+            });
+
+            if (this._app.state == Shell.AppState.RUNNING) {
+                invoke();
+            } else {
+                this.showSplash();
+                invoke();
+            }
+        }));
+    },
+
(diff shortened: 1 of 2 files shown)


js/ui/shellDBus.js
```diff
@@ -592,26 +612,50 @@ const AppLauncher = new Lang.Class({
...
+            invocation.return_error_literal(Gio.IOErrorEnum,
+                                            Gio.IOErrorEnum.NOT_FOUND,
+                                            'Unable to launch app ' + appName + ': Not installed');
+            return;
+        }
+
+        activationContext.activateViaDBusCall(busName, path, interfaceName, method, args, function(error, result) {
+            if (error) {
+                logError(error);
+                invocation.return_error_literal(Gio.IOErrorEnum,
```
cosimoc: Can you simply return the error here instead of reconstructing it?

js/ui/shellDBus.js
```diff
@@ -592,24 +604,23 @@ const AppLauncher = new Lang.Class({
     },
 
     LaunchAsync: function(params, invocation) {
-        if (name == 'eos-app-store') {
+        let [appName, timestamp] = params;
+
+        if (appName == 'eos-app-store') {
             Main.appStore.show(timestamp, true);
+            Main.appStore.appLaunched = true;
```
smspillaz: My feeling here is that the old behaviour is wrong - there's no reason to touch Main.appStore.appLaunched if we're launching some other app. That said, I'll separate it out.

js/ui/appActivation.js
```diff
@@ -90,6 +90,48 @@ const AppActivationContext = new Lang.Class({
...
+            try {
+                bus = Gio.bus_get_finish(result);
+            } catch (error) {
+                done(error, null);
+                return;
+            }
+
+            let proxy = new Gio.DBusProxy({
+                g_bus_type: Gio.BusType.SESSION,
+                g_connection: bus,
```
smspillaz: Ah okay, that's handy. I'll look into doing that.

smspillaz: Damn, I had a try of this and it doesn't seem to work :( I've left the old code in place for now.

js/ui/shellDBus.js
smspillaz: I don't think so in the sense that we need to convert the error into a GIOError (and should probably wrap it too.

smspillaz:
Comments addressed - I also fixed a bug where we were incorrect wrapping return values where we didn't need to.

js/ui/shellDBus.js
cosimoc: It does not look like you separated it out?

js/ui/shellDBus.js
```diff
@@ -607,7 +615,6 @@ const AppLauncher = new Lang.Class({
 
         if (appName == 'eos-app-store') {
             Main.appStore.show(timestamp, true);
-            Main.appStore.appLaunched = true;
```
cosimoc: Oh, I see that you undo the change in this commit. Please squash it instead with the previous commit, so that none of the two changes this behavior.

smspillaz: Oops, yeah, this revert needed to go into the other commit. Fixed it now.

smspillaz:
I also added a third commit for what I understand to be the correct behaviour of appLaunched.

PR #41, opened 2017-06-13
endlessButton: Change hover label according to current view

GeorgesStavracas:
When viewing apps, the hover label should read "Show Desktop",
and while viewing the desktop, the hover label should read
"Show Apps".

https://phabricator.endlessm.com/T17520

### DIFF
--- a/js/ui/endlessButton.js
+++ b/js/ui/endlessButton.js
@@ -46,7 +46,7 @@ const EndlessButton = new Lang.Class({
             (event.type() == Clutter.EventType.TOUCH_BEGIN ||
              event.type() == Clutter.EventType.BUTTON_PRESS)) {
 
-            Main.overview.toggle();
+            Main.overview.toggleApps();
         }
 
         return Clutter.EVENT_PROPAGATE;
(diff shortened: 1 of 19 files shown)


js/ui/endlessButton.js
```diff
@@ -38,6 +39,14 @@ const EndlessButton = new Lang.Class({
             this._labelOffsetX = this._label.get_theme_node().get_length('-label-offset-x');
             this._labelOffsetY = this._label.get_theme_node().get_length('-label-offset-y');
         }));
+
+        Main.overview.connect('page-changed', Lang.bind(this, function() {
```
cosimoc: Instead of setting the label from two different places, can you split the body of this function into its own private method, and call it from both the constructor and as a `page-changed` signal callback?

Also, I believe that the overview may outlive the button, so perhaps it would be a good idea to disconnect this signal when the button is destroyed.

mariospr:
This should be rebase on top of master now instead of my WIP branch, which will be disappearing soon enough (as I'll be using specific tickets numbers now that T17230 is moving to QA Blocked)

PR #42, opened 2017-06-14
Implement a basic version of the EOS desktop

mariospr:
This includes basic desktop functionality (desktop search, icon grid, (upstream) folders, window picker), and is yet another step in the direction of having EOS implemented on top of GNOME Shell, but it's not yet a complete experience.

Some things missing are: correct styling of the icon grid and the search results panel, editable labels, drag'n'drop for icons and folders (to reorder things), desaturated version of the desktop (ClutterClone) to show underneath running applications, and probably something else I'm not recalling right now.

Also, this is different than EOS in that the icon grid uses a paginated icon grid instead of a scrolled window, which makes two dots (pagination indicators) to show up on the right when not possible to fit all the icons. 

https://phabricator.endlessm.com/T17230


js/ui/viewSelector.js
```diff
@@ -132,6 +135,17 @@ const ViewsDisplayLayout = new Lang.Class({
 
     _onStyleChanged: function() {
         this.layout_changed();
+    },
+
+    vfunc_allocate: function (actor, box, flags) {
```
cosimoc: Remove space before opening paren

js/ui/viewSelector.js
```diff
@@ -146,13 +160,28 @@ const ViewsDisplayContainer = new Lang.Class({
...
                       y_expand: true });
 
+        layoutManager.connect('allocated-size-changed', Lang.bind(this, this._onAllocatedSizeChanged));
+
         this.add_actor(this._appDisplay.actor);
     },
 
+    _onAllocatedSizeChanged: function(actor, width, height) {
+        let box = new Clutter.ActorBox();
+        box.x1 = box.y1 =0;
```
cosimoc: Add space before `0`

js/ui/viewSelector.js
```diff
@@ -346,6 +346,14 @@ const ViewSelector = new Lang.Class({
 
         let symbol = event.get_key_symbol();
 
+        if (this._activePage == this._workspacesPage) {
+            if (symbol == Clutter.Escape) {
+                Main.overview.toggleWindows();
+                return true;
+            }
+            return false;
```
cosimoc: Why not simply let code fall through in this case?

js/ui/viewSelector.js
```diff
@@ -346,6 +346,14 @@ const ViewSelector = new Lang.Class({
 
         let symbol = event.get_key_symbol();
 
+        if (this._activePage == this._workspacesPage) {
+            if (symbol == Clutter.Escape) {
+                Main.overview.toggleWindows();
+                return true;
```
cosimoc: Return `Clutter.EVENT_STOP` instead

js/ui/search.js
```diff
@@ -113,6 +138,12 @@ const ListSearchResult = new Lang.Class({
                                        x_align: St.Align.START,
                                        y_align: St.Align.END });
         }
+
+        let hoverIcon = new St.Icon({ style_class: 'list-search-result-arrow-icon',
+                                      icon_name: 'go-next-symbolic' });
+        content.add(hoverIcon,  { x_fill: false,
```
cosimoc: Extra space before params

js/ui/search.js
```diff
@@ -367,14 +398,58 @@ const GridSearchResults = new Lang.Class({
 });
 Signals.addSignalMethods(GridSearchResults.prototype);
 
+const SearchResultsBin = new Lang.Class({
+    Name: 'SearchResultsBin',
+    Extends: St.BoxLayout,
+
+    vfunc_get_preferred_height: function(forHeight) {
+        let themeNode = this.get_theme_node();
+        let marginBottom = themeNode.get_length('-margin-bottom');
```
cosimoc: Are regular margins not supported by the theme machinery these days?

js/ui/search.js
```diff
@@ -367,14 +398,58 @@ const GridSearchResults = new Lang.Class({
...
+        return [min + marginBottom, nat + marginBottom];
+    },
+
+    vfunc_get_preferred_width: function(forHeight) {
+        let themeNode = this.get_theme_node();
+        let maxWidth = themeNode.get_max_width();
+        let marginHorizontal = themeNode.get_length('-margin-horizontal');
+
+        let [min, nat] = this.parent(forHeight);
+        return [Math.min(maxWidth, min) + marginHorizontal,
```
cosimoc: If this worked like a regular margin, you'd want to consider it twice (one for each horizontal direction), but I hope that we can simply use margins and remove this custom code.

js/ui/search.js
```diff
@@ -69,11 +69,36 @@ const SearchResult = new Lang.Class({
 });
 Signals.addSignalMethods(SearchResult.prototype);
 
+const ListDescriptionBox = new Lang.Class({
+    Name: 'ListDescriptionBox',
+    Extends: St.BoxLayout,
+
+    _init: function(params) {
+        this.parent(params);
```
cosimoc: No need to have an `_init()` function if it only chains to parent

js/ui/viewSelector.js
```diff
@@ -355,6 +370,13 @@ const ViewsDisplay = new Lang.Class({
         this.actor = new ViewsDisplayContainer(this.entry, this._appDisplay, this._searchResults);
     },
 
+    _recordDesktopSearchMetric: function (query, searchProvider) {
```
cosimoc: Remove space before opening paren

js/ui/overview.js
```diff
@@ -555,8 +552,6 @@ const Overview = new Lang.Class({
                            onComplete: this._showDone,
                            onCompleteScope: this
                          });
-        this._shadeBackgrounds();
```
mariospr: Right. I think we can remove everything then

js/ui/dash.js
```diff
@@ -343,7 +343,8 @@ const DashActor = new Lang.Class({
         let layout = new Clutter.BoxLayout({ orientation: Clutter.Orientation.VERTICAL });
         this.parent({ name: 'dash',
                       layout_manager: layout,
-                      clip_to_allocation: true });
+                      clip_to_allocation: true,
+                      visible: false });
```
mariospr: Actually, I just gave it a try now and the change seems to be doable:
```
 js/perf/core.js           |   2 --
 js/perf/hwtest.js         |   1 -
 js/ui/appDisplay.js       |   6 +++---
 js/ui/dash.js             |   2 --
 js/ui/iconGrid.js         | 110 --------------------------------------------------------------------------------------------------------------
 js/ui/overview.js         |  12 ------------
 js/ui/overviewControls.js |  37 ++-----------------------------------
 js/ui/viewSelector.js     |  39 +++++----------------------------------
 8 files changed, 10 insertions(+), 199 deletions(-)
```
Maybe it's one of those things that looked messier and harder to me one month ago, out of pure ignorance. I'll keep addressing your comments and rebasing things based on this for now, see how it turns out...

js/ui/appDisplay.js
```diff
@@ -373,7 +373,7 @@ const AllView = new Lang.Class({
 
     _init: function() {
         this.parent({ usePagination: true }, null);
-        this._scrollView = new St.ScrollView({ style_class: 'all-apps',
+        this._scrollView = new St.ScrollView({ style_class: 'all-apps vfade',
```
mariospr: It's an upstream issue, just didn't have time to report it yet. From what I can see we also added ' vfade' in eos-desktop in the past, so it looks like it's been there for a while.

mariospr: I just mentioned this to fmuellner on IRC, to get his opinion. Perhaps it's not the real bug and just a sympton, but for now I'd keep it here, as it's consistent with what we already did so far too

mariospr: https://bugzilla.gnome.org/show_bug.cgi?id=783823

js/ui/overview.js
```diff
@@ -229,20 +229,8 @@ const Overview = new Lang.Class({
                                         opacity: 0 });
         this._overview.add_actor(this._panelGhost);
 
-        this._searchEntry = new St.Entry({ style_class: 'search-entry',
```
mariospr: That would be really nice, specially the upstream part. I'll keep it in mind for the next rebase, which I hope will be easier and leave more time for upstreaming and doing things in a better way.

js/ui/viewSelector.js
```diff
@@ -205,9 +205,9 @@ const ViewSelector = new Lang.Class({
         Main.overview.show();
     },
 
-    show: function() {
+    show: function(viewPage) {
```
mariospr: Yes

js/ui/overview.js
```diff
@@ -523,6 +526,41 @@ const Overview = new Lang.Class({
...
+            // TODO: Show a dialog with a 'No apps open' message to the user.
+            return;
+        }
+
+        if (this.viewSelector.getActivePage() !== ViewSelector.ViewPage.WINDOWS) {
+            this.showWindows();
+            return;
+        }
+
+        if (!this._toggleToHidden) {
```
mariospr: Good catch. That's a leftover from a previous wip branch. I'll remove it

js/ui/overview.js
```diff
@@ -523,6 +526,41 @@ const Overview = new Lang.Class({
...
+    toggleWindows: function() {
+        if (this.isDummy)
+            return;
+
+        if (!this.visible) {
+            this.showWindows();
+            return;
+        }
+
+        if (!Main.workspaceMonitor.hasActiveWindows) {
```
mariospr: I put this check here because every possible action that comes after this point relies on having active windows in the workspace:
  1. If you have active (hidden or not) windows and you're not in the windows picker => switch to it
  2. Otherwise, if you have active windows **and none** of them are hidden => switch to the apps grid
  3. In any other case (active windows, but all hidden) => hide the overview

All of 1-3 require active windows (hidden or not) to be available, so it made sense to me to add the check right before (1) to early return there, also providing the place to show the "No apps yet" message to the user.

js/ui/viewSelector.js
```diff
@@ -207,7 +207,7 @@ const ViewSelector = new Lang.Class({
 
     show: function(viewPage) {
         this._activePage = null;
-        this._showPage(this._pageFromViewPage(viewPage));
+        this._showPage(this._pageFromViewPage(viewPage), true);
```
mariospr: Another leftover. Will remove it.

js/ui/viewSelector.js
mariospr: I actually did in one of the latest commits, but forgot to squash it. Will do.

js/ui/viewSelector.js
mariospr: It's probably fine, but I liked to think of this block as specific to the workspace page, where the `!global.stage.key_focus` statement that comes afterwards does not make sense.

js/ui/workspace.js
```diff
@@ -1556,6 +1556,10 @@ const Workspace = new Lang.Class({
 
     _windowRemoved : function(metaWorkspace, metaWin) {
         this._doRemoveWindow(metaWin);
+
+        // Make sure to show the icon grid if we removed the last window
+        if (!Main.workspaceMonitor.hasActiveWindows)
+            Main.overview.showApps();
```
mariospr: I thought it would, but for some reason that does not work when you're in the "window picker" mode, since you apparently get the `destroy-completed` signal before the actual window has been removed, and `showApps()` never gets called, which is why I added this explicitly here.

I haven't investigated this deeply yet, but I'm guessing it might have something to do with the fact that when you close a window from the window picker you're not quite closing the real window at that time, but a clone of it. I can investigate it further if you want, but it looks like a safe workaround to me. What do you think?

js/ui/backgroundMenu.js
```diff
@@ -46,10 +46,12 @@ function _addBackgroundMenuFull(actor, clickAction, layoutManager) {
     if (!actor && !clickAction)
         return;
 
-    if (actor)
+    if (actor) {
         clickAction = new Clutter.ClickAction();
-    else
+        actor.add_action(clickAction);
```
mariospr: Oops!

js/ui/backgroundMenu.js
```diff
@@ -1,6 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 
 const Clutter = imports.gi.Clutter;
+const Gdk = imports.gi.Gdk;
```
mariospr: Leftover

js/ui/viewSelector.js
```diff
@@ -113,6 +119,84 @@ const ShowOverviewAction = new Lang.Class({
...
+        this._activePage = ViewsDisplayPage.APP_GRID;
+
+        this.parent({ layout_manager: new ViewsDisplayLayout(this._appDisplay.actor),
+                      x_expand: true,
+                      y_expand: true });
+
+        this.add_actor(this._appDisplay.actor);
+    },
+
+    showPage: function(page) {
```
mariospr: Actually I don't think we need to emit any signal, at least not for now. It was used in the old shell to call `updateBackgroundShade()` from the overview, but we're not doing that for now here, so the emission can be removed altogether.

js/ui/viewSelector.js
mariospr: Oops

js/ui/viewSelector.js
mariospr: Oops 2

js/ui/viewSelector.js
```diff
@@ -126,26 +128,116 @@ const ViewsDisplayLayout = new Lang.Class({
...
 
+        this._entry = entry;
         this._appDisplayActor = appDisplayActor;
+        this._searchResultsActor = searchResultsActor;
+
+        this._entry.connect('style-changed', Lang.bind(this, this._onStyleChanged));
         this._appDisplayActor.connect('style-changed', Lang.bind(this, this._onStyleChanged));
+
+        this._heightAboveEntry = 0;
+        this.searchResultsTween = 0;
```
mariospr: I honestly haven't thought much about it, this is pretty much carried over from eos-desktop :-)

I assume it's fine, though, knowing it's being working fine for a while?

js/ui/viewSelector.js
```diff
@@ -126,26 +128,116 @@ const ViewsDisplayLayout = new Lang.Class({
...
 
     _onStyleChanged: function() {
         this.layout_changed();
     },
 
-    vfunc_allocate: function (actor, box, flags) {
-        let availWidth = box.x2 - box.x1;
-        let availHeight = box.y2 - box.y1;
+    _centeredHeightAbove: function (height, availHeight) {
+        return Math.floor(Math.max((availHeight - height) / 2, 0));
```
mariospr: Sure.  Again, not my code but I agree with you. I already changed a few other things too anyway..

js/ui/viewSelector.js
```diff
@@ -199,9 +314,99 @@ const ViewsDisplay = new Lang.Class({
     Name: 'ViewsDisplay',
 
     _init: function() {
+        this._enterSearchTimeoutId = 0;
+        this._localSearchMetricTimeoutId = 0;
```
mariospr: Right, it belongs to the following commit

js/ui/viewSelector.js
```diff
@@ -199,9 +314,99 @@ const ViewsDisplay = new Lang.Class({
...
+        this._searchResults.actor.bind_property('mapped', clickAction, 'enabled', GObject.BindingFlags.SYNC_CREATE);
+
+        this.actor = new ViewsDisplayContainer(this.entry, this._appDisplay, this._searchResults);
+    },
+
+    _updateSpinner: function() {
+        this.entry.setSpinning(this._searchResults.searchInProgress);
+    },
+
+    _enterLocalSearch: function() {
```
mariospr: Agreed. I was already confused by what a "local" search was... I don't think that's defined here, but by the providers used anyway

js/ui/search.js
mariospr: I think this was part of some debugging... will remove it

js/ui/search.js
mariospr: Ok

js/ui/search.js
mariospr: Correct, will change this

js/ui/search.js
mariospr: Again, not my code. But I agree with you it's probably better to remove this if possible

js/ui/overview.js
mariospr: Actually, no... further testing while addressing your comments made me realize I actually need that, together with the remaining code not present here which I removed. Will add it back

js/ui/search.js
```diff
@@ -367,14 +394,35 @@ const GridSearchResults = new Lang.Class({
 });
 Signals.addSignalMethods(GridSearchResults.prototype);
 
+const SearchResultsBin = new Lang.Class({
```
mariospr: Note that we still need to use our own subclass of St.BoxLayout to override the allocate vfunc, since max-width is not a CSS property that clutter understands out-of-the-box, but something that's implemented at the level of the Shell Toolkit:
https://developer.gnome.org/st/stable/st-st-theme-node.html#st-theme-node-get-max-width

js/ui/viewSelector.js
rshuler: Looks like this was not changed per Cosimo's request.

js/ui/viewSelector.js
rshuler: Same here

js/ui/viewSelector.js
rshuler: And same here

js/ui/viewSelector.js
rshuler: Unless I'm confused, it looks like this same line is added on two commits. End result is fine, but perhaps the commit history is confusing.

js/ui/viewSelector.js
mariospr: I'm confused too. I definitely changed that, maybe I pushed the wrong commit

mariospr: Hmm... You're reviewing the outdated commit. The updated one in the current PR is https://github.com/endlessm/gnome-shell/pull/42/commits/126b4604e71d38254de263f11aef3706abfd3e05

rshuler: Oh, the end result is correct -- maybe it just showed up on a later commit.

mariospr: No, it's my fault, and github does not make it any easier: I squashed by mistake some of the changes that should go into that commit in the previous one, leaving just a small subset of the changes required to add the search entry in this commit. I'll fix that

js/ui/viewSelector.js
mariospr: Again, same problem mentioned before with some weird squashing. I'll fix that up

js/ui/viewSelector.js
mariospr: yep

rshuler: Yes, this also ended up getting addressed on a later commit.  Now, I see...  This commit and the next one are a bit mangled in the history.  I'll see if I can clean that up before merging.

mariospr: I already fixed that up, should be ok in the current PR

js/ui/viewSelector.js
```diff
@@ -200,13 +197,6 @@ const ViewSelector = new Lang.Class({
                 }
             }));
 
-        Main.wm.addKeybinding('toggle-application-view',
```
cosimoc: If I'm not mistaken, this is used for Super+A and would be nice to keep

js/ui/dash.js
cosimoc: Nice! I believe that since the actor is never created now, you should not need to make the dash invisible, or even need to remove the redisplay code below though.

js/ui/viewSelector.js
cosimoc: I guess we're left with a broken commit here then, but perhaps it's not such a big deal. Something to take care of with the next rebase.

js/ui/workspace.js
cosimoc: I am afraid that we'll then have two code paths trying to show the desktop at the same time when the last window closes. I am OK deferring this to a separate ticket to be investigated later though.

js/ui/viewSelector.js
cosimoc: I still think it's cleaner to make this a property, since the code in this and the following method is exactly what you'd expect to find in a property setter and getter.

js/ui/appDisplay.js
```diff
@@ -1665,7 +1665,9 @@ const AppIcon = new Lang.Class({
     },
 
     getDragActor: function() {
-        return this.app.create_icon_texture(Main.overview.dashIconSize);
+        // This is a temporary change not to depend on the dash while
+        // we don't implement our own Drag'n'Drop mechanism in EOS.
+        return this.app.create_icon_texture(64);
```
cosimoc: Can you use the constant in `iconGrid.js`?

js/ui/viewSelector.js
cosimoc: It sounds like it would be easy enough to try...
But if it's carried over from eos-desktop it's probably OK.

js/ui/viewSelector.js
```diff
@@ -195,9 +310,98 @@ const ViewsDisplay = new Lang.Class({
...
+        // We give a very short time for search results to populate before
+        // triggering the animation, unless an animation is already in progress
+        if (this._searchResults.isAnimating) {
+            this.actor.showPage(ViewsDisplayPage.SEARCH, true);
+            return;
+        }
+
+        this._enterSearchTimeoutId = Mainloop.timeout_add(SEARCH_ACTIVATION_TIMEOUT, Lang.bind(this, function () {
+            this._enterSearchTimeoutId = 0;
+            this.actor.showPage(ViewsDisplayPage.SEARCH, true);
```
cosimoc: `return false;` (or `GLib.SOURCE_REMOVE`) at the end of this

js/ui/search.js
```diff
@@ -367,14 +394,35 @@ const GridSearchResults = new Lang.Class({
...
 
     _init: function() {
-        this.actor = new St.BoxLayout({ name: 'searchResults',
-                                        vertical: true });
+        this.actor = new SearchResultsBin({ name: 'searchResults',
+                                            vertical: true });
+        Util.blockClickEventsOnActor(this.actor);
 
-        this._content = new St.BoxLayout({ name: 'searchResultsContent',
+        this._content = new St.BoxLayout({ name: 'searchResultsBox',
```
cosimoc: The name of this actor is changed, yet nothing in the theme reflects the new name. I believe the change is accidental and would not be surprised if it caused some of the issues mentioned in the commit message.

js/ui/appDisplay.js
mariospr: WHAT? I did this change yesterday and I'm pretty sure I pushed it... but you're right, it's not there, and I'm very confused...

mariospr: Ah, I see now... this might have been overriden by the force push Roddy did, even after I checked the last commits (but probably failed to compare the whole thing). Fortunately, that was the only change that got lost together with this other one I did following Roddy's review:
```
--- a/js/ui/overviewControls.js
+++ b/js/ui/overviewControls.js
@@ -452,7 +452,7 @@ const ControlsManager = new Lang.Class({
 
         geometry.width -= thumbnailsWidth;
 
-        if (this.actor.get_text_direction() != Clutter.TextDirection.LTR)
+        if (this.actor.get_text_direction() == Clutter.TextDirection.RTL)
             geometry.x += thumbnailsWidth;
 
         this.viewSelector.setWorkspacesFullGeometry(geometry);
```
I'll put them back

js/ui/dash.js
mariospr: Oops! You're right, I'll put that back with a note to be squashed with the original commit in future rebases

js/ui/viewSelector.js
mariospr: This seems like a mistake, I was not meaning to remove this

js/ui/viewSelector.js
```diff
@@ -200,13 +197,6 @@ const ViewSelector = new Lang.Class({
...
-        Main.wm.addKeybinding('toggle-application-view',
-                              new Gio.Settings({ schema_id: SHELL_KEYBINDINGS_SCHEMA }),
-                              Meta.KeyBindingFlags.NONE,
-                              Shell.ActionMode.NORMAL |
-                              Shell.ActionMode.OVERVIEW,
-                              Lang.bind(this, this._toggleAppsPage));
-
         Main.wm.addKeybinding('toggle-overview',
                               new Gio.Settings({ schema_id: SHELL_KEYBINDINGS_SCHEMA }),
                               Meta.KeyBindingFlags.NONE,
```
mariospr: This should call toggleWindows() too, I'll change both things as one fixup commit

js/ui/viewSelector.js
mariospr: Oops, saw your comment and didn't realize that would cause a broken commit :(

js/ui/workspace.js
mariospr: I'm not sure what was going on but all my experiments led to the same point where I could not get the event at the right time unless I did something explicitly like this. But I agree something funny is going on here, will try to get to it later, while doing general cleanups and stabilization.

js/ui/viewSelector.js
mariospr: Ok

js/ui/search.js
mariospr: There is no 'searchResultsContent' in the CSS either, so I doubt that's the issue. Still, this was a mistake, will fix it.

PR #45, opened 2017-06-16
Implement desktop's clones to show underneath visible windows 

mariospr:
This PR moves a few things around to make it possible to create a ClutterClone of the ClutterActor that holds the desktop's grid, following the lead of existing code in eos-desktop, but adapted to the new shell.

https://phabricator.endlessm.com/T17662

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -387,7 +387,9 @@ const AllViewContainer = new Lang.Class({
     Name: 'AllViewContainer',
     Extends: St.Widget,
 
-    _init: function(gridActor) {
+    _init: function(gridActor, params) {
+        params = Params.parse(params, { allowScrolling: true });
+
         this.parent({ layout_manager: new Clutter.BinLayout(),
                       x_expand: true,
                       y_expand: true });
@@ -402,7 +404,7 @@ const AllViewContainer = new Lang.Class({
                                               y_expand: true,
                                               x_fill: true,
                                               y_fill: false,
-                                              reactive: true,
+                                              reactive: params.allowScrolling,
                                               hscrollbar_policy: Gtk.PolicyType.NEVER,
                                               vscrollbar_policy: Gtk.PolicyType.EXTERNAL,
                                               y_align: Clutter.ActorAlign.START });
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -371,25 +371,55 @@ const PageIndicators = new Lang.Class({
 });
 Signals.addSignalMethods(PageIndicators.prototype);
 
+const AllViewContainer = new Lang.Class({
+    Name: 'AllViewContainer',
+    Extends: St.Widget,
+
+    _init: function(gridActor) {
+        this.parent({ layout_manager: new Clutter.BinLayout(),
+                      x_expand: true,
+                      y_expand: true });
+
+        this.gridActor = gridActor;
+
+        gridActor.y_expand = true;
+        gridActor.y_align = Clutter.ActorAlign.CENTER;
+
+        this.scrollView = new St.ScrollView({ style_class: 'all-apps',
+                                              x_expand: true,
+                                              y_expand: true,
+                                              x_fill: true,
+                                              y_fill: false,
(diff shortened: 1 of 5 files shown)


js/ui/appDisplay.js
```diff
@@ -371,25 +371,54 @@ const PageIndicators = new Lang.Class({
 });
 Signals.addSignalMethods(PageIndicators.prototype);
 
+const AllViewContainer = new Lang.Class({
+    Name: 'AllViewContainer',
+    Extends: St.Widget,
+
+    _init: function(gridActor) {
+        this.parent({ layout_manager: new Clutter.BinLayout(),
+                      x_expand:true, y_expand:true });
```
cosimoc: Missing spaces before colons

js/ui/viewSelector.js
```diff
@@ -453,6 +453,99 @@ const ViewsDisplay = new Lang.Class({
     }
 });
 
+const ViewsClone = new Lang.Class({
+    Name: 'ViewsClone',
+    Extends: St.Widget,
+
+    _init: function(viewSelector, viewsDisplay, forOverview) {
+        let settings = Clutter.Settings.get_default();
```
cosimoc: Nit: move this line down where the settings are actually used

js/ui/viewSelector.js
```diff
@@ -644,6 +647,18 @@ const ViewSelector = new Lang.Class({
         global.stage.add_action(gesture);
     },
 
+    _addViewsPageClone: function() {
+        let layoutViewsClone = new ViewsClone(this, this._viewsDisplay, false);
+        Main.layoutManager.setViewsClone(layoutViewsClone);
+
+        this._overviewViewsClone = new ViewsClone(this, this._viewsDisplay, true);
```
cosimoc: Do you need to hold a reference to this in `ViewSelector`?

js/ui/layout.js
```diff
@@ -309,6 +309,11 @@ const LayoutManager = new Lang.Class({
         this.emit('hot-corners-changed');
     },
 
+    setViewsClone: function(actor) {
+        this._viewsClone = actor;
```
cosimoc: Perhaps should also initialized to `null` in the constructor, like in the layout manager class? (Or remove it from there too.)

js/ui/appDisplay.js
mariospr: You mean 'after', I guess :-). Ok

js/ui/viewSelector.js
mariospr: Ok

js/ui/viewSelector.js
```diff
@@ -453,6 +453,99 @@ const ViewsDisplay = new Lang.Class({
...
+        entry.clutter_text.reactive = false;
+
+        let iconGridClone = new Clutter.Clone({ source: appDisplay.gridActor,
+                                                x_expand: true,
+                                                y_expand: true });
+
+        let appGridContainer = new AppDisplay.AllViewContainer(iconGridClone);
+        appGridContainer.reactive = false;
+
+        let layoutManager = new ViewsDisplayLayout(entry, appGridContainer, null);
```
mariospr: Technically, the second parameter of ViewsDisplayLayout's constructor is an `appDisplay` which is no longer true when creating the clone (it's a `AllViewContainer`) so I think I'll rename some names in ViewsDisplayContainer and ViewsDisplayLayout to make it more generic (e.g. maybe call it 'grid container')

js/ui/layout.js
mariospr: I'll do, thanks for pointing it out. Yet another case of code brought from the old shell

js/ui/viewSelector.js
mariospr: I guess no? Again, not my code, so I just ported it and assumed this was needed for some reason. I'll try removing the reference though, can't find any other place this is needed, or a reason to keep a ref from here when the overview is already keeping one.

js/ui/viewSelector.js
mariospr: Updated in the new PR

PR #47, opened 2017-06-19
T16876

smspillaz:
https://phabricator.endlessm.com/T16876


js/ui/viewSelector.js
```diff
@@ -74,6 +74,72 @@ function getTermsForSearchString(searchString) {
     return terms;
 }
 
+function maybeCreateInactiveDiscoveryFeedButton() {
```
cosimoc: Can you move all of these feed-related functions in a separate file?

js/ui/viewSelector.js
```diff
@@ -243,6 +322,11 @@ const ViewsDisplayLayout = new Lang.Class({
         this._appDisplayActor.opacity = (1 - v) * 255;
         this._searchResultsActor.opacity = v * 255;
 
+        if (this._discoveryFeedButton !== null) {
```
cosimoc: This check is inconsistent with the one at line 311. I'd use that style in both places.

js/ui/viewSelector.js
```diff
@@ -260,21 +344,29 @@ const ViewsDisplayContainer = new Lang.Class({
...
+                                                   searchResults.actor);
+
         this.parent({ layout_manager: layoutManager,
                       x_expand: true,
                       y_expand: true });
 
         layoutManager.connect('allocated-size-changed', Lang.bind(this, this._onAllocatedSizeChanged));
 
         this.add_child(this._entry);
+        if (this._discoveryFeedButton !== null)
```
cosimoc: See above

js/ui/viewSelector.js
```diff
@@ -327,6 +419,7 @@ const ViewsDisplay = new Lang.Class({
     _init: function() {
         this._enterSearchTimeoutId = 0;
         this._localSearchMetricTimeoutId = 0;
+        this.discoveryFeed = maybeCreateDiscoveryFeed();
```
cosimoc: I don't think there's a reason to make this "public", so probably should be `this._discoveryFeed`

js/ui/overview.js
```diff
@@ -271,14 +271,14 @@ const Overview = new Lang.Class({
...
                                         opacity: 0 });
-        this._overview.add_actor(this._panelGhost);
 
         // Create controls
         this._controls = new OverviewControls.ControlsManager();
         this.viewSelector = this._controls.viewSelector;
 
         // Add our same-line elements after the search entry
         this._overview.add(this._controls.actor, { y_fill: true, expand: true });
+        this._overview.add(this._panelGhost);
```
cosimoc: Mario is already fixing this in https://phabricator.endlessm.com/T17662 and I prefer to wait until that's merged before changing this code -- so we'll wait to merge this PR until after that.

js/ui/overview.js
```diff
@@ -791,6 +791,10 @@ const Overview = new Lang.Class({
             this.hide();
         else
             this.show();
+    },
+
+    getActivePage: function() {
```
cosimoc: Is this actually used somewhere?

js/ui/overview.js
smspillaz: Heh, I spent ages trying to work out what was going on here ;-) @mariospr  <- not sure if this was a problem you were looking at but feel free to cherry pick this commit.

js/ui/overview.js
smspillaz: Yeah, it is used by SideComponent. It looks like we forgot to port this code over when we ported over SideComponent.

js/ui/overview.js
mariospr: The problem I saw was that the cloned desktop was being shifted vertically to the top because of this ghost, that was there because upstream the panel is at the top. Moving it to the bottom fixed it

PR #49, opened 2017-06-19
endlessButton: Change hover label according to current view

GeorgesStavracas:
When viewing apps, the hover label should read "Show Desktop",
and while viewing the desktop, the hover label should read
"Show Apps".

https://phabricator.endlessm.com/T17520

### DIFF
--- a/js/ui/endlessButton.js
+++ b/js/ui/endlessButton.js
@@ -7,6 +7,7 @@ const St = imports.gi.St;
 
 const Main = imports.ui.main;
 const PanelMenu = imports.ui.panelMenu;
+const ViewSelector = imports.ui.viewSelector;
 
 const EndlessButton = new Lang.Class({
     Name: 'EndlessButton',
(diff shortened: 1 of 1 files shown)


js/ui/endlessButton.js
```diff
@@ -29,15 +30,30 @@ const EndlessButton = new Lang.Class({
...
         this._labelOffsetY = 0;
         this._label.connect('style-changed', Lang.bind(this, function(actor, forHeight, alloc) {
             this._labelOffsetX = this._label.get_theme_node().get_length('-label-offset-x');
             this._labelOffsetY = this._label.get_theme_node().get_length('-label-offset-y');
         }));
+
+        Main.overview.connect('page-changed', Lang.bind(this, this._updateHoverLabel));
+
+        this.actor.connect('destroy', Lang.bind(this, function() {
+            Main.overview.disconnect(this._updateHoverLabel);
```
cosimoc: Pretty sure this doesn't do what you expect... `disconnect()` takes a signal ID, so the correct way to do this would be
```
let pageChangedId = Main.overview.connect(...);
this.actor.connect('destroy', function() {
    Main.overview.disconnect(pageChangedId);
});
```

PR #51, opened 2017-06-20
theme: Remove spacing between elements in overview

smspillaz:
We don't want this as we need perfect alignment.

### DIFF
--- a/data/theme/gnome-shell.css
+++ b/data/theme/gnome-shell.css
@@ -2119,6 +2119,10 @@ stage {
     margin: 14px 0px;
 }
 
+#overview {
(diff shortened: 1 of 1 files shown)


mariospr:
One of the two commits has already being applied as part of PR #45: https://github.com/endlessm/gnome-shell/pull/45/commits/83e239e5acd131750d587b56e39179e4559d7431

Would you mind preparing another PR with the remaining commit? That's still valid

smspillaz:
@mariosp rebased and done

mariospr:
Thanks, specially because I just found out that I needed this change anyway for T17658 :-). Merged

PR #54, opened 2017-06-20
DiscoveryFeed: add hover state for tile

erikos:
This also makes the tile itself a bit bigger.


js/ui/discoveryFeedButton.js
```diff
@@ -51,15 +51,25 @@ const DiscoveryFeedButton = new Lang.Class({
...
+        let iconFileHover = Gio.File.new_for_uri('resource:///org/gnome/shell/theme/discovery-feed-open-tab_hover.png');
+        this._giconHover = new Gio.FileIcon({ file: iconFileHover });
+        this._icon = new St.Icon({ gicon: this._giconNormal,
                                    style_class: 'discovery-feed-icon',
                                    track_hover: true });
         this.parent({ name: 'discovery-feed',
                       child: this._icon,
                       x_align: Clutter.ActorAlign.CENTER,
                       y_align: Clutter.ActorAlign.CENTER });
+        this.connect('notify::hover', Lang.bind(this, this._onHoverChanged));
```
smspillaz: Hmm, my preference here would be to use the css `background-image` attribute with a `:hover` state.

erikos: Thanks Sam for pushing me to this, that looks much better now indeed. I was not sure about the exact alignment in css, there are different styles used, I followed the page-indicator example.

erikos:
I sneaked in here a small fix for CodeView to not have another PR: Gjs-Message: JS WARNING: [/home/mario/work/endless/shell-rebase/gnome-shell/js/ui/codeView.js 1006]: anonymous function does not always return a value

js/ui/discoveryFeedButton.js
```diff
@@ -51,12 +51,10 @@ const DiscoveryFeedButton = new Lang.Class({
     Extends: St.Button,
 
     _init: function() {
-        let iconFile = Gio.File.new_for_uri('resource:///org/gnome/shell/theme/discovery-feed-open-tab.png');
-        let gicon = new Gio.FileIcon({ file: iconFile });
-        this._icon = new St.Icon({ gicon: gicon,
-                                   style_class: 'discovery-feed-icon',
-                                   track_hover: true });
+        this._icon = new St.Icon({style_class: 'discovery-feed-tile-icon',
```
cosimoc: If you set the background with CSS, you should not need this `St.Icon` child on the button anymore. Try simply setting the background image on the `.discovery-feed-tile` style class of the button (which however I would rename to `.discovery-feed-button`)

erikos: Event better, thanks for the suggestion - it really starts to look nice clean code now.

cosimoc:
@erikos Can you rebase this to the latest master? Somehow it has a commit of mine that's already in master.

erikos:
@cosimoc, rebased, sorry must have happened in rebasing here when tweaking.

PR #55, opened 2017-06-20
Style changes for the desktop's icon grid

mariospr:
This PR makes some changes to make the desktop look more EOS-ish, with the grid of icons properly spaced and centered and the icons being of the right size.

https://phabricator.endlessm.com/T17658

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -33,7 +33,7 @@ const MAX_APPLICATION_WORK_MILLIS = 75;
 const MENU_POPUP_TIMEOUT = 600;
 const MAX_COLUMNS = 7;
 const MIN_COLUMNS = 4;
-const MIN_ROWS = 4;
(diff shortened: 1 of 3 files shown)


js/ui/appDisplay.js
```diff
@@ -457,8 +457,8 @@ const AllView = new Lang.Class({
         panAction.connect('gesture-cancel', Lang.bind(this, this._onPanEnd));
         panAction.connect('gesture-end', Lang.bind(this, this._onPanEnd));
         this._panAction = panAction;
-        this._scrollView.add_action(panAction);
         this._panning = false;
+
```
joaquimrocha: This newline does not belong to this patch..

mariospr: I added it on purpose to separate these two blocks, as one is about panning and the other about folders. I'd like to keep it

PR #57, opened 2017-06-20
Implement the EOS style for the search results panel

mariospr:
https://phabricator.endlessm.com/T17659

### DIFF
--- a/js/ui/search.js
+++ b/js/ui/search.js
@@ -25,6 +25,7 @@ const SEARCH_PROVIDERS_SCHEMA = 'org.gnome.desktop.search-providers';
 
 const MAX_LIST_SEARCH_RESULTS_ROWS = 3;
 const MAX_GRID_SEARCH_RESULTS_ROWS = 1;
+const MAX_GRID_SEARCH_RESULTS_COLS = 8;
 
 const MaxWidthBin = new Lang.Class({
     Name: 'MaxWidthBin',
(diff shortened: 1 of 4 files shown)


js/ui/search.js
```diff
@@ -819,7 +819,6 @@ const ProviderIcon = new Lang.Class({
 
         this._content = new St.Widget({ layout_manager: new Clutter.BinLayout() });
         this._content.add_actor(icon);
-        this._content.add_actor(this.moreIcon);
```
cosimoc: Please remove `moreIcon` altogether, instead of simply not adding it to the container.

js/ui/search.js
```diff
@@ -377,7 +377,7 @@ const GridSearchResults = new Lang.Class({
     },
 
     _createResultDisplay: function(meta) {
-        return this.parent(meta) || new GridSearchResult(this.provider, meta);
```
cosimoc: It strikes me that we may as using `createResultObject()`, which creates an `AppInfo` for the app search provider, is actually a good fit for the applications. Any reason not to style that, instead of changing the upstream behavior?

js/ui/search.js
```diff
@@ -333,7 +333,7 @@ const ListSearchResults = new Lang.Class({
     },
 
     _createResultDisplay: function(meta) {
-        return this.parent(meta) || new ListSearchResult(this.provider, meta);
```
cosimoc: I don't believe there's anything that would override this for list results, so you could remove this change

mariospr: I think you're right, I'll remove it. Also, I've just noticed a bug in my PR that caused apps launched from the grid not to work because I didn't add the `activateResult()` method in `AppSearchProvider`, which I'm adding now. And while adding that I'll probably can remove the `createResultObject()` method from there as it won't be needed anymore, rendering this change here useless.

js/ui/search.js
mariospr: I'm not 100% sure now, but when I looked into this I think that would certainly be possible but I ended up discarding it because (1) that would force me to re-style things without reusing much of the stuff from eos 3.1, (2) because that way I could simply integrate the results with AppActivation by means of an `activateResult()` method in the AppSearchProvider (which I forgot to add, but let's ignore that detail for now...) and (3) because the patch looked pretty simple anyway.

I can spend more time if you want trying to do that, but with more pressing needs happening maybe that's a separate ticket that we can work on afterwards (and perhaps a good one for @GeorgesStavracas as he's way better than me at styling things :-))?

js/ui/search.js
mariospr: Ok

js/ui/search.js
cosimoc: OK, no worries -- we can leave with your approach for now.

PR #61, opened 2017-06-21
appDisplay, iconGrid: Fix unwanted scrollbars inside folder popups

mariospr:
https://phabricator.endlessm.com/T17658

### DIFF
--- a/js/ui/iconGrid.js
+++ b/js/ui/iconGrid.js
@@ -807,7 +807,12 @@ const PaginatedIconGrid = new Lang.Class({
         // We want to contain the grid inside the parent box with padding
         this._rowsPerPage = Math.min(Math.max (this._minRows, nRows), this.rowsForHeight(availHeightPerPage));
         this._nPages = Math.ceil(nRows / this._rowsPerPage);
-        this._spaceBetweenPages = this._nPages > 1 ? (availHeightPerPage - (this.topPadding + this.bottomPadding) - this._availableHeightPerPageForItems()) : 0;
+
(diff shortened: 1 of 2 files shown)


js/ui/iconGrid.js
```diff
@@ -807,7 +807,7 @@ const PaginatedIconGrid = new Lang.Class({
         // We want to contain the grid inside the parent box with padding
         this._rowsPerPage = Math.min(Math.max (this._minRows, nRows), this.rowsForHeight(availHeightPerPage));
         this._nPages = Math.ceil(nRows / this._rowsPerPage);
-        this._spaceBetweenPages = this._nPages > 1 ? (availHeightPerPage - (this.topPadding + this.bottomPadding) - this._availableHeightPerPageForItems()) : 0;
+        this._spaceBetweenPages = this._nPages > 1 ? (availHeightPerPage - (this.topPadding + this.bottomPadding) - this._availableHeightPerPageForItems()) : this._getSpacing();
```
joaquimrocha: This line should probably be broken as it's getting long.

mariospr: Ok

mariospr:
> ... which now get a smaller allocation than the grid than the one they need...
> 
> The message is confusing.
> Having more succinct messages (when possible) would also make it easier to understand.

I shortened a bit and made it more concise, but I'd rather fail on the side of having extra text here because it was a very tricky thing to realize and don't want to go through the same pain if possible when doing the next rebase :-)

PR #64, opened 2017-06-22
iconGrid: Recompute spaces when going low resolution

GeorgesStavracas:
When the screen is in the range of what we consider a low
resolution screen, the calculated spacing of the desktop
icon grid must be adapted.

Currently, however, this is not happening, since the spacing
calculation method is not being called anymore. This was
necessary to guarantee spacing consistency in normal screens,
but low resolution must have the spacings adapted.

Fix that by calling _updateSpacingForSize when in low resolution
displays.

https://phabricator.endlessm.com/T17733


mariospr:
I'm closing this for now, as it's been open since June and does not really apply anymore. 

@GeorgesStavracas, if you want to rebase it and work on it after the upgrade to 3.26, then it's fine but for now it's just cluttering the list of open PRs.

PR #65, opened 2017-06-23
T17847

joaquimrocha:
https://phabricator.endlessm.com/T17847

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -548,6 +548,11 @@ const AllView = new Lang.Class({
         this._addedFolderId = null;
         IconGridLayout.layout.connect('folder-added',
                                       Lang.bind(this, function(iconGridLayout, id){
+                                          // Go to last page; ideally the grid should know in
+                                          // which page the change took place and show it automatically
+                                          // which would avoid us having to navigate there directly
+                                          this.goToPage(this._grid.nPages() - 1);
+
                                           // Save the folder ID so we know which one was added
                                           // and set it to edit mode
                                           this._addedFolderId = id;
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1097,12 +1097,19 @@ const FolderView = new Lang.Class({
         this.actor = new St.ScrollView({ overlay_scrollbars: true });
         this.actor.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC);
         let scrollableContainer = new St.BoxLayout({ vertical: true, reactive: true });
+        this._noAppsLabel = new St.Label({ text: _("No apps in this folder! To add an app, drag it onto the folder."),
+                                           style_class: 'folder-no-apps-label'});
+        scrollableContainer.add_actor(this._noAppsLabel);
         scrollableContainer.add_actor(this._grid.actor);
         this.actor.add_actor(scrollableContainer);
 
         let action = new Clutter.PanAction({ interpolate: true });
         action.connect('pan', Lang.bind(this, this._onPan));
         this.actor.add_action(action);
+    },
+
+    updateNoAppsLabelVisibility: function() {
+        this._noAppsLabel.visible = this._grid.visibleItemsCount() == 0;
     },
 
     _keyFocusIn: function(actor) {
@@ -1298,8 +1305,8 @@ const FolderIcon = new Lang.Class({
         let folderApps = IconGridLayout.layout.getIcons(this.id);
         folderApps.forEach(addAppId);
 
-        this.actor.visible = this.view.getAllItems().length > 0;
         this.view.loadGrid();
+        this.view.updateNoAppsLabelVisibility();
         this.emit('apps-changed');
     },
 
@@ -1319,7 +1326,8 @@ const FolderIcon = new Lang.Class({
                 this._popup.popup();
                 this._updatePopupPosition();
             }));
-        this._parentView.openSpaceForPopup(this, this._boxPointerArrowside, this.view.nRowsDisplayedAtOnce());
+        this._parentView.openSpaceForPopup(this, this._boxPointerArrowside,
+                                           Math.max(this.view.nRowsDisplayedAtOnce(), 1));
     },
 
     _calculateBoxPointerArrowSide: function() {
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -18,6 +18,7 @@ const AppFavorites = imports.ui.appFavorites;
 const BackgroundMenu = imports.ui.backgroundMenu;
 const BoxPointer = imports.ui.boxpointer;
 const DND = imports.ui.dnd;
+const EditableLabelMode = imports.ui.editableLabel.EditableLabelMode;
 const GrabHelper = imports.ui.grabHelper;
 const IconGrid = imports.ui.iconGrid;
 const IconGridLayout = imports.ui.iconGridLayout;
@@ -193,21 +194,29 @@ const BaseAppView = new Lang.Class({
     },
 
     selectApp: function(id) {
+        this.selectAppWithLabelMode(id, null);
+    },
+
+    selectAppWithLabelMode: function(id, labelMode) {
         if (this._items[id] && this._items[id].actor.mapped) {
             this._selectAppInternal(id);
+            if (labelMode !== null)
+                this._items[id].icon.setLabelMode(labelMode);
         } else if (this._items[id]) {
             // Need to wait until the view is mapped
             let signalId = this._items[id].actor.connect('notify::mapped', Lang.bind(this, function(actor) {
                 if (actor.mapped) {
                     actor.disconnect(signalId);
                     this._selectAppInternal(id);
+                    if (labelMode !== null)
+                        this._items[id].icon.setLabelMode(labelMode);
                 }
             }));
         } else {
             // Need to wait until the view is built
             let signalId = this.connect('view-loaded', Lang.bind(this, function() {
                 this.disconnect(signalId);
-                this.selectApp(id);
+                this.selectAppWithLabelMode(id, labelMode);
             }));
         }
     },
@@ -536,6 +545,14 @@ const AllView = new Lang.Class({
             Main.queueDeferredWork(this._redisplayWorkId);
         }));
 
+        this._addedFolderId = null;
+        IconGridLayout.layout.connect('folder-added',
+                                      Lang.bind(this, function(iconGridLayout, id){
+                                          // Save the folder ID so we know which one was added
+                                          // and set it to edit mode
+                                          this._addedFolderId = id;
+                                      }));
+
         this._loadApps();
     },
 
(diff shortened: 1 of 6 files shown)


js/ui/appDisplay.js
```diff
@@ -1091,6 +1111,9 @@ const FolderView = new Lang.Class({
         this.actor = new St.ScrollView({ overlay_scrollbars: true });
         this.actor.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC);
         let scrollableContainer = new St.BoxLayout({ vertical: true, reactive: true });
+        this.no_apps_label = new St.Label({ text: 'No apps in this folder! To add an app, drag it onto the folder.',
```
rshuler: Please make this string translatable.  I.e., use `_("...")`

joaquimrocha: Oops... I left it out when doing the first version and forgot to update it.

js/ui/appDisplay.js
```diff
@@ -1091,6 +1091,9 @@ const FolderView = new Lang.Class({
         this.actor = new St.ScrollView({ overlay_scrollbars: true });
         this.actor.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC);
         let scrollableContainer = new St.BoxLayout({ vertical: true, reactive: true });
+        this.no_apps_label = new St.Label({ text: _('No apps in this folder! To add an app, drag it onto the folder.'),
```
cosimoc: Please use camel case -- there should also be no reason to make this "public".
Finally, the coding style uses double quotes for translatable strings.

js/ui/iconGridLayout.js
```diff
@@ -379,6 +380,13 @@ const IconGridLayout = new Lang.Class({
                 Lang.bind(this, this._enumerateDirectoryFiles));
     },
 
+    addFolder: function() {
+        let id = FolderModel.model.createFolder(_('New Folder'), 'icon');
```
cosimoc: Double quotes for translatable strings

js/ui/folderModel.js
```diff
@@ -0,0 +1,63 @@
...
+                n = Math.max(n, parseInt(new_n));
+            }
+        }
+        enumerator.close(null);
+
+        let filename = prefix + (n+1) + suffix;
+
+        let buf = keyFile.to_data();
+
+        let channel = GLib.IOChannel.new_file(dirpath + '/' + filename, 'w');
```
cosimoc: This is a very weird way to save a keyfile... What's wrong with `g_key_file_save_to_file()`?

js/ui/folderModel.js
```diff
@@ -0,0 +1,63 @@
...
+
+        let dirpath = GLib.build_filenamev([GLib.get_user_data_dir(), '/desktop-directories']);
+
+        if (GLib.mkdir_with_parents(dirpath, parseInt('0755', 8), null) < 0) {
+            log('could not create the directory ' + dirpath);
+            return null;
+        }
+
+        let dir =  Gio.File.new_for_path(dirpath);
+        let enumerator = dir.enumerate_children('standard::name',
```
cosimoc: This code needs some try/catch, since the operation can fail

js/ui/folderModel.js
```diff
@@ -0,0 +1,63 @@
...
+
+        keyFile.set_value(GLib.KEY_FILE_DESKTOP_GROUP,
+                          GLib.KEY_FILE_DESKTOP_KEY_ICON, iconName);
+
+        keyFile.set_value(GLib.KEY_FILE_DESKTOP_GROUP,
+                          GLib.KEY_FILE_DESKTOP_KEY_TYPE, GLib.KEY_FILE_DESKTOP_TYPE_DIRECTORY);
+
+        let dirpath = GLib.build_filenamev([GLib.get_user_data_dir(), '/desktop-directories']);
+
+        if (GLib.mkdir_with_parents(dirpath, parseInt('0755', 8), null) < 0) {
```
cosimoc: You can use `g_file_make_directory_with_parents()`

js/ui/folderModel.js
```diff
@@ -0,0 +1,63 @@
...
+        let channel = GLib.IOChannel.new_file(dirpath + '/' + filename, 'w');
+        channel.write_chars(buf[0], buf[1]);
+        channel.shutdown(true);
+
+        return filename;
+    }
+});
+
+// Use as a singleton
+const model = new FolderModel();
```
cosimoc: Instead of making this a singleton, you could simply avoid making a class and call the function directly

js/ui/appDisplay.js
```diff
@@ -193,21 +194,29 @@ const BaseAppView = new Lang.Class({
     },
 
     selectApp: function(id) {
+        this.selectAppWithLabelMode(id, null);
+    },
+
+    selectAppWithLabelMode: function(id, label_mode) {
```
cosimoc: Use camel case

js/ui/appDisplay.js
```diff
@@ -536,6 +545,13 @@ const AllView = new Lang.Class({
             Main.queueDeferredWork(this._redisplayWorkId);
         }));
 
+        this._addedFolderId = null;
+        IconGridLayout.layout.connect('folder-added',
```
cosimoc: Could you pass a callback to `addFolder()` directly instead of adding the signal indirection and the extra state?

js/ui/folderModel.js
joaquimrocha: I don't know. I just copied this file from //eos-desktop// and didn't really change anything but the return value.

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+        let keyFile = new GLib.KeyFile();
+
+        keyFile.set_value(GLib.KEY_FILE_DESKTOP_GROUP,
+                          GLib.KEY_FILE_DESKTOP_KEY_NAME, name);
+
+        keyFile.set_value(GLib.KEY_FILE_DESKTOP_GROUP,
+                          GLib.KEY_FILE_DESKTOP_KEY_TYPE, GLib.KEY_FILE_DESKTOP_TYPE_DIRECTORY);
+
+        let dir = Gio.File.new_for_path(GLib.get_user_data_dir() + '/desktop-directories');
+        if (!dir.query_exists(null))
```
cosimoc: Instead of doing this, simply call `make_directory_with_parents()` and catch the error that matches `G_IO_ERROR_EXISTS` code

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+            try {
+                dir.make_directory_with_parents(null);
+            } catch (err) {
+                log('Error creating %s: %s'.format(dir.get_path(), err.message));
```
cosimoc: Use `logError()` instead

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+                return null;
+            }
+
+        let enumerator = null;
+        try {
+            enumerator = dir.enumerate_children('standard::name', Gio.FileQueryInfoFlags.NONE, null);
+        } catch (err) {
+            log('Error trying to traverse %s: %s'.format(dir.get_path(), err.message));
```
cosimoc: Use `logError()` instead

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+        try {
+            for (let f = enumerator.next_file(null); f != null; f = enumerator.next_file(null)) {
+                let result = re.exec(f.get_name());
+                if (result != null) {
+                    let new_n = result[1];
+                    n = Math.max(n, parseInt(new_n));
+                }
+            }
+        } catch (err) {
+            log('Error traversing %s: %s'.format(dir.get_path(), err.message));
```
cosimoc: Use `logError()` instead

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+            return null;
+        }
+
+        try {
+            enumerator.close(null);
+        } catch (err) {
+            log('Error closing file enumerator for %s: %s'.format(dir.get_path(), err.message));
```
cosimoc: Use `logError()` instead -- but I don't think this error should be considered fatal

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+            return null;
+        }
+
+        let filename = prefix + (n+1) + suffix;
+        let absFilename = GLib.build_filenamev([dir.get_path(), filename]);
+
+        try {
+            keyFile.save_to_file(absFilename);
+        } catch (err) {
+            log('Failed to save key file for directory %s: %s'.format(absFilename, err.message));
```
cosimoc: Use `logError()` instead

js/ui/iconGridLayout.js
```diff
@@ -379,6 +379,77 @@ const IconGridLayout = new Lang.Class({
...
+        let prefix = 'eos-folder-user-';
+        let suffix = '.directory';
+        let re = new RegExp(prefix + '([0-9]+)' + suffix);
+        let n = -1;
+
+        try {
+            for (let f = enumerator.next_file(null); f != null; f = enumerator.next_file(null)) {
+                let result = re.exec(f.get_name());
+                if (result != null) {
+                    let new_n = result[1];
```
cosimoc: Use camel case

js/ui/editableLabel.js
```diff
@@ -205,7 +195,41 @@ const EditableLabel = new Lang.Class({
...
+                this.setMode(EditableLabelMode.HIGHLIGHT);
+
+            // enter the edit mode on highlight ungrab
+            if (this._grabHelper.grabbed)
+                this._grabHelper.ungrab({ actor: this });
+            this.grab_key_focus();
+            break;
+
+        default:
+            log('Invalid mode for EditableLabel: ' + mode);
```
cosimoc: Should probably throw an error here

js/ui/appDisplay.js
```diff
@@ -548,6 +548,11 @@ const AllView = new Lang.Class({
         this._addedFolderId = null;
         IconGridLayout.layout.connect('folder-added',
                                       Lang.bind(this, function(iconGridLayout, id){
+                                          // Go to last page; ideally the grid should know in
+                                          // which page the change took place and show it automatically
+                                          // which would avoid us having to navigate there directly
+                                          this.goToPage(this._grid.nPages() - 1);
```
cosimoc: I am OK with this change for now, but wonder how much more difficult would it be to indicate to the icon grid layout to add the folder at the right place, instead of at the end? I guess you'd still have to take into account the case when the user right clicks e.g. outside of the grid bounds entirely, in which case you'd still need to append it.

joaquimrocha: The logic is obviously easier and in place for an append, rather than an inline addition that would depend on the position of the pointer, etc. I think we should keep this for this new release and work with the Design team is having an inline insertion is important.

js/ui/iconGridLayout.js
joaquimrocha: I usually try to catch exceptions only when something is not expected to fail normally. This directory will exist right after the first folder is added to the desktop, so its existence is what's expected. But I will modify it as you say.

PR #66, opened 2017-06-26
appDisplay: don't crash if app center isn't installed

wjt:
`eosinstaller` images do not have the app center. Ideally we should arrange for `enable-app-center` to be `false` in GSettings, but we should still not crash if it's `true`.

https://phabricator.endlessm.com/T17875

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -594,9 +594,15 @@ const AllView = new Lang.Class({
                 this.addItem(icon);
         }));
 
-        // Add the App Center icon if it is enabled
-        if (global.settings.get_boolean(EOS_ENABLE_APP_CENTER_KEY))
-            this.addItem(new AppCenterIcon());
+        // Add the App Center icon if it is enabled and installed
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -594,9 +594,16 @@ const AllView = new Lang.Class({
...
 
-        // Add the App Center icon if it is enabled
-        if (global.settings.get_boolean(EOS_ENABLE_APP_CENTER_KEY))
-            this.addItem(new AppCenterIcon());
+        // Add the App Center icon if it is enabled and installed
+        if (global.settings.get_boolean(EOS_ENABLE_APP_CENTER_KEY)) {
+            let app = appSys.lookup_app(EOS_APP_CENTER_ID);
+            if (app) {
+                this.addItem(new AppCenterIcon(app));
+            } else {
```
joaquimrocha: One-liner blocks should not have curly brackets.

wjt: Updated. I don't like this style and found examples of my preferred style in this file ☺ but I'm happy to conform.

PR #69, opened 2017-06-27
T17840

joaquimrocha:
https://phabricator.endlessm.com/T17840

### DIFF
--- a/js/ui/iconGrid.js
+++ b/js/ui/iconGrid.js
@@ -884,6 +884,13 @@ const PaginatedIconGrid = new Lang.Class({
                                                : sourceRow;
         let nRowsBelow = this._rowsPerPage - nRowsAbove;
 
+        // Since it always tries to show up the folder icon, then when only 1 row is
+        // being displayed, the number of rows (to be moved out) here is 0; however
+        // we override that because it's better to move the folder icon out of the
+        // view than not showing the folder popup at all.
+        if (nRows == 0)
+            nRows = 1;
+
         let nRowsUp, nRowsDown;
         if (side == St.Side.TOP) {
             nRowsDown = Math.min(nRowsBelow, nRows);
--- a/js/ui/iconGrid.js
+++ b/js/ui/iconGrid.js
@@ -766,7 +766,7 @@ const PaginatedIconGrid = new Lang.Class({
         }
 
         let x = box.x1 + leftEmptySpace + this.leftPadding;
-        let y = box.y1 + this.topPadding;
+        let y = box.y1 + this.topPadding + this._spaceBetweenPages / 2;
         let columnIndex = 0;
         let rowIndex = 0;
 
@@ -846,7 +846,7 @@ const PaginatedIconGrid = new Lang.Class({
 
         let firstPageItem = pageNumber * this._childrenPerPage
         let childBox = this._getVisibleChildren()[firstPageItem].get_allocation_box();
-        return childBox.y1 - this.topPadding;
+        return childBox.y1 - this.topPadding - this._spaceBetweenPages / 2;
     },
 
     getItemPage: function(item) {
--- a/js/ui/viewSelector.js
+++ b/js/ui/viewSelector.js
@@ -141,8 +141,8 @@ const ShowOverviewAction = new Lang.Class({
 const ViewsDisplayLayout = new Lang.Class({
     Name: 'ViewsDisplayLayout',
     Extends: Clutter.BinLayout,
-    Signals: { 'allocated-size-changed': { param_types: [GObject.TYPE_INT,
-                                                         GObject.TYPE_INT] } },
+    Signals: { 'grid-available-size-changed': { param_types: [GObject.TYPE_INT,
+                                                              GObject.TYPE_INT] } },
 
     _init: function(entry, discoveryFeedButton, gridContainerActor, searchResultsActor) {
         this.parent();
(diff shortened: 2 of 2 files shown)


mariospr:
Small nit in the message from the first commit: "The icon grid was being given a bigger hight..." --> "height"

js/ui/viewSelector.js
```diff
@@ -141,8 +141,8 @@ const ShowOverviewAction = new Lang.Class({
 const ViewsDisplayLayout = new Lang.Class({
     Name: 'ViewsDisplayLayout',
     Extends: Clutter.BinLayout,
-    Signals: { 'allocated-size-changed': { param_types: [GObject.TYPE_INT,
-                                                         GObject.TYPE_INT] } },
+    Signals: { 'grid-available-size-changed': { param_types: [GObject.TYPE_INT,
+                                                              GObject.TYPE_INT] } },
```
mariospr: I understand the name change and I agree it makes sense as we're now sending the available size that we calculated for the grid (instead of the size allocated to the ViewsDisplayContainer), but I thinking twice about this I don't think we should emit any signal at all:

This signal is being connected to from ViewsDisplayContainer only, and its only purpose is to make sure we call `adaptToSize()` over the grid container before the allocation, which we could do directly from here by calling `gridContainerActor.adaptToSize()`, allowing us to remove all the code related to this signal from here and move the logic of `_onGridAvailableSizeChanged()` from ViewsDisplayContainer right into a method in the layout manager.

It would be nice if you could make that change, I think it would simplify things and make it more readable.

js/ui/viewSelector.js
```diff
@@ -211,7 +211,9 @@ const ViewsDisplayLayout = new Lang.Class({
 
         let gridContainerBox = allocation.copy();
         gridContainerBox.y1 = this._computeGridContainerPlacement(gridContainerHeight, entryHeight, availHeight);
-        gridContainerBox.y2 = Math.min(gridContainerBox.y1 + gridContainerHeight, allocation.y2);
+        // If the icon grid's current height is 0 we still want to give its box the available height
```
mariospr: This seems a bit confusing to read. Also, it seems wrong: `gridContainerHeight` plays a part in calculating the `y1` coordinate just above, so it looks like ignoring that here and use the allocation.y2 value when the determined height for the grid is zero could cause some trouble. Are you sure this works as expected?

js/ui/viewSelector.js
```diff
@@ -211,7 +211,9 @@ const ViewsDisplayLayout = new Lang.Class({
 
         let gridContainerBox = allocation.copy();
         gridContainerBox.y1 = this._computeGridContainerPlacement(gridContainerHeight, entryHeight, availHeight);
```
mariospr: I know this line is not part of this patch (it's my fault actually), but it would probably be nice to replace it with `gridContainerBox.y1 = heightAboveGrid` to prevent doing the same calculations twice

js/ui/viewSelector.js
```diff
@@ -233,7 +235,8 @@ const ViewsDisplayLayout = new Lang.Class({
         // We want to emit the signal BEFORE any allocation has happened since the
         // icon grid will need to precompute certain values before being able to
         // report a sensible preferred height for the specified width.
-        this.emit('allocated-size-changed', allocation.x2 - allocation.x1, allocation.y2 - allocation.y1);
+        this.emit('grid-available-size-changed', gridContainerBox.x2 - gridContainerBox.x1,
+                  gridContainerBox.y2 - gridContainerBox.y1);
```
mariospr: This block would be replaced with a direct call to `adaptToSize()` if you implement my suggestion before

js/ui/iconGrid.js
```diff
@@ -759,7 +759,7 @@ const PaginatedIconGrid = new Lang.Class({
         }
 
         let x = box.x1 + leftEmptySpace + this.leftPadding;
-        let y = box.y1 + this.topPadding;
+        let y = box.y1 + this.topPadding + this._spaceBetweenPages / 2;
```
mariospr: It looks like might work when you have multiple pages, but break things when you only have one, as this spacing is not meant to be used for centering things, but to separate multiple pages. For centering things I believe you'd have to implement a `yAlign` property, similar to the `xAlign` one you can see here.

js/ui/iconGrid.js
```diff
@@ -839,7 +839,7 @@ const PaginatedIconGrid = new Lang.Class({
 
         let firstPageItem = pageNumber * this._childrenPerPage
         let childBox = this._getVisibleChildren()[firstPageItem].get_allocation_box();
-        return childBox.y1 - this.topPadding;
+        return childBox.y1 - this.topPadding - this._spaceBetweenPages / 2;
```
mariospr: Same comment applies here

js/ui/appDisplay.js
```diff
@@ -1191,7 +1191,7 @@ const FolderView = new Lang.Class({
 
     nRowsDisplayedAtOnce: function() {
         let [availWidthPerPage, availHeightPerPage] = this._getPageAvailableSize();
-        let maxRows = this._grid.rowsForHeight(availHeightPerPage) - 1;
```
mariospr: I'm a bit surprised to see that this hasn't been a problem for us now and that simply removing it doesn't cause any regression, but I've tested this branch and it seems to work so I guess it might be fine considering the other downstream changes we made so far.

In any case it would be interesting to try to understand why that `- 1` was there in the first place.

mariospr: @csoriano89 Explained on IRC that this `- 1` is here so that there's always space in the grid to show at leat the "outside row" in the icon grid with the folder's icon plus the opened folder with the icons inside. In other words, if there's enough screen real state for 4 rows and a folder has so many icons that would need 4 or more rows, that `- 1` makes sure the folder only gets allocation for `4 -1 = 3` rows, so that the folder's icon in the desktop does not get shifted up.

So, this is just a problem if we only have space for 1 row, where we'd probably want to sacrifice even the folder's icon and show the folder's contents on top, but I just checked and EOS 3.2 has enough space for 2 rows even at 640x480, and I don't think we need to support resolutions smaller than this (even without any low-res quirk in place), so this change might not be needed after all.

js/ui/viewSelector.js
mariospr: Looks like this wouldn't be that easy after all because we only have access to the actor from the layout manager, so I'm ok with leaving this like this for now. Maybe we can clean things up a bit during the next rebase now we know more stuff (tm)

mariospr:
I'm not entirely convinced by these changes (specially about the last commit, that we shouldn't need -but we do), but considering the time constraints and the fact that they seem to fix the issues at hand without apparently introducing any regressions, I think they are worth a try. Merging

PR #76, opened 2017-06-29
T17245

mcatanzaro:
https://phabricator.endlessm.com/T17245

### DIFF
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -126,6 +126,21 @@ const AuthPrompt = new Lang.Class({
                                         y_fill: true,
                                         y_align: St.Align.START });
 
+        let passwordHintLabel = new St.Label({ text: _("Show password hint"),
+                                               style_class: 'login-dialog-password-recovery-link' });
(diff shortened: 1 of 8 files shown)


js/gdm/authPrompt.js
```diff
@@ -183,10 +229,17 @@ const AuthPrompt = new Lang.Class({
 
         this._entry.clutter_text.connect('text-changed',
                                          Lang.bind(this, function() {
-                                             if (!this._userVerifier.hasPendingMessages)
-                                                 this._fadeOutMessage();
-
-                                             this._updateNextButtonSensitivity(this._entry.text.length > 0);
+                                             if (this._passwordResetCode == null &&
+                                                 !this._displayingPasswordHint) {
```
mcatanzaro: Hm, this needs to be moved to the next conditional down, or entering the correct password after requesting a hint won't work. I'll fix it.

mcatanzaro:
We can optionally dist that file, but I don't think there's much point, because it's only intended for developers. We don't want to install it.

mariospr:
Ok

PR #77, opened 2017-06-29
T17840

joaquimrocha:
https://phabricator.endlessm.com/T17840

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -412,7 +412,7 @@ const AllViewContainer = new Lang.Class({
         this.gridActor = gridActor;
 
         gridActor.y_expand = true;
-        gridActor.y_align = Clutter.ActorAlign.CENTER;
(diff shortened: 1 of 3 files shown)


js/ui/appDisplay.js
```diff
@@ -412,7 +412,7 @@ const AllViewContainer = new Lang.Class({
         this.gridActor = gridActor;
 
         gridActor.y_expand = true;
-        gridActor.y_align = Clutter.ActorAlign.CENTER;
+        gridActor.y_align = Clutter.ActorAlign.START;
```
mariospr: This, combined with the fact that we are allocating all the available height to the grid now will result in a non-centered grid on the screen, which will also push the search entry to the top, being more noticeable the more icons we have on screen. See the screenshots I pasted in the ticket.

mariospr:
Merging after further consideration (we'll merge this now and fix the rows later)

PR #93, opened 2017-07-05
Prevent hiding app main window when speedwagon is gone 

mariospr:
The actual fix for this issue is 8b7a25f7 ("ShellApp: Consider speedwagon windows as 'not interesting' for a given app"), but this PR also proposes a few related commits more that fix issues related to app activation (unveiled after fixing this problem) along with backporting a few (also related) patches that got lost during the rebase process.

https://phabricator.endlessm.com/T17937

### DIFF
--- a/src/shell-window-tracker.c
+++ b/src/shell-window-tracker.c
@@ -887,6 +887,12 @@ shell_window_tracker_is_window_interesting (MetaWindow *window)
   if (meta_window_is_skip_taskbar (window))
     return FALSE;
 
+  /* HACK: see https://github.com/endlessm/eos-shell/issues/548 and
+   * https://github.com/linuxmint/Cinnamon/issues/728
+   */
+  if (g_strcmp0 (meta_window_get_title (window), "JavaEmbeddedFrame") == 0)
+    return FALSE;
+
   return TRUE;
 }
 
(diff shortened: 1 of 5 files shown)


src/shell-window-tracker.c
```diff
@@ -782,6 +792,40 @@ shell_window_tracker_get_app_from_pid (ShellWindowTracker *tracker,
...
+
+void
+_shell_window_tracker_add_child_process_app (ShellWindowTracker *tracker,
+                                             GPid                pid,
+                                             ShellApp           *app)
+{
+  gpointer pid_ptr = GINT_TO_POINTER((int)pid);
+
+  if (g_hash_table_lookup (tracker->launched_pid_to_app,
+                           &pid_ptr))
```
cosimoc: It strikes me that this should be `pid_ptr`, not its address (which is a `void **`) unless I'm missing something

mariospr: Yes, I think you're right: this is wrong. I got it from the previous shell, so I didn't notice. Guess this just worked anyway because the effect is that g_hash_table_insert() would always be called afterwards, but still the right pid would be retrieved from `get_app_from_window_pid()`.

I'll fix that.

PR #94, opened 2017-07-06
Remove GSettings schema override from debian package

mariospr:
We now specify this from the master branch (see commit 6f5e31dd) so
we don't need to do this again via an override. Also, it's not only
that's not needed but also that keeping this in place is harmful,
since it would break set ups like the live installer where we patch
the defaults via other means, as this override file will then be
applied, rendering those other customizations useless.

https://phabricator.endlessm.com/T17622


wjt:
> Also, it's not only
> that's not needed but also that keeping this in place is harmful,
> since it would break set ups like the live installer where we patch
> the defaults via other means, as this override file will then be
> applied, rendering those other customizations useless

This isn't true – this override applies to the systemwide defaults, and https://github.com/endlessm/gnome-initial-setup/blob/master/gnome-initial-setup/pages/live-chooser/eos-setup-live-user applies the necessary changes to the live user's gsettings database.

PR #96, opened 2017-07-12
Fix Jenkins build: use RPATH instead of RUNPATH for ELF binaries

mariospr:
This is a workaround for a change in behaviour in binutils, which we're basically bringing from the debian/rules file right into the autotools sources, with the hope of getting it merged upstream too.

https://phabricator.endlessm.com/T18131

### DIFF
--- a/docs/reference/shell/Makefile.am
+++ b/docs/reference/shell/Makefile.am
@@ -113,7 +113,7 @@ expand_content_files=
 # e.g. GTKDOC_CFLAGS=-I$(top_srcdir) -I$(top_builddir) $(GTK_DEBUG_FLAGS)
 # e.g. GTKDOC_LIBS=$(top_builddir)/gtk/$(gtktargetlib)
 GTKDOC_CFLAGS=$(GNOME_SHELL_CFLAGS)
-GTKDOC_LIBS=$(GNOME_SHELL_LIBS) $(top_builddir)/src/libgnome-shell-menu.la $(top_builddir)/src/libgnome-shell-base.la $(top_builddir)/src/libgnome-shell.la $(top_builddir)/src/libeos-shell-fx.la -rpath $(MUTTER_TYPELIB_DIR)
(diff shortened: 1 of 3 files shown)


dbnicholson:
I found a simpler and more upstreamable fix in https://phabricator.endlessm.com/T18131#424272:
```
diff --git a/src/Makefile.am b/src/Makefile.am
index 4b73234..adb0ae7 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -249,6 +249,7 @@ libeos_shell_fx_la_LIBADD = \
 
 libeos_shell_fx_la_LDFLAGS =   \
        -avoid-version          \
+       -R $(MUTTER_TYPELIB_DIR) \
        $(NULL)
 
 libeos_shell_fx_la_CPPFLAGS =  \
@@ -319,7 +320,7 @@ shell-enum-types.c: $(srcdir)/shell-enum-types.c.in stamp-shell-enum-types.h
        rm -f $(@F).tmp
 EXTRA_DIST += shell-enum-types.c.in
 
-libgnome_shell_ldflags = -avoid-version
+libgnome_shell_ldflags = -avoid-version -R $(MUTTER_TYPELIB_DIR)
 libgnome_shell_libadd =                \
        -lm                     \
        $(GNOME_SHELL_LIBS)     \
```
The reason being the `DT_RUNPATH` requires the libraries themselves to have the path specified and not just the toplevel executable (in the way that `g-ir-scanner` is putting things together).

What do you think?

mariospr:
@dbnicholson I think it makes sense, and I've just tried out your approach and can confirm it works fine, so I'm all for it, specially because it looks more like a real fix and less as a silly workaround.

The only "but" is that, after reading the libtool docs for -R, I think it makes sense to replace every instance of -rpath in the shell's Makefiles (just 6 places) to use -R instead for consistency, and also because -R seems to do the right thing in every single case, regardless of which type of dtags are in use.

PR #99, opened 2017-07-13
search: Activate the app context before the result

joaquimrocha:
The reason for this is that showing the search results in an app can
take some time, and thus its speedwagon should be shown if needed (which
accomplished by activating the app context).

https://phabricator.endlessm.com/T3968

### DIFF
--- a/js/ui/remoteSearch.js
+++ b/js/ui/remoteSearch.js
@@ -7,6 +7,7 @@ const Lang = imports.lang;
 const St = imports.gi.St;
 const Shell = imports.gi.Shell;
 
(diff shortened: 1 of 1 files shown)


smspillaz:
Ah, I actually had something for this in the shell already. Its currently used by the discovery feed. Have a look here: https://github.com/endlessm/eos-discovery-feed/blob/master/src/main.js#L492

smspillaz:
The current approach is certainly more terse, which I like, though there might be a difference between launching the app directly to the result through dbus activation vs launching it then calling the dbus method

PR #101, opened 2017-07-14
remoteSearch: sort providers for apps on the desktop before

mariospr:
This should be squashed with f5dc86c3 in future rebases.

https://phabricator.endlessm.com/T4008


rshuler:
Looks like this was never cherry-picked to eos3.2.  Cleaning that up now.

PR #103, opened 2017-07-17
Fix notifications

erikos:
https://phabricator.endlessm.com/T18310

### DIFF
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -6,8 +6,9 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const Gtk = imports.gi.Gtk;
 const Lang = imports.lang;
-const Shell = imports.gi.Shell;
 const Mainloop = imports.mainloop;
+const Shell = imports.gi.Shell;
+const Signals = imports.signals;
 const St = imports.gi.St;
 
 const Config = imports.misc.config;
@@ -805,6 +806,7 @@ const GtkNotificationDaemon = new Lang.Class({
         source.connect('count-updated', Lang.bind(this, this._saveNotifications));
         Main.messageTray.add(source);
         this._sources[appId] = source;
+        this.emit('new-gtk-notification-source', source);
         return source;
     },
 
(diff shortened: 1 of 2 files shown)


erikos:
"notificationDaemon: Fix a notification deleting itself...": looks like this has been fixed upstream already: https://git.gnome.org/browse/gnome-shell/commit/js/ui/notificationDaemon.js?id=accd24e379374b202791af2b3e065e1fca4fa406

js/ui/notificationDaemon.js
```diff
@@ -740,6 +745,11 @@ const GtkNotificationDaemonAppSource = new Lang.Class({
         else
             this.pushNotification(notification);
 
+        if (oldNotification) {
+            oldNotification.disconnect(oldNotificationDestroyId);
+            oldNotification.destroy();
+        }
+
```
smspillaz: Is there any reason why we can't disconnect and destroy the old notification earlier, before we add the new notification?

js/ui/notificationDaemon.js
```diff
@@ -805,6 +819,7 @@ const GtkNotificationDaemon = new Lang.Class({
         source.connect('count-updated', Lang.bind(this, this._saveNotifications));
         Main.messageTray.add(source);
         this._sources[appId] = source;
+        this.emit('new-gtk-notification-source', source);
```
smspillaz: This is not used

js/ui/notificationDaemon.js
```diff
@@ -762,6 +772,10 @@ const GtkNotificationDaemonAppSource = new Lang.Class({
         }
         return [this._appId, notifications];
     },
+
+    get app() {
+        return this._app;
+    }
```
smspillaz: It does not look like this is used

js/ui/notificationDaemon.js
```diff
@@ -871,7 +886,16 @@ const GtkNotificationDaemon = new Lang.Class({
 
         invocation.return_value(null);
     },
+
+    getSourceFor: function(appId) {
+        if (this._sources[appId]) {
+            return this._sources[appId];
+        }
+        return null;
+    },
```
smspillaz: It does not look like this is used

js/ui/notificationDaemon.js
```diff
@@ -880,4 +904,13 @@ const NotificationDaemon = new Lang.Class({
...
         this._gtkNotificationDaemon = new GtkNotificationDaemon();
     },
+
+    get fdo() {
+        return this._fdoNotificationDaemon;
+    },
+
+    get gtk() {
+        return this._gtkNotificationDaemon;
+    },
```
smspillaz: Now that we are using `messageTray` to check notifications, is it necessary to expose this?

cosimoc:
Superseded by #104.

PR #104, opened 2017-07-18
Desktop icons: add notification information

erikos:
When there are new notifications from the app (e.g. the ChatBox notifies the user regarding certain lessons) a badge is added to the desktop icon.

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -24,6 +24,8 @@ const GrabHelper = imports.ui.grabHelper;
 const IconGrid = imports.ui.iconGrid;
 const IconGridLayout = imports.ui.iconGridLayout;
 const Main = imports.ui.main;
+const MessageTray = imports.ui.messageTray;
+const NotificationDaemon = imports.ui.notificationDaemon;
 const Overview = imports.ui.overview;
 const OverviewControls = imports.ui.overviewControls;
 const PopupMenu = imports.ui.popupMenu;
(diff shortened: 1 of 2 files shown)


js/ui/appDisplay.js
```diff
@@ -2382,7 +2401,10 @@ const AppIcon = new Lang.Class({
         return [sourceActor.actor];
     },
 
-    _onNewGtkNotificationSource: function(daemon, source) {
+    _sourceAdded: function(tray, source) {
+        if (!(source instanceof NotificationDaemon.GtkNotificationDaemonAppSource))
+            return;
```
smspillaz: I don't know if it makes sense to have this limitation. What was the reason for not allowing Fdo notifications?

js/ui/notificationDaemon.js
```diff
@@ -762,6 +764,10 @@ const GtkNotificationDaemonAppSource = new Lang.Class({
         }
         return [this._appId, notifications];
     },
+
+    get app() {
+        return this._app;
+    }
```
smspillaz: Ah, got it, this is used in

    if (source.app != this.app)
              return;

js/ui/appDisplay.js
erikos: Well, we only connect to the signal for the ChatBox case. So I want to filter and bail out as early as possible. If we open it up for more apps we can remove that special casing.

js/ui/notificationDaemon.js
erikos: Right, that is needed for this check.

js/ui/appDisplay.js
smspillaz: Hmm okay. Can you add a comment specifying this? It might be confusing for future maintainers if they see this.

erikos: I added a comment for that one.

PR #106, opened 2017-07-18
main: Add $(pkglibdir) to GIR's list of search path for scanned libraries

mariospr:
This is not strictly necessary when linking the shell with DT_RPATH as
the runtime path will be extracted from the introspected libraries and
used from gobject introspection to find them when loading them based on
the typelib files, but when linking with DT_RUNPATH, as it's the case
for some linkers (e.g. ld.gold, or ld.bfd in some distros like Debian).

By explicitly prepending this directory as a library path for GIR, we're
making sure that the wrapped shared objects can be found in all cases.

https://phabricator.endlessm.com/T18325

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -201,6 +201,13 @@ shell_introspection_init (void)
 
   g_irepository_prepend_search_path (MUTTER_TYPELIB_DIR);
   g_irepository_prepend_search_path (GNOME_SHELL_PKGLIBDIR);
+
+  /* We need to explicitly add the directory where the private libraries are
+   * copied to to the GIR's library path, so that they can be found at runtime
+   * when linking using DT_RUNPATH (instead of DT_RPATH), which is the default
+   * for some linkers (e.g. gold) and in some distros (e.g. Debian).
+   */
+  g_irepository_prepend_library_path (GNOME_SHELL_PKGLIBDIR);
 }
 
 static void


src/main.c
```diff
@@ -201,6 +201,13 @@ shell_introspection_init (void)
 
   g_irepository_prepend_search_path (MUTTER_TYPELIB_DIR);
   g_irepository_prepend_search_path (GNOME_SHELL_PKGLIBDIR);
+
+  /* We need to explicitly add the directory where the private libraries are
+   * copied to to the GIR's library path, so that they can be found at runtime
+   * when linking using DT_RUNPATH (instead of DT_RPATH), which is the default
+   * for some linkers (e.g. gold) and in some distros (e.g. Debian).
+   */
+  g_irepository_prepend_library_path (GNOME_SHELL_PKGLIBDIR);
```
dbnicholson: It probably wouldn't be a bad idea to add the mutter library path, too. I think the issue with mutter is avoided because gnome-shell is actually linked to the mutter libraries through the shell libraries, so the linker already has it through the standard loading mechanisms. However, I think it would be more robust to add the library path for the mutter libraries here, too.

mariospr: Looks good, just prepared a PR to add that in https://github.com/endlessm/gnome-shell/pull/108

PR #110, opened 2017-07-20
appDisplay: Workaround issue for misplaced folder popup's arrow

mariospr:
There's something wrong with the logic that updates the arrow's position
for a previously opened popup that causes the arrow to be drawn in the
wrong place. After some debugging, it seems that the issue happens because
the call to clutter_actor_get_transformed_position() in boxpointer.js:245
is returning wrong values after the folder has been created and opened at
least once in the past.

Sadly, I could not figure out a proper solution for this problem but an
easy enough workaround would be to avoid recycling the AppFolderPopup and
create one each time a folder is open. Note that this does not mean extra
I/O to load the list of apps since that's still loaded once only by the
FolderView instance, so at least it's not as bad as it could still be.

See https://bugzilla.gnome.org/show_bug.cgi?id=785173 for the upstream bug.

https://phabricator.endlessm.com/T17838

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1954,9 +1954,6 @@ const FolderIcon = new Lang.Class({
         this.actor.add_style_class_name('app-folder');
         this.actor.set_child(this.icon.actor);
 
-        // whether we need to update arrow side, position etc.
-        this._popupInvalidated = false;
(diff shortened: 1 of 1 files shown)


mariospr:
Closed in favour of #111

PR #111, opened 2017-07-21
boxPointer: Fix arrow's position the second time a folder is opened

joaquimrocha:
The second time a folder is opened, the arrow of the folder popup will
be shown in the wrong position (as if pushed to the left). This is due
to its transformed position, which is affected depending on the
allocation of the boxPointer's container.

This patch simply queues a repaint when the allocation is changed which
fixes the issue.

https://phabricator.endlessm.com/T17838

### DIFF
--- a/js/ui/boxpointer.js
+++ b/js/ui/boxpointer.js
@@ -234,6 +234,10 @@ const BoxPointer = new Lang.Class({
             this._reposition();
             this._updateFlip();
         }
+
+        // Since the allocation changed, we need to redraw the arrow, otherwise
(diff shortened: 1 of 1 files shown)


cosimoc:
Odd that this does not happen automatically when the border is given the new allocation; in any case it looks like a good solution, and I'd accept this patch over the one in #110. I'll let @mariospr approve this ultimately though.

mariospr:
Looks good to me (and better than my crappy workaround, thanks @joaquimrocha!). Merged

PR #116, opened 2017-08-06
iconGridLayout: Add ReplaceAppIfPossible method to ShellAppStore

smspillaz:
This removes the old icon and inserts the new one in its place. We'll
use this from the app store in order to replace the "Get X" launchers
with the actual application once installed.

https://phabricator.endlessm.com/T18657

### DIFF
--- a/js/ui/iconGridLayout.js
+++ b/js/ui/iconGridLayout.js
@@ -33,6 +33,14 @@ const APPEND_CONFIG_NAME_BASE = 'icon-grid-append';
  */
 const SHELL_APP_REMOVED_EVENT = '683b40a7-cac0-4f9a-994c-4b274693a0a0';
 
+function findInArray(array, func) {
+    for (let i = 0; i < array.length; ++i)
+        if (func(array[i]))
+            return array[i];
+
+    return null;
+}
+
 const IconGridLayout = new Lang.Class({
     Name: 'IconGridLayout',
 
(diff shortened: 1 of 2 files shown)


js/ui/shellDBus.js
```diff
@@ -490,6 +490,10 @@ const AppStoreIface = '<node> \
 <method name="AddAppIfNotVisible"> \
     <arg type="s" direction="in" name="id" /> \
 </method> \
+<method name="ReplaceAppIfPossible"> \
```
cosimoc: This should just be called `ReplaceApplication()`

js/ui/shellDBus.js
```diff
@@ -542,6 +546,21 @@ const AppStoreService = new Lang.Class({
...
+        eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
+
+        // Can't replace a folder
+        if (IconGridLayout.layout.iconIsFolder(originalId))
+            return;
+
+        // We can just replace the app icon directly now,
+        // since the replace operation degenerates to
+        // append if the source icon was not available
+        IconGridLayout.layout.replaceIcon(originalId, replacementId, IconGridLayout.DESKTOP_GRID_ID);
```
cosimoc: The last argument is wrong here; we should replace the icon in the folder where it's currently located instead of always putting it on the desktop.

js/ui/iconGridLayout.js
```diff
@@ -216,6 +216,14 @@ const IconGridLayout = new Lang.Class({
         this.repositionIcon(id, null, folderId);
     },
 
+    // Two operations, first insert the new icon
+    // to the left of the old one, then remove
+    // the old one
+    replaceIcon: function(originalId, replacementId, folderId) {
+        this.repositionIcon(replacementId, originalId, folderId);
+        this.repositionIcon(originalId, null, null);
```
cosimoc: You could use `removeIcon(originalId, false)` to remove the old icon.

js/ui/shellDBus.js
smspillaz: That is a good point. Do you know if there is a way to look up the folder an app is already in?

PR #117, opened 2017-08-07
paginatedIconGrid: consider opened space when calculating dnd

GeorgesStavracas:
When calculating the row that the user is hovering with the dragged
icon, the current implementation of the paginated icon grid does not
take into account the extra space opened for folders, and ends up
miscalculating the target row.

Fix that by storing and using the opened extra space. This is
implemented by an overrided function to keep IconGrid logic
isolated from PaginatedIconGrid details.

https://phabricator.endlessm.com/T18038

### DIFF
--- a/js/ui/iconGrid.js
+++ b/js/ui/iconGrid.js
@@ -800,6 +800,13 @@ const IconGrid = new Lang.Class({
         return -1;
     },
 
+    // This function is overriden by the PaginatedIconGrid subclass so we can
+    // take into account the extra space when dragging from a folder
+    _calculateDndRow: function(y) {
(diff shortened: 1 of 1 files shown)


js/ui/iconGrid.js
```diff
@@ -800,6 +800,13 @@ const IconGrid = new Lang.Class({
         return -1;
     },
 
+    // This function is overriden by the PaginatedIconGrid subclass so we can
+    // take into account the extra space when dragging from a folder
+    _calculateDndRow: function(y) {
+        let rowHeight = this._vItemSize + this._getSpacing();
```
mariospr: Use `_getVItemSize()` instead of `this._vItemSize`

PR #122, opened 2017-08-10
Recreate the user's avatar when scaling factor changes

mariospr:
We need to regenerate the user's avatar in the bottom panel and the
user menu when the scaling factor changes, so that it's rendered
properly according to the factor in effect.

https://phabricator.endlessm.com/T18352

### DIFF
--- a/js/ui/userWidget.js
+++ b/js/ui/userWidget.js
@@ -35,6 +35,10 @@ const Avatar = new Lang.Class({
                                   reactive: params.reactive,
                                   width: this._iconSize * scaleFactor,
                                   height: this._iconSize * scaleFactor });
+
+        // Monitor the scaling factor to make sure we recreate the avatar when needed.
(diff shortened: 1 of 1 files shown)


js/ui/panel.js
```diff
@@ -791,22 +796,37 @@ const UserMenu = new Lang.Class({
...
+        // Monitor the scaling factor to make sure we recreate the button and the
+        // menu showing the Avatar when it changes, so that it's rendered correctly.
+        let themeContext = St.ThemeContext.get_for_stage(global.stage);
+        themeContext.connect('notify::scale-factor', Lang.bind(this, this._redisplay));
+
+        this._redisplay();
+    },
+
+    _redisplay: function() {
+        let scaleFactor = St.ThemeContext.get_for_stage(global.stage).scale_factor;
```
mariospr: I forgot about that, will remove it.

PR #128, opened 2017-08-21
discoveryFeedButton: don't use a hardcoded asset

cosimoc:
Instead, simply draw the semi-circle using CSS and use an icon from the
theme for the down arrow.
This will ensure crisp rendering at all resolutions, including HiDpi.

https://phabricator.endlessm.com/T18850


data/theme/gnome-shell.css
```diff
@@ -1883,11 +1883,29 @@ stage {
...
-  width: 44px;
-  height: 23px;
-  background-image: url(resource:///org/gnome/shell/theme/discovery-feed-open-tab.png); }
+    width: 44px;
+    height: 22px;
+    border-radius: 0px 0px 22px 22px;
+
+    background-gradient-start: rgba(255,95,42,0.9);
+    background-gradient-end: rgba(248,63,0,0.9);
+    background-gradient-direction: vertical;
```
smspillaz: Nit: There should be spaces between `,` and the numbers here.

data/theme/gnome-shell.css
```diff
@@ -1883,11 +1883,29 @@ stage {
...
+
+    box-shadow: inset 0 -1px 1px rgba(0,0,0,0.25);
+
+    color: #eeeeec;
+}
+
 .discovery-feed-button:hover {
-  background-image: url(resource:///org/gnome/shell/theme/discovery-feed-open-tab_hover.png); }
+    background-gradient-start: rgba(255,95,42,1.0);
+    background-gradient-end: rgba(248,63,0,1.0);
```
smspillaz: Same applies here

data/theme/gnome-shell.css
```diff
@@ -1883,11 +1883,29 @@ stage {
...
+
+    color: white;
```
smspillaz: I'm not sure if this arrow is meant to be a "perfect" white - my colour sampler says it is `#dedede`

cosimoc: It's supposed to be white only on hover. See also the "search icon" theming further down in the file.

cosimoc:
@smspillaz Comments addressed

data/theme/gnome-shell.css
smspillaz: Cool, makes sense!

PR #129, opened 2017-08-22
Adapt to the new filenames and versions of mutter and its private libraries

mariospr:
In mutter's commit 4ebc55f2 libmutter.so and the mutter-*.so libraries
were made parallel installable [1] which, among other things, impacted
the filenames of the shared objects and .pc files that get installed.

GNOME Shell 3.24 is already adapted to this situation (see [2]), so
we need this small fix here to keep our fork of GNOME Shell 3.22
building and running find with the newer version of mutter.

[1] https://git.gnome.org/browse/mutter/commit/?id=4ebc55f2
[2] https://git.gnome.org/browse/gnome-shell/commit/?id=a46ea3f8

https://phabricator.endlessm.com/T18787

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,7 +53,7 @@ if $PKG_CONFIG --exists gstreamer-1.0 '>=' $GSTREAMER_MIN_VERSION ; then
    AC_MSG_RESULT(yes)
    build_recorder=true
    recorder_modules="gstreamer-1.0 gstreamer-base-1.0 x11 gtk+-3.0"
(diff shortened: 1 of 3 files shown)


cosimoc:
This will need another PR for the debian branch to depend on the new version of mutter packages.

mariospr:
> This will need another PR for the debian branch to depend on the new version of mutter packages.

Right, I missed that because I was testing locally, will add it

PR #144, opened 2017-09-21
Switch the keyboard layout to 'us' for the polkitAgent pw entry

erikos:
We switch back to the layout used before when the dialog
closes. The change is reflected in the layout switcher in
the taskbar.

https://phabricator.endlessm.com/T18598

### DIFF
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -15,6 +15,7 @@ const Animation = imports.ui.animation;
 const Batch = imports.gdm.batch;
 const Config = imports.misc.config;
 const GdmUtil = imports.gdm.util;
+const InputSourceManager = imports.ui.status.keyboard;
 const Params = imports.misc.params;
 const ShellEntry = imports.ui.shellEntry;
 const Tweener = imports.ui.tweener;
@@ -115,6 +116,10 @@ const AuthPrompt = new Lang.Class({
                          y_fill: false,
                          x_align: St.Align.START });
 
+        // for setting passwords we switch to a latin keyboard layout
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceManager.passwordModeEnabled = true;
+
         this._entry.grab_key_focus();
 
         this._message = new St.Label({ opacity: 0,
@@ -183,6 +188,7 @@ const AuthPrompt = new Lang.Class({
     _onDestroy: function() {
         this._userVerifier.destroy();
         this._userVerifier = null;
+        this._inputSourceManager.passwordModeEnabled = false;
     },
 
     _initButtons: function() {
--- a/js/misc/keyboardManager.js
+++ b/js/misc/keyboardManager.js
@@ -4,6 +4,7 @@ const GLib = imports.gi.GLib;
 const GnomeDesktop = imports.gi.GnomeDesktop;
 const Lang = imports.lang;
 const Meta = imports.gi.Meta;
+const Shell = imports.gi.Shell;
 
 const Main = imports.ui.main;
 
@@ -150,5 +151,15 @@ const KeyboardManager = new Lang.Class({
     _buildOptionsString: function() {
         let options = this._xkbOptions.join(',');
         return options;
+    },
+
+    isLatinLayout: function(id) {
+        let info = this._layoutInfos[id];
+        if (!info)
+            return false;
+        let options = this._buildOptionsString();
+        let [_layouts, variants] = this._buildGroupStrings(info.group);
+        let [_found, , , layout, _variant] = this._xkbInfo.get_layout_info(id);
+        return !Shell.util_needs_secondary_layout(layout, variants, options);
     }
 });
--- a/js/ui/components/networkAgent.js
+++ b/js/ui/components/networkAgent.js
@@ -12,6 +12,8 @@ const Shell = imports.gi.Shell;
 const St = imports.gi.St;
 
 const Config = imports.misc.config;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
+const InputSourceManager = imports.ui.status.keyboard;
 const Main = imports.ui.main;
 const MessageTray = imports.ui.messageTray;
 const ModalDialog = imports.ui.modalDialog;
@@ -37,6 +39,12 @@ const NetworkSecretDialog = new Lang.Class({
             this._content = contentOverride;
         else
             this._content = this._getContent();
+
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        let manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        manager.addMenu(this._inputSourceIndicator.menu);
+        this._inputSourceManager.passwordModeEnabled = true;
 
         let mainContentBox = new St.BoxLayout({ style_class: 'prompt-dialog-main-layout',
                                                 vertical: false });
@@ -123,11 +131,18 @@ const NetworkSecretDialog = new Lang.Class({
                 secret.valid = true;
 
             if (rtl) {
-                layout.attach(secret.entry, 0, pos, 1, 1);
-                layout.attach(label, 1, pos, 1, 1);
+                let j = 0;
+                if (i === 0) {
+                    layout.attach(this._inputSourceIndicator.container, j, pos, 1, 1);
+                    j++;
+                }
+                layout.attach(secret.entry, j+1, pos, 1, 1);
+                layout.attach(label, j+2, pos, 1, 1);
             } else {
                 layout.attach(label, 0, pos, 1, 1);
                 layout.attach(secret.entry, 1, pos, 1, 1);
+                if (i === 0)
+                    layout.attach(this._inputSourceIndicator.container, 2, pos, 1, 1);
             }
             pos++;
 
@@ -171,6 +186,8 @@ const NetworkSecretDialog = new Lang.Class({
                 this._agent.set_password(this._requestId, secret.key, secret.value);
         }
 
+        this._inputSourceManager.passwordModeEnabled = false;
+
         if (valid) {
             this._agent.respond(this._requestId, Shell.NetworkAgentResponse.CONFIRMED);
             this.close(global.get_current_time());
@@ -179,6 +196,8 @@ const NetworkSecretDialog = new Lang.Class({
     },
 
     cancel: function() {
+        this._inputSourceManager.passwordModeEnabled = false;
+
         this._agent.respond(this._requestId, Shell.NetworkAgentResponse.USER_CANCELED);
         this.close(global.get_current_time());
     },
--- a/js/ui/components/polkitAgent.js
+++ b/js/ui/components/polkitAgent.js
@@ -15,7 +15,10 @@ const PolkitAgent = imports.gi.PolkitAgent;
 
 const Animation = imports.ui.animation;
 const Components = imports.ui.components;
+const InputSourceManager = imports.ui.status.keyboard;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
 const ModalDialog = imports.ui.modalDialog;
+const PopupMenu = imports.ui.popupMenu;
 const ShellEntry = imports.ui.shellEntry;
 const UserWidget = imports.ui.userWidget;
 const Tweener = imports.ui.tweener;
@@ -149,6 +152,12 @@ const AuthenticationDialog = new Lang.Class({
         this._passwordEntry.clutter_text.connect('activate', Lang.bind(this, this._onEntryActivate));
         this._passwordBox.add(this._passwordEntry,
                               { expand: true });
+
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._passwordBox.add(this._inputSourceIndicator.container);
+        let manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        manager.addMenu(this._inputSourceIndicator.menu);
 
         let spinnerIcon = Gio.File.new_for_uri('resource:///org/gnome/shell/theme/process-working.svg');
         this._workSpinner = new Animation.AnimatedIcon(spinnerIcon, WORK_SPINNER_ICON_SIZE);
@@ -336,6 +345,7 @@ const AuthenticationDialog = new Lang.Class({
         else
             this._passwordEntry.clutter_text.set_password_char('\u25cf'); // ● U+25CF BLACK CIRCLE
 
+        this._inputSourceManager.passwordModeEnabled = true;
         this._passwordBox.show();
         this._passwordEntry.set_text('');
         this._passwordEntry.grab_key_focus();
@@ -363,6 +373,7 @@ const AuthenticationDialog = new Lang.Class({
 
     destroySession: function() {
         if (this._session) {
+            this._inputSourceManager.passwordModeEnabled = false;
             if (!this._completed)
                 this._session.cancel();
             this._completed = false;
--- a/js/ui/shellMountOperation.js
+++ b/js/ui/shellMountOperation.js
@@ -11,10 +11,13 @@ const St = imports.gi.St;
 const Shell = imports.gi.Shell;
 
 const CheckBox = imports.ui.checkBox;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
+const InputSourceManager = imports.ui.status.keyboard;
 const Main = imports.ui.main;
 const MessageTray = imports.ui.messageTray;
 const ModalDialog = imports.ui.modalDialog;
 const Params = imports.misc.params;
+const PopupMenu = imports.ui.popupMenu;
 const ShellEntry = imports.ui.shellEntry;
 
 const LIST_ITEM_ICON_SIZE = 48;
@@ -379,6 +382,13 @@ const ShellMountPasswordDialog = new Lang.Class({
         this._passwordBox.add(this._passwordEntry, {expand: true });
         this.setInitialKeyFocus(this._passwordEntry);
 
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._passwordBox.add(this._inputSourceIndicator.container);
+        let manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        manager.addMenu(this._inputSourceIndicator.menu);
+        this._inputSourceManager.passwordModeEnabled = true;
+
         this._errorMessageLabel = new St.Label({ style_class: 'prompt-dialog-error-label',
                                                  text: _("Sorry, that didn’t work. Please try again.") });
         this._errorMessageLabel.clutter_text.ellipsize = Pango.EllipsizeMode.NONE;
@@ -414,11 +424,13 @@ const ShellMountPasswordDialog = new Lang.Class({
     },
 
     _onCancelButton: function() {
+        this._inputSourceManager.passwordModeEnabled = false;
         this.emit('response', -1, '', false);
     },
 
     _onUnlockButton: function() {
         this._onEntryActivate();
+        this._inputSourceManager.passwordModeEnabled = false;
     },
 
     _onEntryActivate: function() {
--- a/js/ui/status/keyboard.js
+++ b/js/ui/status/keyboard.js
@@ -317,6 +317,9 @@ const InputSourceManager = new Lang.Class({
         this._ibusSources = {};
 
         this._currentSource = null;
+        this._lastSource = null;
+        this._passwordModeEnabled = false;
+        this._fallbackInputSource = null;
 
         // All valid input sources currently in the gsettings
         // KEY_INPUT_SOURCES list ordered by most recently used
@@ -480,6 +483,51 @@ const InputSourceManager = new Lang.Class({
             this._updateMruSettings();
     },
 
+    get passwordModeEnabled() {
+        return this._passwordModeEnabled;
+    },
+
+    set passwordModeEnabled(enable) {
+        if (this._passwordModeEnabled === enable)
+            return;
+
+        this._passwordModeEnabled = enable;
+        this._syncPasswordMode();
+    },
+
+    _syncPasswordMode: function() {
+        let source;
+
+        if (this._passwordModeEnabled) {
+            // do not switch the keyboard layout if the current one is a latin one
+            if (this._keyboardManager.isLatinLayout(this._currentSource.id))
+                return;
+
+            // look for latin keyboard layout to switch to
+            for (let i = 0; i < this._mruSources.length; ++i) {
+                let src = this._mruSources[i];
+                if (this._keyboardManager.isLatinLayout(src.id)) {
+                    source = src;
+                    break;
+                }
+            }
+            if (!source) {
+                // fallback to an 'en' keyboard layout
+                this._updateInputSources(true);
+                source = this._fallbackInputSource;
+            }
+            this._lastSource = this._currentSource;
+            this._activateInputSource(source, false);
+        } else {
+            if (!this._lastSource)
+                return;
+            source = this._lastSource;
+            this._activateInputSource(source, false);
+            this._lastSource = null;
+            this._updateInputSources(false);
+        }
+    },
+
     _updateMruSources: function() {
         let sourcesList = [];
         for (let i in this._inputSources)
@@ -526,6 +574,10 @@ const InputSourceManager = new Lang.Class({
     },
 
     _inputSourcesChanged: function() {
+        this._updateInputSources(false);
+    },
+
+    _updateInputSources: function(needsFallbackSource = false) {
         let sources = this._settings.inputSources;
         let nSources = sources.length;
 
@@ -563,7 +615,7 @@ const InputSourceManager = new Lang.Class({
                 infosList.push({ type: type, id: id, displayName: displayName, shortName: shortName });
         }
 
-        if (infosList.length == 0) {
+        if (infosList.length == 0 || needsFallbackSource) {
             let type = INPUT_SOURCE_TYPE_XKB;
             let id = KeyboardManager.DEFAULT_LAYOUT;
             let [ , displayName, shortName, , ] = this._xkbInfo.get_layout_info(id);
@@ -588,6 +640,11 @@ const InputSourceManager = new Lang.Class({
             if (is.type == INPUT_SOURCE_TYPE_IBUS)
                 this._ibusSources[is.id] = is;
         }
+
+        if (needsFallbackSource)
+            this._fallbackInputSource = this._inputSources[infosList.length - 1];
+        else
+            this._fallbackInputSource = null;
 
         for (let i in this._inputSources) {
             let is = this._inputSources[i];
@@ -778,11 +835,12 @@ const InputSourceIndicator = new Lang.Class({
     Name: 'InputSourceIndicator',
     Extends: PanelMenu.Button,
 
-    _init: function() {
+    _init: function(parent, showLayout = true) {
         this.parent(0.0, _("Keyboard"));
 
         this._menuItems = {};
         this._indicatorLabels = {};
+        this._showLayoutItem = null;
 
         this._container = new Shell.GenericContainer();
         this._container.connect('get-preferred-width', Lang.bind(this, this._containerGetPreferredWidth));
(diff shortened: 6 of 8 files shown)


js/ui/status/keyboard.js
```diff
@@ -480,6 +481,31 @@ const InputSourceManager = new Lang.Class({
             this._updateMruSettings();
     },
 
+    passwordMode: function(enable) {
+        let source;
+        if (enable) {
+            for (let i = 0; i < this._mruSources.length; ++i) {
+                let src = this._mruSources[i];
+                if (src.id === 'us') {
```
wjt: I don't think this check is correct. Is `us` really what we want here? I use the `gb` layout. This change would silently swap `"` and `@` and turn `#` into `\`. I accept that these characters are not common in wifi passwords but in that case there is no reason to switch from `gb` to `us` layout in this dialogue. (I actually use `gb dvorak` so this would be extra infuriating but this is a niche case.) The `de` case is even worse, right? You don't want to swap `z` and `y` when entering wifi passwords…

I think the behaviour we actually want is to find the most recently-used keyboard layout that can enter Latin characters. Sure, if there isn't one, fall back to picking the `us` layout.

wjt: Ah, I see this is a proof-of-concept :)

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,41 @@ const InputSourceManager = new Lang.Class({
...
+            }
+
+            // look for latin keyboard layout to switch to
+            for (let i = 0; i < this._mruSources.length; ++i) {
+                let src = this._mruSources[i];
+                if (this._keyboardManager.is_latin_layout(src.id)) {
+                    source = src;
+                    break;
+                }
+            }
```
smspillaz: Assuming that we have ES6 suport in Gjs, you can express this as

    source = this._mruSources.find(src => this._keyboardManager.is_latinlayout(src.id)) || new InputSource('xkb', 'us', 'English (US)', 'us', 0);

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+              const char         *layouts,
+              const char         *variants,
+              const char         *options)
+{
+    struct xkb_keymap *keymap;
+    struct xkb_rule_names rmlvo = {
+        .rules = "evdev",
+        .model = "pc104",
+        .layout = layouts,
+	.variant = variants,
```
smspillaz: This seems to be out of alignment

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+        .options = options
+    };
+
+    if (!layouts && !variants && !options)
```
smspillaz: I wonder if this should be `!layouts || !variants || !options` - seems strange that only one of them needs to be set to use values from all of them.

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+            {
+              state->required_keysyms_found[keysym_index] = TRUE;
+              state->n_required_keysyms--;
+            }
+        }
+    }
+}
+
+static struct xkb_keymap *
+create_keymap(struct xkb_context *context,
```
smspillaz: No space before paren

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+    FALSE, /* XKB_KEY_r */
+    FALSE, /* XKB_KEY_s */
+    FALSE, /* XKB_KEY_t */
+    FALSE, /* XKB_KEY_u */
+    FALSE, /* XKB_KEY_v */
+    FALSE, /* XKB_KEY_w */
+    FALSE, /* XKB_KEY_x */
+    FALSE, /* XKB_KEY_y */
+    FALSE, /* XKB_KEY_z */
+  };
```
smspillaz: It feels like a shorter way of doing this is to just create an array of size 24 and memset it all to FALSE:

    gboolean required_keysyms_found[24];
    memset (required_keysyms_found, FALSE, sizeof (gboolean) * 24);

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+    .required_keysyms_found = required_keysyms_found,
+    .n_required_keysyms = G_N_ELEMENTS (required_keysyms_found),
+  };
+  struct xkb_keymap *keymap;
+  struct xkb_context *ctx;
+  enum xkb_context_flags ctx_flags;
+
+  ctx_flags = XKB_CONTEXT_NO_DEFAULT_INCLUDES;
+  ctx = xkb_context_new (ctx_flags);
+  xkb_context_include_path_append_default (ctx);
```
smspillaz: If you adjust the ordering of the declarations above, you can declare and initialise these as the same time.

smspillaz: (That is unless the `create_keymap` call has a dependency on `xkb_context_include_path_append_default`)

smspillaz: (At which point my personal inclination would be to wrap both `xkb_context_new` and that ^ in another function and return the created context).

src/shell-util.c
```diff
@@ -492,3 +499,117 @@ shell_util_composite_capture_images (ClutterCapture  *captures,
...
+static struct xkb_keymap *
+create_keymap(struct xkb_context *context,
+              const char         *layouts,
+              const char         *variants,
+              const char         *options)
+{
+    struct xkb_keymap *keymap;
+    struct xkb_rule_names rmlvo = {
+        .rules = "evdev",
+        .model = "pc104",
```
cosimoc: Is it a good idea for these two strings to be hardcoded? You could use `#defines` if so.

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,41 @@ const InputSourceManager = new Lang.Class({
             this._updateMruSettings();
     },
 
+    passwordMode: function(enable) {
```
cosimoc: You could probably make this a property instead of a method

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,41 @@ const InputSourceManager = new Lang.Class({
...
+        let source;
+
+        if (enable) {
+            // do not switch the leyboard layout if the current one is a latin one
+            if (this._keyboardManager.is_latin_layout(this._currentSource.id)) {
+                return;
+            }
```
cosimoc: Remove curly braces for single-line statement

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,41 @@ const InputSourceManager = new Lang.Class({
             this._updateMruSettings();
     },
 
+    passwordMode: function(enable) {
+        let source;
+
+        if (enable) {
+            // do not switch the leyboard layout if the current one is a latin one
```
cosimoc: Typo: `leyboard`

js/misc/keyboardManager.js
```diff
@@ -150,5 +151,14 @@ const KeyboardManager = new Lang.Class({
     _buildOptionsString: function() {
         let options = this._xkbOptions.join(',');
         return options;
+    },
+
+    is_latin_layout: function(id) {
```
cosimoc: Please use camel case for method names

js/misc/keyboardManager.js
```diff
@@ -150,5 +151,14 @@ const KeyboardManager = new Lang.Class({
...
+        let info = this._layoutInfos[id];
+        if (!info)
+            return;
```
cosimoc: Should `return false`

js/gdm/authPrompt.js
```diff
@@ -19,6 +19,7 @@ const Params = imports.misc.params;
 const ShellEntry = imports.ui.shellEntry;
 const Tweener = imports.ui.tweener;
 const UserWidget = imports.ui.userWidget;
+const InputSourceManager = imports.ui.status.keyboard;
```
cosimoc: Alphabetical order

src/shell-util.c
erikos: In Mutter they are defined as evdev and pc105(pc105+inet).

https://github.com/GNOME/mutter/blob/master/src/backends/meta-backend-private.h#L48

https://github.com/GNOME/mutter/blob/master/clutter/clutter/evdev/clutter-device-manager-evdev.c#L1970

js/ui/status/keyboard.js
erikos: To me this is not a property. This is switching the keyboard layout to a latin one and keeps track of the current layout to be able to switch back. Maybe the method should be named differently? "switchKeyboardLayout(['default'/'password'])"

cosimoc: I'd be OK to rename the method to `setPasswordMode()`; but that would be equivalent to having a `passwordMode` property and setting it to `true`. `setPasswordLayout()` could also work.

js/ui/status/keyboard.js
```diff
@@ -801,8 +837,10 @@ const InputSourceIndicator = new Lang.Class({
         this.menu.addMenuItem(this._propSection);
         this._propSection.actor.hide();
 
-        this.menu.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
-        this._showLayoutItem = this.menu.addAction(_("Show Keyboard Layout"), Lang.bind(this, this._showLayout));
+        if(showLayout) {
```
cosimoc: Missing space

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,40 @@ const InputSourceManager = new Lang.Class({
...
+                                         'us',
+                                         'English (US)',
+                                         'us',
+                                         0);
+            this._lastSource = this._currentSource;
+        } else {
+            if (!this._lastSource)
+                return;
+            source = this._lastSource;
+            this._lastSource = null;
```
cosimoc: Do we need to remove the fallback `us` here source if we added it before?

js/ui/components/polkitAgent.js
```diff
@@ -150,6 +153,16 @@ const AuthenticationDialog = new Lang.Class({
         this._passwordBox.add(this._passwordEntry,
                               { expand: true });
 
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._passwordBox.add(this._inputSourceIndicator.container);
+        this._manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        this._manager.addMenu(this._inputSourceIndicator.menu);
+        this._inputSourceIndicator.connect('destroy', Lang.bind(this, function() {
+            this._inputSourceIndicato.container.destroy();
```
cosimoc: Typo: `Indicato`

js/ui/status/keyboard.js
erikos: Yes, actually that code was a bit uncomplete, I had to add it to the list of layouts as well and signal that change in order to be picked up.

js/ui/status/keyboard.js
```diff
@@ -778,11 +821,12 @@ const InputSourceIndicator = new Lang.Class({
     Name: 'InputSourceIndicator',
     Extends: PanelMenu.Button,
 
-    _init: function() {
+    _init: function(parent, showLayout = true) {
```
cosimoc: Don't think you need to pass the parent here -- it looks unused.

js/ui/status/keyboard.js
```diff
@@ -480,6 +481,48 @@ const InputSourceManager = new Lang.Class({
...
+                    break;
+                }
+            }
+            if (!source) {
+                // fallback to an us keyboard layout
+                source = new InputSource('xkb',
+                                         'us',
+                                         'English (US)',
+                                         'en',
+                                         Object.keys(this._inputSources).length);
```
cosimoc: Instead of doing all this work here (updating mru sources, emitting signals...) I'd save this new source as e.g. `this._fallbackInputSource`, then call `this._inputSourcesChanged()` and add code in there that adds it to `this._inputSources` when set.
In other words, the only place that changes `this._inputSources` should be `this._inputSourcesChanged()`.

js/ui/status/keyboard.js
erikos: It is unused here indeed, but we do pass it when instantiating for the panel: https://github.com/endlessm/gnome-shell/blob/master/js/ui/panel.js#L1190

If I want to keep my default argument I have to add it I think.

cosimoc: Sounds good!

js/ui/status/keyboard.js
erikos: Much nicer indeed! Thanks for the hint, adjusted accordingly.

js/ui/components/networkAgent.js
```diff
@@ -38,6 +39,9 @@ const NetworkSecretDialog = new Lang.Class({
         else
             this._content = this._getContent();
 
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceManager.passwordModeEnabled = true;
```
cosimoc: Are we not packing the switching indicator here?

js/ui/components/polkitAgent.js
```diff
@@ -150,6 +153,16 @@ const AuthenticationDialog = new Lang.Class({
         this._passwordBox.add(this._passwordEntry,
                               { expand: true });
 
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
```
cosimoc: The indicator already has a reference to the input source manager (https://github.com/endlessm/gnome-shell/blob/cf9b988ef5cd9483eeadbc929f1451d85424b2f3/js/ui/status/keyboard.js#L829).

js/ui/components/polkitAgent.js
```diff
@@ -150,6 +153,16 @@ const AuthenticationDialog = new Lang.Class({
...
+        this._passwordBox.add(this._inputSourceIndicator.container);
```
cosimoc: Isn't the container already added to another actor, in the indicator code itself? Any reason not to use `this._inputSourceIndicator.actor` instead?

js/ui/status/keyboard.js
```diff
@@ -480,6 +483,53 @@ const InputSourceManager = new Lang.Class({
...
+                this._inputSourcesChanged();
+            }
+            this._lastSource = this._currentSource;
+        } else {
+            if (!this._lastSource)
+                return;
+            source = this._lastSource;
+            this._lastSource = null;
+            this._fallbackInputSource = null;
+            this._inputSourcesChanged();
```
cosimoc: Since this is common code, I'd move it to the end.

js/ui/components/polkitAgent.js
erikos: The actor has a parent already:

Attempting to add actor of type 'ShellGenericContainer' to a container of type 'StBoxLayout', but the actor has already a parent of type 'StBin'.

js/ui/status/keyboard.js
erikos: Hmm, the this._inputSourcesChanged(); should not happen in all the cases.

js/ui/status/keyboard.js
```diff
@@ -480,6 +483,53 @@ const InputSourceManager = new Lang.Class({
...
+    get passwordModeEnabled() {
+        return this._passwordModeEnabled;
+    },
+
+    set passwordModeEnabled(enable) {
+        this._passwordModeEnabled = enable;
+        this._switchToLatinLayout();
+    },
+
+    _switchToLatinLayout: function() {
```
cosimoc: This function does not actually always switch to latin layout; it will also switch back to the old layout in case password mode is disabled.
I would improve it this way:
```
set passwordModeEnabled(enable) {
  if (enable == this._passwordEnabled)
    return;

  this._syncPasswordMode();
},

_syncPasswordMode: function() {
  // the code that is now in _switchToLatinLayout()
}
```

js/ui/status/keyboard.js
cosimoc: I would go one step further, and move the creation of this source also in `this._inputSourceChanged()`; you could simply add/clear a flag in this function that is e.g. `this._needsFallbackSource`. That way you would not need to look at the length of `this._inputSources` keys from here.

js/ui/status/keyboard.js
cosimoc: I am suggesting that you put it right before the call to `this._activateInputSource()`. That should be equivalent to the current implementation.

js/ui/components/networkAgent.js
```diff
@@ -123,11 +130,21 @@ const NetworkSecretDialog = new Lang.Class({
...
-                layout.attach(secret.entry, 0, pos, 1, 1);
-                layout.attach(label, 1, pos, 1, 1);
+                let i = 0;
+                if (!this._inputSourceIndicator) {
+                    this._addKbdLayoutSwitcher();
+                    layout.attach(this._inputSourceIndicator.container, i, pos, 1, 1);
+                    i++;
+                }
+                layout.attach(secret.entry, i, pos, 1, 1);
+                layout.attach(label, i, pos, 1, 1);
```
cosimoc: Should this not be `i + 1`?

js/ui/components/networkAgent.js
```diff
@@ -123,11 +130,21 @@ const NetworkSecretDialog = new Lang.Class({
                 secret.valid = true;
 
             if (rtl) {
-                layout.attach(secret.entry, 0, pos, 1, 1);
-                layout.attach(label, 1, pos, 1, 1);
+                let i = 0;
+                if (!this._inputSourceIndicator) {
+                    this._addKbdLayoutSwitcher();
```
cosimoc: I would simply call `this._addKbdLayoutSwitcher()` (or even just run the code that is now inside that function) before this loop, and remove the conditionals in here.

js/ui/components/networkAgent.js
```diff
@@ -151,6 +168,16 @@ const NetworkSecretDialog = new Lang.Class({
         this._updateOkButton();
     },
 
+    _addKbdLayoutSwitcher: function() {
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        this._manager.addMenu(this._inputSourceIndicator.menu);
```
cosimoc: The `PopupMenuManager` is never used outside of this function, so you should not need to save it as a private member in `this._manager`

js/ui/components/networkAgent.js
```diff
@@ -151,6 +168,16 @@ const NetworkSecretDialog = new Lang.Class({
...
+        this._inputSourceIndicator.connect('destroy', Lang.bind(this, function() {
+            this._inputSourceIndicato.container.destroy();
```
cosimoc: Typo: `inputSourceIndicato`

js/ui/components/networkAgent.js
```diff
@@ -151,6 +168,16 @@ const NetworkSecretDialog = new Lang.Class({
         this._updateOkButton();
     },
 
+    _addKbdLayoutSwitcher: function() {
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
+        this._manager.addMenu(this._inputSourceIndicator.menu);
+        this._inputSourceIndicator.connect('destroy', Lang.bind(this, function() {
```
cosimoc: I don't think this is really necessary -- I'd expect the container to also be destroyed when the parent actor is destroyed... The real issue here I can see is that we're "abusing" a bit the container/panel menu hierarchy, but that's probably fine...

js/ui/components/polkitAgent.js
```diff
@@ -150,6 +153,16 @@ const AuthenticationDialog = new Lang.Class({
...
+        this._manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
```
cosimoc: See comment above about the popup menu manager not being needed outside of this function

js/ui/components/polkitAgent.js
```diff
@@ -150,6 +153,16 @@ const AuthenticationDialog = new Lang.Class({
...
+        this._manager.addMenu(this._inputSourceIndicator.menu);
+        this._inputSourceIndicator.connect('destroy', Lang.bind(this, function() {
```
cosimoc: See other comment about this possibly not being needed

js/ui/components/polkitAgent.js
```diff
@@ -363,6 +377,9 @@ const AuthenticationDialog = new Lang.Class({
 
     destroySession: function() {
         if (this._session) {
+            this._inputSourceIndicator.menu.close();
```
cosimoc: Can we get to this point with the menu open?

js/ui/components/networkAgent.js
erikos: That one has been added to the network dialog.

js/ui/status/keyboard.js
erikos: Doing that would also call "this._inputSourcesChanged()" for the case where we have a latin layout already available and we can switch to it. It should only be called when we add the "en" layout or remove it.

js/ui/status/keyboard.js
erikos: I have renamed the function to _syncPasswordMode. So the early return you have above does not work as I have to clean up things when switch back to the "original" source.

js/ui/components/networkAgent.js
erikos: Fixed.

js/ui/components/networkAgent.js
erikos: Yes, fixed. Thanks for the catch.

js/ui/components/networkAgent.js
erikos: Yeah, I thought about that one too. I used "this._inputSourceIndicator" to know if we already have added a switcher in the case when we have n entry fields. I am using the loop variable "i" now to only add one switcher for n pw entries.

js/ui/components/polkitAgent.js
cosimoc: Still needs to be addressed

js/ui/components/polkitAgent.js
cosimoc: Possibly still relevant

js/ui/components/polkitAgent.js
cosimoc: Still relevant

js/ui/shellMountOperation.js
```diff
@@ -379,6 +385,11 @@ const ShellMountPasswordDialog = new Lang.Class({
         this._passwordBox.add(this._passwordEntry, {expand: true });
         this.setInitialKeyFocus(this._passwordEntry);
 
+        this._inputSourceIndicator = new InputSourceIndicator(this, false);
+        this._passwordBox.add(this._inputSourceIndicator.container);
+        manager = new PopupMenu.PopupMenuManager({ actor: this._inputSourceIndicator.container });
```
cosimoc: Missing `let`

js/ui/status/keyboard.js
```diff
@@ -527,6 +576,14 @@ const InputSourceManager = new Lang.Class({
 
     _inputSourcesChanged: function() {
         let sources = this._settings.inputSources;
+        if (this._needsFallbackSource) {
```
cosimoc: I think it would be possible and cleaner to do this together with the `infosList.length == 0` case below. It looks like the same input source would be created in that case.

js/ui/status/keyboard.js
```diff
@@ -480,6 +484,51 @@ const InputSourceManager = new Lang.Class({
...
+            this._lastSource = this._currentSource;
+            this._activateInputSource(source, false);
+        } else {
+            if (!this._lastSource)
+                return;
+            source = this._lastSource;
+            this._activateInputSource(source, false);
+            this._lastSource = null;
+            this._needsFallbackSource = false;
+            this._fallbackInputSource = null;
```
cosimoc: Since `this._inputSourcesChanged()` sets `this._fallbackInputSource` when `this._needsFallbackSource == true`, it should also unset it when `this._needsFallbackSource == false`. On a related note, you could also pass directly `needsFallbackSoruce` to `this._inputSourcesChanged()` instead of saving it as a private member. Or alternatively, if `this._inputSourcesChanged()` is called from too many places to make that practical, you can move the body of that function to another method that takes the parameter, and call the new method from both places.

js/ui/status/keyboard.js
```diff
@@ -480,6 +484,51 @@ const InputSourceManager = new Lang.Class({
             this._updateMruSettings();
     },
 
+    get passwordModeEnabled() {
+        return this._passwordModeEnabled;
+    },
+
+    set passwordModeEnabled(enable) {
+        this._passwordModeEnabled = enable;
```
cosimoc: I don't see why you would need to call `this._syncPasswordMode()` if `enable == this._passwordModeEnabled` already.

erikos: We do need to reset the kbd layout to the previous one. 

passwordModeEnabled(true):
- store the current kbd layout
- lookup a possible latin layout and switch to it, or switch to 'en'

passwordModeEnabled(false):
- switch back to stored kbd layout
- reset flags etc

cosimoc: I understand that, but I don't think you need to do any of that work when setting the same state again, do you?

erikos: Eeh, yeah - sorry for misreading. I have it now.

js/ui/status/keyboard.js
erikos: Good catch, I merged the passwordFallbackCreation code into that one. I store the source information into this._fallbackInputSource so we can activate it from _syncPasswordMode. I was wondering if I should store just the index but then I would need to access this._inputSources directly outside of inputŚourcesChanged.

js/ui/status/keyboard.js
erikos: Indeed, Isolved this by adding a default param to this._inputSourcesChanged(), so I only need to pass the flag from _syncPasswordMode.

js/ui/status/keyboard.js
```diff
@@ -317,6 +317,10 @@ const InputSourceManager = new Lang.Class({
         this._ibusSources = {};
 
         this._currentSource = null;
+        this._lastSource = null;
+        this._passwordModeEnabled = false;
+        this._fallbackInputSource = null;
+        this._needsFallbackSource = false;
```
cosimoc: Not required anymore

js/ui/status/keyboard.js
```diff
@@ -480,6 +484,52 @@ const InputSourceManager = new Lang.Class({
...
+            }
+            this._lastSource = this._currentSource;
+            this._activateInputSource(source, false);
+        } else {
+            if (!this._lastSource)
+                return;
+            source = this._lastSource;
+            this._activateInputSource(source, false);
+            this._lastSource = null;
+            this._fallbackInputSource = null;
```
cosimoc: I think it would be cleaner if this happened inside `this._inputSourcesChanged()` when `needsFallbackSource == false`

js/ui/status/keyboard.js
```diff
@@ -525,7 +575,7 @@ const InputSourceManager = new Lang.Class({
         this._mruSources = mruSources.concat(sourcesList);
     },
 
-    _inputSourcesChanged: function() {
+    _inputSourcesChanged: function(needsFallbackSource = false) {
```
cosimoc: In certain cases, `_inputSourcesChanged()` is called as a callback for settings changed signals. In that case, the first parameter here will be the settings object itself. I think to avoid churn, it would be cleanest if `inputSourcesChanged()` was left without additional parameters, and its body split into e.g. `_updateInputSources()` that has the extra parameter. `_syncPasswordMode()` would then just call `_updateInputSources()`.

js/ui/status/keyboard.js
erikos: Done.

js/ui/status/keyboard.js
erikos: Done.

js/ui/status/keyboard.js
erikos: You are absolutely right. Side-note: It would be cleaner if those callbacks would list the arguments and if unused do prepend with a '_' or similar to make linters happy. Not starting to think about named arguments e.g. Gtk.ComboBox(model=self.auth_store)...Sigh :)

Anyhow, adjusted accordingly.

js/ui/components/networkAgent.js
```diff
@@ -27,6 +29,8 @@ const NetworkSecretDialog = new Lang.Class({
     _init: function(agent, requestId, connection, settingName, hints, contentOverride) {
         this.parent({ styleClass: 'prompt-dialog' });
 
+        this._inputSourceIndicator = null;
```
cosimoc: No need to separately initialize this to `null`.

cosimoc:
The code in this PR looks good to me. There are two things left:
- some of these commits should be squashed
- as discussed in the ticket, we should not automatically switch to the Latin layout for now, at least until the FBE side is also implemented. (Even though if we start switching by default, potentially that would still be a problem for users that have set the password with previously.)

Keeping unmerged for now pending those two.

erikos:
Has been superseded, only bits in here are about the automatic switching to a latin layout for password entries. Closing.

PR #146, opened 2017-09-25
Improve handling being headless

mariospr:
This PR reverts a patch for this issue that was not actually fixing the problem since more changes were needed both in the shell (which were [changes that got lost during the previous rebase](https://github.com/endlessm/eos-shell/pull/2354/commits)) and in mutter itself.

Instead of doing that, it picks a patch from Jonas Adahl attached to [GNOME Bug 730551](https://bugzilla.gnome.org/show_bug.cgi?id=730551) that, along with the patches for mutter attached there, does actually fix the problem without more changes needed in the shell.

https://phabricator.endlessm.com/T19389

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -76,6 +76,7 @@ const LayoutManager = new Lang.Class({
         this._topActors = [];
         this._isPopupWindowVisible = false;
         this._startingUp = true;
+        this._pendingLoadBackground = false;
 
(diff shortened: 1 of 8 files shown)


mariospr:
I updated the PR based on Jonas' new patch and adding another patch on top that I also proposed upstream. I also tested everything and it works, but let's wait first to see what gets decided in bgo 730551.

PR #149, opened 2017-09-28
brightness: When dragging the slider, only update the value on drag-end events

mariospr:
Otherwise, it would make changes to the org.gnome.SettingsDaemon.Power.Screen
proxy faster than how they can actually be handled, resulting in a noticeable
delay both updating the brightness and moving the slider.

https://phabricator.endlessm.com/T19097

### DIFF
--- a/js/ui/status/brightness.js
+++ b/js/ui/status/brightness.js
@@ -41,7 +41,13 @@ const Indicator = new Lang.Class({
 
         this._slider = new Slider.Slider(0);
         this._slider.connect('value-changed', Lang.bind(this, this._sliderChanged));
+        this._slider.connect('drag-begin', Lang.bind(this, this._sliderDragBegan));
+        this._slider.connect('drag-end', Lang.bind(this, this._sliderDragEnded));
+
         this._slider.actor.accessible_name = _("Brightness");
+
+        this._sliderIsDragging = false;
+        this._sliderValue = this._proxy.Brightness / 100.0;
 
         let icon = new St.Icon({ icon_name: 'display-brightness-symbolic',
                                  style_class: 'popup-menu-icon' });
@@ -56,9 +62,28 @@ const Indicator = new Lang.Class({
 
     },
 
-    _sliderChanged: function(slider, value) {
-        let percent = value * 100;
+    _updateBrightness: function() {
+        let percent = this._sliderValue * 100;
         this._proxy.Brightness = percent;
+    },
+
+    _sliderChanged: function(slider, value) {
+        this._sliderValue = value;
+
+        // It's ok to change the brightness here only if the slider didn't
+        // change because of a mouse dragging event (e.g. keyboard), otherwise
+        // it could push changes faster than how they can actually be handled.
+        if (!this._sliderIsDragging)
+            this._updateBrightness();
+    },
+
+    _sliderDragBegan: function(slider) {
+        this._sliderIsDragging = true;
+    },
+
+    _sliderDragEnded: function(slider) {
+        this._updateBrightness();
+        this._sliderIsDragging = false;
     },
 
     _sync: function() {


js/ui/status/brightness.js
```diff
@@ -56,11 +62,34 @@ const Indicator = new Lang.Class({
...
+    },
+
+    _sliderDragBegan: function(slider) {
+        this._sliderIsDragging = true;
+    },
+
+    _sliderDragEnded: function(slider) {
+        // We only need to update the brightness here for mouse dragging events,
+        // since other events will update it on the 'value-changed' callback.
+        if (this._sliderIsDragging)
```
joaquimrocha: Is this really needed here? Shouldn't we assume that the slider must be dragging if you have a _drag-end_ event? Or does it receive a drag event when the value changes?
Also without this _if_ I don't think the comment is very valuable here, it's clear that this is for a drag event related change.

joaquimrocha:
Just a small comment in the code.
Although the PR looks fine for your approach, I think there's another technique for these use-cases that may increase the UX: you only update the slider on _drag-end_ **or** when the value changes after a small time-out.
This improves the UX when users drag a bit and stop without releasing the slider (as some kind of preview) because it will give them the feedback they expect: the brightness changes.
A timeout of ~150ms (if it gives enough time for the brightness to effectively change) should maybe be enough to give the "illusion" that the slider is changing the values, without creating a queue of events.

If you then increase the "last registered" value-change in line 94 for this feature (i.e. force a time-out), it will also avoid the time-out delay when the slider values are changed as a response to a HW event, and has the side effect that we won't need the _isDragging_ flag nor the _drag-begin_ callback.

js/ui/status/brightness.js
mariospr: You're perfectly right, this is a mistake, not sure why I added it back. I'll remove it

mariospr:
Thanks for the review. I'm ok with changing the `_sliderDragEnded()` function as you suggest, but I would not enter into the game of using timeouts here, because any timeout you set is going to be random and, depending on the actual speed of the machine you're running it in, could be either "too slow" or "too fast", so I'd rather go with this simpler approach.

I commented about the same issue upstream in for the reasons explained in https://bugzilla.gnome.org/show_bug.cgi?id=788288#c4. Let's see what they say there too...

joaquimrocha:
I hadn't seen the conversation about the use of a timeout upstream. Sorry for the noise.
You are right about the timeout maybe not being enough for slower machines but we can surely make an educated guess (if 100ms is enough for your fast machine, maybe 250ms is enough margin for a slower machine). It's a trade-off between improving the UX and risking users of very slow machines see this issue still. I'd prefer to improve the UX and eventually deal with it if there's a clear problem in slower machines. But it's of course up to you.

mariospr:
I see what you mean, but I honestly don't see such a big benefit on improving the UX in this particular case, I'd rather keep it simple and predictable instead. As I said, let's see what upstream thinks.

cosimoc:
I haven't *tried* the patch, but if it changed the behavior in the way that it looks from the code, I'm not a huge fan of it.
Basically I think that it would be confusing to only update the actual brightness on "mouse release"; it's nice to see the brightness change while the slider is moved.

If on the other hand a "drag end" is received also when the mouse sits still with the button pressed down, perhaps it would be a good compromise... What is the UX after this patch is applied?

It strikes me that another approach could be to change the g-s-d side so that changes are not propagated down the stack faster than they can be processed?

mariospr:
> I haven't tried the patch, but if it changed the behavior in the way that it looks from the code, I'm not a huge fan of it.
> Basically I think that it would be confusing to only update the actual brightness on "mouse release"; it's nice to see the brightness change while the slider is moved.

I know it's not perfect but, I thought this would improv  the UX already since what we have now is that when you drag the slider the slider too fast both the slider and the brightness do "funny things" in that the changes don't keep up with the movement of the mouse, and so I don't see the fact that the brightness changes in real time as a better UX, really :-).

That said, I agree ideally we should be fixing this in the best possible way, and exchanging a bad situation for a less-bad-but-still-not-good is probably not the best approach, but I didn't have time to investigate it further, this was a really quick patch that I though might be useful, that's why I proposed it.

Also, I by just re-testing it more carefully I noticed that a couple of use cases are still pretty broken:
  * Moving the slider with the mouse wheel (as no drag events are generated)
  * Moving the slider with the keyboard by keeping the arrow keys pressed down for a while

> If on the other hand a "drag end" is received also when the mouse sits still with the button pressed down, perhaps it would be a good compromise... What is the UX after this patch is applied?

I'm not sure I understand what you mean here, but I'm closing this request (and commenting in the bug upstream) based on my comment above about this not really working better with the mouse wheel and long key presses.

> It strikes me that another approach could be to change the g-s-d side so that changes are not propagated down the stack faster than they can be processed?

That would be a different way to do it, right. You still wouldn't accomplish a 1:1 situation between the slider's movement and the brightness levels changing, but at least it would change "as soon as possible". I'll mention that in the bug upstream

PR #157, opened 2017-10-16
Don't exclude web links that are not in the desktop from search

mariospr:
We used to hide this results from desktop search, but we want to include
them now so that those "native" applications are found when required, instead
of redirecting the user to the default search engine in the browser.

https://phabricator.endlessm.com/T19646


rshuler:
The simplistic approach is not desired, since it includes no filtering by locale.  We will need to further discuss the desired details of the feature.

PR #160, opened 2017-10-23
shellDBus: Don't send metric on AddAppIfNotVisible for IDs in the desktop

mariospr:
This pollutes the metrics server due to apps like Chrome, which already
have an icon in the desktop even before being actually instelled-

https://phabricator.endlessm.com/T19717

### DIFF
--- a/js/ui/shellDBus.js
+++ b/js/ui/shellDBus.js
@@ -534,16 +534,19 @@ const AppStoreService = new Lang.Class({
     },
 
     AddAppIfNotVisible: function(id) {
-        let eventRecorder = EosMetrics.EventRecorder.get_default();
-        let appId = new GLib.Variant('s', id);
-        eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
+        let visibleIcons = IconGridLayout.layout.getIcons(IconGridLayout.DESKTOP_GRID_ID);
+        let isIconVisible = visibleIcons.indexOf(id) !== -1;
+
+        if (!isIconVisible) {
+            let eventRecorder = EosMetrics.EventRecorder.get_default();
+            let appId = new GLib.Variant('s', id);
+            eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
+        }
 
         if (IconGridLayout.layout.iconIsFolder(id))
             return;
 
-        let visibleIcons = IconGridLayout.layout.getIcons(IconGridLayout.DESKTOP_GRID_ID);
-        if (visibleIcons.indexOf(id) == -1)
-            IconGridLayout.layout.appendIcon(id, IconGridLayout.DESKTOP_GRID_ID);
+        IconGridLayout.layout.appendIcon(id, IconGridLayout.DESKTOP_GRID_ID);
     },
 
     ReplaceApplication: function(originalId, replacementId) {
--- a/js/ui/shellDBus.js
+++ b/js/ui/shellDBus.js
@@ -514,6 +514,17 @@ const AppStoreIface = '<node> \
 </interface> \
 </node>';
 
+function _iconIsVisibleOnDesktop(id) {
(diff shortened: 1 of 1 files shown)


js/ui/shellDBus.js
```diff
@@ -534,15 +534,16 @@ const AppStoreService = new Lang.Class({
     },
 
     AddAppIfNotVisible: function(id) {
-        let eventRecorder = EosMetrics.EventRecorder.get_default();
-        let appId = new GLib.Variant('s', id);
-        eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
+        let visibleIcons = IconGridLayout.layout.getIcons(IconGridLayout.DESKTOP_GRID_ID);
+        let isIconVisible = visibleIcons.indexOf(id) != -1;
```
smspillaz: !==

js/ui/shellDBus.js
```diff
@@ -534,15 +534,16 @@ const AppStoreService = new Lang.Class({
...
-            return;
+        if (!isIconVisible) {
+            let eventRecorder = EosMetrics.EventRecorder.get_default();
+            let appId = new GLib.Variant('s', id);
+            eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
+        }
 
-        let visibleIcons = IconGridLayout.layout.getIcons(IconGridLayout.DESKTOP_GRID_ID);
-        if (visibleIcons.indexOf(id) == -1)
+        if (!IconGridLayout.layout.iconIsFolder(id))
```
smspillaz: It probably makes sense to keep this as an early return

js/ui/shellDBus.js
mariospr: Ok

js/ui/shellDBus.js
mariospr: I don't mind either way. Since there's only one thing to be done in the `!iconIsFolder(id)` case, I kind of liked the more compact style, but I will put the early return back in place.

js/ui/shellDBus.js
```diff
@@ -562,6 +570,10 @@ const AppStoreService = new Lang.Class({
         // since the replace operation degenerates to
         // append if the source icon was not available
         IconGridLayout.layout.replaceIcon(originalId, replacementId, IconGridLayout.DESKTOP_GRID_ID);
+
+        if (!iconWasVisible) {
+            _reportAppAddedMetric(replacementId);
```
mariospr: Nit: no need for braces in single-line if statements

js/ui/shellDBus.js
```diff
@@ -534,25 +545,22 @@ const AppStoreService = new Lang.Class({
...
-            eventRecorder.record_event(SHELL_APP_ADDED_EVENT, appId);
-        }
+        let iconWasVisible = _iconIsVisibleOnDesktop(id);
 
         if (IconGridLayout.layout.iconIsFolder(id))
             return;
 
         IconGridLayout.layout.appendIcon(id, IconGridLayout.DESKTOP_GRID_ID);
+
+        if (!iconWasVisible) {
```
mariospr: Same comment about the braces here. Also, is there any reason you're moving this after the early return `iconIsFolder()`?

mariospr: FWIW, it looks good to me, I think it does not make any difference in practical terms

smspillaz: Ah - yeah I spoke to Cosimo about this yesterday and reporting the metric *before* doing the action is a broken behaviour - we don't want to report a metric if it would have been blocked by the application logic.

js/ui/shellDBus.js
smspillaz: Fixed in a followup force push

mariospr:
Thanks @smspillaz. Merging now everything to master.

PR #161, opened 2017-10-23
Fix window picker animations

mariospr:
https://phabricator.endlessm.com/T18032


mariospr:
This is not the complete solution, closing for now while I work on a better PR, to prevent accidental merges

PR #164, opened 2017-10-26
Fix strange animation to the left when unminimizing windows

mariospr:
https://phabricator.endlessm.com/T19659

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1581,6 +1581,10 @@ const WindowManager = new Lang.Class({
             let monitor = Main.layoutManager.monitors[actor.meta_window.get_monitor()];
             let xDest = monitor.x + actor.x;
             let yDest = monitor.y + monitor.height;
+            if (!monitor) {
+                this._minimizeWindowDone(shellwm, actor);
+                return;
+            }
 
             Tweener.addTween(actor,
                              { x: xDest,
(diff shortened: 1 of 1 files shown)


js/ui/windowManager.js
```diff
@@ -1579,8 +1579,10 @@ const WindowManager = new Lang.Class({
                          });
         } else {
             let monitor = Main.layoutManager.monitors[actor.meta_window.get_monitor()];
-            let xDest = monitor.x + actor.x;
```
rshuler: The removal of xDest and yDest should be moved to the next commit.

mariospr: Oops! Late-evening rebasing after long days is not a good idea :). Fixed

js/ui/windowManager.js
```diff
@@ -1641,14 +1642,19 @@ const WindowManager = new Lang.Class({
                          });
         } else {
             let monitor = Main.layoutManager.monitors[actor.meta_window.get_monitor()];
+            if (!monitor) {
+                this._unminimizeWindowDone(shellwm, actor);
+                return;
+            }
```
smspillaz: This strikes me as a little strange - `monitor` would only be `null` in cases where the window didn't have a valid monitor. Is this a common case?

mariospr: I believe it can also be null if you don't have any monitor connected at the time.

PR #168, opened 2017-10-31
overview, viewSelector: Handle first time the overview is shown

mariospr:
The fact that we want to start up directly in the Overview mode
on Endless combined with the possibility of having lots of icons
on the desktop makes it necessary to handle the first time the
overview is shown in a particular way in order to make it less
prone to flickering and confusion (e.g. already running apps
disappearing when restarting the shell).

However, we can't check the value of the startingUp property
from the layout because by the time Overview::show() gets called
the layout manager is no longer "starting up". Thus, add an
additional property to the overview so that we can do an extra
check to see if it's the first time it's shown, when needed.

This should be squashed with 4d34e93d in future rebases.

https://phabricator.endlessm.com/T19778

### DIFF
--- a/js/ui/overview.js
+++ b/js/ui/overview.js
@@ -226,6 +226,7 @@ const Overview = new Lang.Class({
         this._toggleToHidden = false;   // Whether to hide the overview when either toggle function is called
         this._targetPage = null;        // do we have a target page to animate to?
         this._modal = false;            // have a modal grab
+        this._shownOnce = false;         // useful for handling events relevant only the first time it's shown
         this.animationInProgress = false;
         this.visibleTarget = false;
         this.opacityPrepared = false;
@@ -729,7 +730,7 @@ const Overview = new Lang.Class({
         this._coverPane.show();
         this.emit('showing');
 
-        if (Main.layoutManager.startingUp || this.opacityPrepared) {
+        if (!this._shownOnce || this.opacityPrepared) {
             this._overview.opacity = EOS_ACTIVE_GRID_OPACITY;
             this.opacityPrepared = false;
             this._showDone();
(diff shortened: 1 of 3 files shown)


js/ui/overview.js
```diff
@@ -756,6 +757,9 @@ const Overview = new Lang.Class({
         this._coverPane.hide();
 
         this.emit('shown');
+        if (!this._shownOnce)
```
rshuler: It's just a nit, but I would probably not bother with the `if` clause and just always set the variable true.

cosimoc:
This looks generally OK to me, but is there a way to remove the duplication between the two properties? It would be great if we could use one instead.
Otherwise, feel free to merge with @rshuler's suggestion.

js/ui/overview.js
mariospr: I thought the same, but in the end I decided to do it this way to make it more explicit that this variable is only expected to change its value once, since otherwise it might look like is one of those flags that should be set to one value in the `show()` function and then set to the opposite value in `showDone()`.

That said, a comment would suffice too, will change it

mariospr:
> This looks generally OK to me, but is there a way to remove the duplication between the two properties? It would be great if we could use one instead.

That's exactly what I tried with my previous approach of setting `startingUp = false` after the `startup-complete` signal got emitted, but that broke the initial setup, so I'd better leave it as it is.

The problem with this two variables is that the lifetimes of their initial value (`startingUp == true` and `shownOnce == false`)  do not match, and we normally (at least at the moment) need to check whether `shownOnce` is still `false` after `startingUp` has changed to `true`, because it's precisely from the callback for `startup-completed` where the first call to `Overview::show()` is done.

PR #170, opened 2017-11-02
Handle the overview on startup-prepared, not on startup-complete

mariospr:
https://phabricator.endlessm.com/T19778

### DIFF
--- a/js/ui/overview.js
+++ b/js/ui/overview.js
@@ -226,7 +226,6 @@ const Overview = new Lang.Class({
         this._toggleToHidden = false;   // Whether to hide the overview when either toggle function is called
         this._targetPage = null;        // do we have a target page to animate to?
         this._modal = false;            // have a modal grab
-        this._shownOnce = false;         // useful for handling events relevant only the first time it's shown
         this.animationInProgress = false;
         this.visibleTarget = false;
         this.opacityPrepared = false;
@@ -730,7 +729,7 @@ const Overview = new Lang.Class({
         this._coverPane.show();
         this.emit('showing');
 
-        if (!this._shownOnce || this.opacityPrepared) {
+        if (Main.layoutManager.startingUp || this.opacityPrepared) {
             this._overview.opacity = EOS_ACTIVE_GRID_OPACITY;
             this.opacityPrepared = false;
             this._showDone();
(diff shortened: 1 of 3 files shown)


js/ui/overview.js
```diff
@@ -525,8 +525,22 @@ const Overview = new Lang.Class({
...
+                                         'gnome-initial-setup-done']);
+        return !GLib.file_test(path, GLib.FileTest.EXISTS);
+    },
+
+    _onStartupPrepared: function() {
+        if (this.isDummy)
+            return;
+
+        // do not show overview when FBE is running
+        if (this._isInitialSetupRunning()) {
```
cosimoc: Remove curly braces

js/ui/overview.js
```diff
@@ -525,8 +525,22 @@ const Overview = new Lang.Class({
         this.viewSelector.focusSearch();
     },
 
-    _onStartupCompleted: function() {
-        this.showApps();
+    _isInitialSetupRunning: function() {
```
cosimoc: I wonder if we couldn't instead more directly check the session mode to determine this information.

js/ui/overview.js
mariospr: Argh! Again...

js/ui/overview.js
mariospr: Hmm... that's a good point, but your suggestion would work fine for the "true FBE", the one that happens only once on the very first boot, but would not work for FBEs triggered for new users, since gnome shell wouldn't be running in initial-setup mode for them.

That said, we are actively disabling FBE for existing installations (see T19802), so perhaps that's a fair assumption to make and we can rely on the session mode. Or, even better, perhaps we can drop this initial-setup related bits altogether since, now I'm thinking twice, I'm having doubts of the impact of changing the viewSelector to point to the APPs page if the shell is running in a different mode anyway... I'll test this theory and push a new patch

mariospr: Indeed, no need to do this at all, as it's only useful for `--existing-user` runs of FBE, which we disable. Sorry for the confusion, and thanks for pointing it out!

cosimoc: Good point about gnome-initial-setup running in existing user mode, even though we don't use it... Feel free to ignore my suggestion if this seems more practical.

PR #177, opened 2017-11-09
RenmoteSearch: exclude content from evergreen apps

erikos:
The apps word of the day and quote of the day do
provide a search provider. We want those to be
excluded from the search.

https://phabricator.endlessm.com/T19013

### DIFF
--- a/js/ui/remoteSearch.js
+++ b/js/ui/remoteSearch.js
@@ -96,6 +96,9 @@ function loadRemoteSearchProviders(searchSettings, callback) {
             try {
                 let desktopId = keyfile.get_string(group, 'DesktopId');
                 appInfo = Gio.DesktopAppInfo.new(desktopId);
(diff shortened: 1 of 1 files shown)


js/ui/remoteSearch.js
```diff
@@ -73,6 +73,10 @@ function loadRemoteSearchProviders(searchSettings, callback) {
     function loadRemoteSearchProvider(file) {
         let keyfile = new GLib.KeyFile();
         let path = file.get_path();
+        let evergreenApps = [
+            'com.endlessm.quote_of_the_day.en.desktop',
+            'com.endlessm.word_of_the_day.en.desktop'
+        ];
```
cosimoc: Instead of hardcoding this list, you can simply use `appInfo.should_show()` and exclude the app if that's false. (It will look at the `NoDisplay` key in the desktop file behind the scenes.)

PR #179, opened 2017-11-16
passwordMode: add keyboard layout switcher for password entries

erikos:
This adds the infrastructure to verify that the current
keyboard layout in use is a latin one when the user is prompted
to enter a password. The code to determine if a keyboard
layout is a latin is using the function "find_latin_keysym"
borrowed from Mutter.

In the near future we will switch automatically to an available
latin layout. If no suitable layout is available we
fallback to the default layout 'en'. The change is reflected in
the layout switcher in the taskbar.

When the password entry is in a modal dialog and the keyboard
switcher in the taskbar can not be accessed, we add another
switcher in the dialog, next to the password entry. The changes
made using this switcher are propagated to the keyboard
switcher in the taskbar and the popover menu revealed by the
super+space keybinding.

The cases handled are the authPrompt used in the lock screen and
the login view, the polkitAgent used to unlock the user page
in the control center for editing, the network security dialog
dealing with passwords for a Wireless or VPN connection and
the dialog prompting for a password when mounting encrypted
partitions.

https://phabricator.endlessm.com/T19956

### DIFF
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -15,7 +15,7 @@ const Animation = imports.ui.animation;
 const Batch = imports.gdm.batch;
 const Config = imports.misc.config;
 const GdmUtil = imports.gdm.util;
-const InputSourceManager = imports.ui.status.keyboard;
+const Keyboard = imports.ui.status.keyboard;
 const Params = imports.misc.params;
 const ShellEntry = imports.ui.shellEntry;
 const Tweener = imports.ui.tweener;
@@ -116,8 +116,8 @@ const AuthPrompt = new Lang.Class({
                          y_fill: false,
                          x_align: St.Align.START });
 
-        // for setting passwords we switch to a latin keyboard layout
-        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        // for setting passwords we enable a password mode for kbd layouts
+        this._inputSourceManager = Keyboard.getInputSourceManager();
         this._inputSourceManager.passwordModeEnabled = true;
 
         this._entry.grab_key_focus();
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -15,6 +15,7 @@ const Animation = imports.ui.animation;
 const Batch = imports.gdm.batch;
 const Config = imports.misc.config;
 const GdmUtil = imports.gdm.util;
+const InputSourceManager = imports.ui.status.keyboard;
 const Params = imports.misc.params;
 const ShellEntry = imports.ui.shellEntry;
 const Tweener = imports.ui.tweener;
@@ -115,6 +116,10 @@ const AuthPrompt = new Lang.Class({
                          y_fill: false,
                          x_align: St.Align.START });
 
+        // for setting passwords we switch to a latin keyboard layout
+        this._inputSourceManager = InputSourceManager.getInputSourceManager();
+        this._inputSourceManager.passwordModeEnabled = true;
+
         this._entry.grab_key_focus();
 
         this._message = new St.Label({ opacity: 0,
@@ -183,6 +188,7 @@ const AuthPrompt = new Lang.Class({
     _onDestroy: function() {
         this._userVerifier.destroy();
         this._userVerifier = null;
+        this._inputSourceManager.passwordModeEnabled = false;
     },
 
     _initButtons: function() {
--- a/js/ui/components/networkAgent.js
+++ b/js/ui/components/networkAgent.js
@@ -12,6 +12,8 @@ const Shell = imports.gi.Shell;
 const St = imports.gi.St;
 
 const Config = imports.misc.config;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
+const InputSourceManager = imports.ui.status.keyboard;
 const Main = imports.ui.main;
 const MessageTray = imports.ui.messageTray;
 const ModalDialog = imports.ui.modalDialog;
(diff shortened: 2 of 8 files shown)


js/ui/status/keyboard.js
```diff
@@ -480,6 +483,49 @@ const InputSourceManager = new Lang.Class({
...
+        this._syncPasswordMode();
+    },
+
+    _syncPasswordMode: function() {
+        let source;
+
+        if (this._passwordModeEnabled) {
+            // do not switch the keyboard layout if the current one is a latin one
+            if (this._keyboardManager.isLatinLayout(this._currentSource.id))
+                return;
```
cosimoc: I think this check is not required now, since we're only interested in ensuring that at least one latin source exists. The loop below should take care of that, since AFAIU `this._mruSources` include `this._currentSource`.

js/ui/status/keyboard.js
```diff
@@ -480,6 +483,49 @@ const InputSourceManager = new Lang.Class({
...
+                    source = src;
+                    break;
+                }
+            }
+            if (!source) {
+                // fallback to an 'en' keyboard layout
+                this._updateInputSources(true);
+                source = this._fallbackInputSource;
+            }
+            this._lastSource = this._currentSource;
```
cosimoc: Is `this._lastSource` still needed right now? We don't seem to do anything with it anymore. Related question: does the "transient" latin source go away from the system configuration when the dialog is dismissed? (It should.)

js/ui/status/keyboard.js
```diff
@@ -480,6 +483,49 @@ const InputSourceManager = new Lang.Class({
...
+                if (this._keyboardManager.isLatinLayout(src.id)) {
+                    source = src;
+                    break;
+                }
+            }
+            if (!source) {
+                // fallback to an 'en' keyboard layout
+                this._updateInputSources(true);
+                source = this._fallbackInputSource;
+            }
```
cosimoc: Since we're simply ensuring that a latin source exists, the loop and this check could be simplified by something like
```
if (!this._mruSources.some(Lang.bind(this, function(src) {
  this._keyboardManager.isLatinLayout(src.id);
}))) {
  this._updateInputSources(true);
}
```

js/gdm/authPrompt.js
```diff
@@ -15,6 +15,7 @@ const Animation = imports.ui.animation;
 const Batch = imports.gdm.batch;
 const Config = imports.misc.config;
 const GdmUtil = imports.gdm.util;
+const InputSourceManager = imports.ui.status.keyboard;
```
mariospr: Is there any reason why this import is `InputSourceManager` instead of `Keyboad` (in reference to the name of the module imported)?

js/ui/components/networkAgent.js
```diff
@@ -12,6 +12,8 @@ const Shell = imports.gi.Shell;
 const St = imports.gi.St;
 
 const Config = imports.misc.config;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
+const InputSourceManager = imports.ui.status.keyboard;
```
mariospr: Same comment

js/ui/components/networkAgent.js
```diff
@@ -123,11 +131,18 @@ const NetworkSecretDialog = new Lang.Class({
                 secret.valid = true;
 
             if (rtl) {
-                layout.attach(secret.entry, 0, pos, 1, 1);
-                layout.attach(label, 1, pos, 1, 1);
+                let j = 0;
+                if (i === 0) {
+                    layout.attach(this._inputSourceIndicator.container, j, pos, 1, 1);
+                    j++;
```
mariospr: Is this one really needed? Looks like you're already accounting for this space below when using `j+1` and `j+2`.

Also, I think this chunk of code requires a comment briefly explaining what's going on here with `i`, `j` and this calls to `attach()`, since it's hard to figure it out just by reading the code.

js/ui/components/networkAgent.js
```diff
@@ -123,11 +131,18 @@ const NetworkSecretDialog = new Lang.Class({
...
+                }
+                layout.attach(secret.entry, j+1, pos, 1, 1);
+                layout.attach(label, j+2, pos, 1, 1);
             } else {
                 layout.attach(label, 0, pos, 1, 1);
                 layout.attach(secret.entry, 1, pos, 1, 1);
+                if (i === 0)
+                    layout.attach(this._inputSourceIndicator.container, 2, pos, 1, 1);
```
mariospr: A comment here would be helpful too

js/ui/components/polkitAgent.js
```diff
@@ -15,7 +15,10 @@ const PolkitAgent = imports.gi.PolkitAgent;
 
 const Animation = imports.ui.animation;
 const Components = imports.ui.components;
+const InputSourceManager = imports.ui.status.keyboard;
```
mariospr: My usually nit here

js/ui/shellMountOperation.js
```diff
@@ -11,10 +11,13 @@ const St = imports.gi.St;
 const Shell = imports.gi.Shell;
 
 const CheckBox = imports.ui.checkBox;
+const InputSourceIndicator = imports.ui.status.keyboard.InputSourceIndicator;
+const InputSourceManager = imports.ui.status.keyboard;
```
mariospr: Ditto

js/ui/status/keyboard.js
```diff
@@ -890,8 +893,10 @@ const InputSourceIndicator = new Lang.Class({
         let newSource = this._inputSourceManager.currentSource;
 
         if (oldSource) {
-            this._menuItems[oldSource.index].setOrnament(PopupMenu.Ornament.NONE);
-            this._indicatorLabels[oldSource.index].hide();
+            if (Object.keys(this._menuItems).indexOf(String(oldSource.index)) !== -1)
```
mariospr: Could you use `oldSource.index in this._menuItems` instead?

js/ui/status/keyboard.js
```diff
@@ -890,8 +893,10 @@ const InputSourceIndicator = new Lang.Class({
...
+                this._menuItems[oldSource.index].setOrnament(PopupMenu.Ornament.NONE);
+            if (Object.keys(this._indicatorLabels).indexOf(String(oldSource.index)) !== -1)
```
mariospr: Similarly, could you use `oldSource.index in this._indicatorLabels` instead?

PR #182, opened 2017-11-27
Avoid executing RegExps to find URLs in the search bar whenever possible

mariospr:
At the moment we were matching the text inserted in the search bar against
non trivial regular expression ANY time that the search terms changed, which
is pretty aggressive on the CPU, easily getting the shell into an apparently
"locked-up" state, especially for longer searchs strings.

Fortunately, we can do some optimizations here to prevent this situations:

  1. If the term is a valid URI in that it's possible to parse at least
     its scheme and host fields, it's considered a valid URL "as-is".
  2. Else, if the term is a generic string exceeding the maximum length
     specified then we simply ignore it and move onto the next term.
  3. In any other case (non-URI term, valid length) we match the term
     passed against the regular expression to determine if it's a URL.

Turns out it's a bit hard to find out up-to-date stats on what the average
length of a domain name is, but I found a couple of articles from a while
ago [1][2] that claim that "many of the world's most visited web sites have
domain names of between 6 - 10 characters", so I think setting a limit of
only 32 characters for the full term (that would account for the domain
name, the extension and a possible subdomain) is reasonable while helping
a lot not to lock the shell up, therefogre feeling much more responsive.

[1] https://www.domainregistration.com.au/news/2013/1301-domain-length.php
[2] https://www.domainregistration.com.au/infocentre/info-domain-length.php

https://phabricator.endlessm.com/T20051

### DIFF
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -69,18 +69,57 @@ const _searchUrlRegexp = new RegExp(
     '^([a-zA-Z0-9]+(\.[a-zA-Z0-9]+)+.*)\\.+[A-Za-z0-9\.\/%&=\?\-_]+$',
     'gi');
 
+const supportedSearchSchemes = [ 'http', 'https', 'ftp' ];
+
 // findSearchUrls:
 // @terms: list of searchbar terms to find URLs in
+// @maxLength: maximum number of characters in each non-URI term to match against, defaults
+//             to 32 characters to prevent hogging the CPU with too long generic strings.
 //
-// Similar to "findUrls", but for use only with terms from the searchbar.
-// The regex for these URLs matches strings such as "google.com" (note the
-// lack of preceding scheme).
+// Similar to "findUrls", but adapted for use only with terms from the searchbar.
+//
+// In order not to be too CPU-expensive, this function is implemented in the following way:
+//   1. If the term is a valid URI in that it's possible to parse at least
+//      its scheme and host fields, it's considered a valid URL "as-is".
+//   2. Else, if the term is a generic string exceeding the maximum length
+//      specified then we simply ignore it and move onto the next term.
+//   3. In any other case (non-URI term, valid length) we match the term
+//      passed against the regular expression to determine if it's a URL.
+//
+// Note that the regex for these URLs matches strings such as "google.com" (no need to the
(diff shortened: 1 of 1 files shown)


js/misc/util.js
```diff
@@ -71,16 +72,45 @@ const _searchUrlRegexp = new RegExp(
...
 // Return value: the list of URLs found in the string
-function findSearchUrls(terms) {
+function findSearchUrls(terms, maxLength=32) {
     let res = [], match;
-    for (let i in terms) {
-        while ((match = _searchUrlRegexp.exec(terms[i])))
+    for (let term of terms) {
+        // We avoid running any RegExp or applying any character limit if
+        // we already now we've got a valid URL to open in the browser.
+        let uri = new Soup.URI(term)
```
cosimoc: Instead of dragging in libsoup for this, you could use `g_uri_parse_scheme()`, no?

js/misc/util.js
```diff
@@ -71,16 +72,45 @@ const _searchUrlRegexp = new RegExp(
...
+        if (uri && uri.get_host()) {
+            res.push(uri.to_string(false));
+            continue
+        }
+
+        // Try to save CPU cycles from regexp-matching too long strings.
+        if ((term.length > maxLength))
```
cosimoc: The hardcoding of the length here is a bit unfortunate -- I was wondering if we couldn't possibly find a better regexp. But the issue is real (experienced it myself), so let's do it :)

js/misc/util.js
mariospr: Libsoup is already used in other parts of the shell, so it's not a new dependency, and I kind of like having SoupURI doing all the usual validations for me instead of me having to manually check for valid schemes, the `://` and the host part. That said, it might be that using all that logic here is a bit overkill and we could just do those checks ourselves in this particular case.

js/misc/util.js
mariospr: I swear I tried to have a solution that didn't involve this kind of constraints, but after quite some experiments and tests I came to conclude that this was actually not too bad, and honestly... if the host part of an URL you're typing in there is more than 32 characters, the "real problem lies somewhere else" IMHO :-)

Also, you can always paste the full URL (with the scheme part) and in that case this check is ignored.

PR #183, opened 2017-11-29
Fix brusque animations showing/hiding the Overview

mariospr:
https://phabricator.endlessm.com/T19824

### DIFF
--- a/js/ui/viewSelector.js
+++ b/js/ui/viewSelector.js
@@ -723,10 +723,10 @@ const ViewSelector = new Lang.Class({
         let layoutViewsClone = new ViewsClone(this, this._viewsDisplay, false);
         Main.layoutManager.setViewsClone(layoutViewsClone);
 
-        let overviewViewsClone = new ViewsClone(this, this._viewsDisplay, true);
-        Main.overview.setViewsClone(overviewViewsClone);
+        this._overviewViewsClone = new ViewsClone(this, this._viewsDisplay, true);
+        Main.overview.setViewsClone(this._overviewViewsClone);
         this._appsPage.bind_property('visible',
-                                     overviewViewsClone, 'visible',
+                                     this._overviewViewsClone, 'visible',
                                      GObject.BindingFlags.SYNC_CREATE |
                                      GObject.BindingFlags.INVERT_BOOLEAN);
     },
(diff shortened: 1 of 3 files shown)


nedrichards:
:gb: language unlocked!

js/ui/viewSelector.js
```diff
@@ -876,20 +828,57 @@ const ViewSelector = new Lang.Class({
...
-        if (oldPage)
-            this._animateOut(oldPage, doFadeAnimation)
-        else
-            this._animateIn(null, doFadeAnimation);
+        if (oldPage && doFadeAnimation) {
+            // When fading to the apps page, tween the opacity of the
+            // clone instead, and set the apps page to full solid immediately
+            if (page == this._appsPage) {
+                page.opacity = 255;
+                this._overviewViewsClone.opacity = AppDisplay.EOS_INACTIVE_GRID_OPACITY;
```
cosimoc: Do you need to set the saturation here as well?

js/ui/viewSelector.js
```diff
@@ -876,20 +828,57 @@ const ViewSelector = new Lang.Class({
...
+                                   saturation: AppDisplay.EOS_INACTIVE_GRID_SATURATION,
+                                   time: OverviewControls.SIDE_CONTROLS_ANIMATION_TIME,
+                                   transition: 'easeOutQuad' });
+                this._fadePageIn(oldPage, doFadeAnimation);
+            } else {
+                Tweener.addTween(oldPage,
+                                 { opacity: 0,
+                                   time: OverviewControls.SIDE_CONTROLS_ANIMATION_TIME,
+                                   transition: 'easeOutQuad',
+                                   onComplete: Lang.bind(this,
```
cosimoc: Use `onCompleteScope` here instead of `Lang.bind()` for consistency with the surrounding code.

js/ui/viewSelector.js
mariospr: Oops! Yes, probably. I took the code from the old shell, but I agree we should set the initial value

js/ui/viewSelector.js
mariospr: Again, took from the original code, but will change it. onCompleteScope seems clearer to me than Lang.bind()

PR #165, opened 2017-10-27
T18012 v2

GeorgesStavracas:
This is a rebase of @mariospr's `T18012` branch, with a couple of additional patches fixing the remaining issues of the original branch.

https://phabricator.endlessm.com/T18012

PR #184, opened 2017-11-30
T20087

GeorgesStavracas:
This brings the animation code back in the Shell. Mostly based on @mariospr's DnD work, but with a few fixups.

https://phabricator.endlessm.com/T20087

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -206,8 +206,7 @@ const BaseAppView = new Lang.Class({
 
     _trimInvisible: function(items) {
         let appSystem = Shell.AppSystem.get_default();
-        return items.filter(Lang.bind(this,
-            function(itemId) {
+        return items.filter(Lang.bind(this, function(itemId) {
                 return IconGridLayout.layout.iconIsFolder(itemId) || appSystem.lookup_app(itemId) || (itemId == EOS_APP_CENTER_ID);
             }));
     },
@@ -242,8 +241,7 @@ const BaseAppView = new Lang.Class({
         newItemLayout = this._trimInvisible(newItemLayout);
 
         let addedIds = [];
-        for (let newItemIdx in newItemLayout) {
-            let newItem = newItemLayout[newItemIdx];
+        for (let newItem of newItemLayout) {
             let oldItemIdx = oldItemLayout.indexOf(newItem);
 
             if (oldItemIdx < 0)
@@ -271,8 +269,8 @@ const BaseAppView = new Lang.Class({
 
         // Create a map from app ids to icon objects
         let iconTable = {};
-        for (let idx in this._allItems)
-            iconTable[this._allItems[idx].getId()] = this._allItems[idx];
+        for (let item of this._allItems)
+            iconTable[item.getId()] = item;
 
         let layoutIds = this.getLayoutIds();
 
@@ -876,22 +874,22 @@ const AllView = new Lang.Class({
 
         let itemId = item.get_id();
 
-        if (IconGridLayout.layout.iconIsFolder(itemId)) {
-            let icon = new FolderIcon(item, this);
-            icon.connect('name-changed', Lang.bind(this, this._itemNameChanged));
-            this.folderIcons.push(icon);
-            if (this._addedFolderId == itemId) {
-                this.selectAppWithLabelMode(this._addedFolderId, EditableLabelMode.EDIT);
-                this._addedFolderId = null;
-            }
-
-            return icon;
+        if (!IconGridLayout.layout.iconIsFolder(itemId)) {
+            return new AppIcon(item,
+                               { isDraggable: true,
+                                 parentView: this },
+                               null);
         }
 
-        return new AppIcon(item,
-                           { isDraggable: true,
-                             parentView: this },
-                           null);
+        let icon = new FolderIcon(item, this);
+        icon.connect('name-changed', Lang.bind(this, this._itemNameChanged));
+        this.folderIcons.push(icon);
+        if (this._addedFolderId == itemId) {
+            this.selectAppWithLabelMode(this._addedFolderId, EditableLabelMode.EDIT);
+            this._addedFolderId = null;
+        }
+
+        return icon;
     },
 
     _maybeAddAppCenterIcon: function() {
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1772,9 +1772,10 @@ const FolderView = new Lang.Class({
             if (!this.iconsNeedRedraw())
                 return;
 
-            this._redisplay();
-
-            this._folderIcon.icon.reloadIcon();
+            if (!this._recreateIconWorkId)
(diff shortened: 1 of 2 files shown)


js/ui/appDisplay.js
```diff
@@ -772,7 +803,14 @@ const AllView = new Lang.Class({
 
     _redisplay: function() {
         if (this.getAllItems().length == 0) {
-            this.addIcons();
+            if (Main.layoutManager.startingUp) {
+                this.addIcons(true);
+
+                Main.layoutManager.connect('startup-complete',
+                                           Lang.bind(this, this._animateIconsIn));
```
mariospr: As mentioned in https://phabricator.endlessm.com/T18012#513379, this is causing problems on startup because startup-complete is too late to do the animation, since Overview's `showApps()` method is already called on startup-prepare.

Maybe call `_animateIconsIn()` rightaway from here?

js/ui/iconGrid.js
```diff
@@ -833,9 +833,9 @@ const IconGrid = new Lang.Class({
             dndDropPosition[1] -= Math.floor(oldIcon.get_size()[1] / 2);
 
             // Draw it at the location where DnD accept occured
-            let releaseOffset = this._findRelativeOffset(oldIcon, dndDropPosition);
-            oldIcon.translation_x = releaseOffset[0];
-            oldIcon.translation_y = releaseOffset[1];
+            //let releaseOffset = this._findRelativeOffset(oldIcon, dndDropPosition);
+            //oldIcon.translation_x = releaseOffset[0];
+            //oldIcon.translation_y = releaseOffset[1];
```
mariospr: Please remove the code, instead of commenting out.

Also, if this is truly temporary, I think it would be nice to comment on the commit message **why** it's temporary and/or what needs to be done to get a "more permanent" fix.

js/ui/iconGrid.js
```diff
@@ -1328,6 +1328,18 @@ const PaginatedIconGrid = new Lang.Class({
...
         }
+    },
+
+    animateShuffling: function(changedItems, removedItems, originalItemData, callback) {
+        // If we try to animate the shuffling before the onComplete callbacks in
+        // closeExtraSpace are called, they are cancelled and we never get to clean
+        // this._extraSpaceData. So, force the cleanup here.
+        if (this._extraSpaceData) {
+            this._extraSpaceData = null;
+            this.emit('space-closed');
```
mariospr: What about adding an `onError` callback to the tween and doing this from there? And if that works, maybe refactoring this code into a helper function, since it's pretty much identical to https://github.com/endlessm/gnome-shell/pull/184/commits/9b6c7272b0157d6fbdcb5984781fe542ff68769a#diff-477f992ae3cb7507f1459b10d86d0e73R1326

mariospr:
Please squash the changes `Trivial: Update style` into their respective commits (also updating those that belong to https://github.com/endlessm/gnome-shell/pull/165

js/ui/appDisplay.js
```diff
@@ -206,8 +206,7 @@ const BaseAppView = new Lang.Class({
 
     _trimInvisible: function(items) {
         let appSystem = Shell.AppSystem.get_default();
-        return items.filter(Lang.bind(this,
-            function(itemId) {
+        return items.filter(Lang.bind(this, function(itemId) {
```
mariospr: This, and the rest of changes in this commit, should be squashed into their respective commits from both this PR and https://github.com/endlessm/gnome-shell/pull/165, instead of having a separate commit for these changes.

js/ui/appDisplay.js
```diff
@@ -1772,16 +1772,18 @@ const FolderView = new Lang.Class({
             if (!this.iconsNeedRedraw())
                 return;
 
-            this._redisplay();
-
-            this._folderIcon.icon.reloadIcon();
+            if (!this._recreateIconWorkId)
+                this._recreateIconWorkId = Main.initializeDeferredWork(this._folderIcon.actor, Lang.bind(this, this._redisplay));
+            else
+                Main.queueDeferredWork(this._recreateIconWorkId);
```
mariospr: Is this really needed now that 9d63828 has been reverted? https://github.com/endlessm/gnome-shell/commit/d5595f5fa1aa1a869426dcb4e3fff3c4869bcafb

js/ui/iconGrid.js
```diff
@@ -1185,8 +1187,16 @@ const PaginatedIconGrid = new Lang.Class({
...
+        let size_changed = availWidth != this._adaptedWidth || availHeight != this._adaptedHeight;
+
         this.parent(availWidth, availHeight);
         this._computePages(availWidth, availHeight);
+
+        // If the adapted size changed, we must queue a relayout in the sizing
+        // machinery to make sure we always get the correct size. This cannot
+        // be done inside adaptToSize(), so we can't just call allocate() here.
+        if (size_changed)
+            this._grid.queue_relayout();
```
mariospr: IIRC, this was a speculative commit to try to fix the issues during startup, but my testing today found out that this is not really needed, please drop it.

js/ui/iconGrid.js
mariospr: Oh! Forget it... the actual problem is that the popup's state is not being updated, so this is not correct: we should make sure that its state changes to "closed", and then this hack won't longer be needed

js/ui/appDisplay.js
mariospr: Actually, this seems wrong and behind the crash I've been seeing: when the folder changes, the workID created for the actor showing the previous state of the folder is invalid, since such actor has been destroyed by the time this is called.

mariospr: Indeed, this change is behind the backtrace I've been seeing that breaks DnD after the first DnD operation. Dropping this commit seems fine in my tests so far, and I can'te see any problem with a "broken grab", but there's not much extra info in the commit message that clarifies what exactly that is about, so I'm not 100% sure.

js/ui/iconGrid.js
mariospr: I finally found why the onComplete callback was never being called when dragging an icon out of a folder, leaving `this._extraSpaceData` set (messing up the calculations) and the signal not being emitted:

Since such a DnD operation means a change of the GSetting holding the icon grid, what's happening is that a call to `_redisplay()` is happening that causes the icons in the grid that are shifted (because of opening a folder) to be destroyed, which means the in-progress Tweens over them defined in `closeExtraSpace()` will be removed before the transition actually finishes, leaving the onComplete() callback uncalled... therefore the "global" variable `_extraSpaceData` set as it was in the first place.

So, this "trick" works because extraSpaceData is actually never used until you open a folder again, at which point it would be overriden, but it's semantically wrong since you'd be doing that reset + emitting a "space-closed" signal at the time you'd opening a new folder and not before.

I think a better way to do it, even if not 100% correct either, is to always reset `extraSpaceData` after the for loop that creates the tweens in `closeExtraSpace()`, since that what we ensure that it gets always reset regardless of actors being destroyed, and that should not be a problem since `extraSpaceData` is only used from `_calculateDndRow()`, called from `canDropAt()`.

mariospr:
This has become superseeded by https://github.com/endlessm/gnome-shell/pull/193

PR #186, opened 2017-12-04
paswordMode, keyboardManager: expose isLatinLayout method over D-Bus

erikos:
This can only handle layouts that are currently loaded
in the KeyboardManager.


mariospr:
Is this PR still relevant? If so... which ticket does it refer to?

erikos:
Leftover indeed - closing.

PR #187, opened 2017-12-05
Implement password reset salting

senko:
https://phabricator.endlessm.com/T20297

### DIFF
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -37,6 +37,8 @@ const CUSTOMER_SUPPORT_LOCATIONS = [
 
 const CUSTOMER_SUPPORT_GROUP_NAME = 'Customer Support';
 const CUSTOMER_SUPPORT_KEY_EMAIL = 'Email';
+const PASSWORD_RESET_GROUP_NAME = 'Password Reset';
+const PASSWORD_RESET_KEY_SALT = 'Salt';
 
 const AuthPromptMode = {
     UNLOCK_ONLY: 0,
@@ -631,14 +633,25 @@ const AuthPrompt = new Lang.Class({
         // Note: These are not secure random numbers. Doesn't matter. The
         // mechanism to convert a reset code to unlock code is well-known, so
         // who cares how random the reset code is?
-        let resetCode = '';
+        // The fist digit is fixed to "1" as version of the hash code (the zeroth
+        // version had one less digit in the code).
+
+        let resetCode = this._getResetCodeSalt() ? '1' : '';
+
         for (let n = 0; n < _RESET_CODE_LENGTH; n++)
             resetCode = '%s%d'.format(resetCode, GLib.random_int_range(0, 10));
         return resetCode;
     },
 
     _computeUnlockCode: function(resetCode) {
+        let salt = this._getResetCodeSalt();
         let checksum = new GLib.Checksum(GLib.ChecksumType.MD5);
+
+        if (salt) {
+            checksum.update(ByteArray.fromString(salt));
+            checksum.update([0]);
+        }
+
         checksum.update(ByteArray.fromString(resetCode));
 
         let unlockCode = checksum.get_string();
@@ -657,10 +670,23 @@ const AuthPrompt = new Lang.Class({
 
         try {
             return keyFile.get_locale_string(CUSTOMER_SUPPORT_GROUP_NAME,
-                                                             CUSTOMER_SUPPORT_KEY_EMAIL,
-                                                             null);
+                                             CUSTOMER_SUPPORT_KEY_EMAIL,
+                                             null);
         } catch (e) {
             logError(e, 'Failed to read customer support email');
+            return null;
+        }
+    },
+
+    _getResetCodeSalt: function() {
+        let keyFile = this._ensureCustomerSupportFile();
+
+        try {
+            return keyFile.get_locale_string(PASSWORD_RESET_GROUP_NAME,
+                                             PASSWORD_RESET_KEY_SALT,
+                                             null);
+        } catch (e) {
+            logError(e, 'Failed to read password reset salt value');
             return null;
         }
     },
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -29,6 +29,15 @@ const MESSAGE_FADE_OUT_ANIMATION_TIME = 0.5;
 
 const _RESET_CODE_LENGTH = 7;
 
+const CUSTOMER_SUPPORT_FILENAME = 'vendor-customer-support.ini';
+const CUSTOMER_SUPPORT_LOCATIONS = [
+    Config.LOCALSTATEDIR + '/lib/eos-image-defaults/' + CUSTOMER_SUPPORT_FILENAME,
+    Config.PKGDATADIR + '/' + CUSTOMER_SUPPORT_FILENAME
+];
+
+const CUSTOMER_SUPPORT_GROUP_NAME = 'Customer Support';
+const CUSTOMER_SUPPORT_KEY_EMAIL = 'Email';
+
 const AuthPromptMode = {
     UNLOCK_ONLY: 0,
     UNLOCK_OR_LOG_IN: 1
@@ -179,10 +188,10 @@ const AuthPrompt = new Lang.Class({
         this._spinner.actor.show();
         this._defaultButtonWell.add_child(this._spinner.actor);
 
-        this._customerSupportEmail = null;
-
         this._displayingPasswordHint = false;
         this._passwordResetCode = null;
+
+        this._customerSupportKeyFile = null;
     },
 
     _onDestroy: function() {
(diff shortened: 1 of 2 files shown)


senko:
@mariospr here’s my initial go at implementing salting support for the password unlock mechanism.

When you have the time please have a look and let me know if there are changes needed (esp. in the keyfile loading part, where I tried to follow the conventions of the surrounding code but may have missed something).

js/gdm/authPrompt.js
```diff
@@ -600,18 +608,36 @@ const AuthPrompt = new Lang.Class({
...
+        return null;
+    },
+
     _ensureCustomerSupportData: function() {
         if (this._customerSupportEmail)
             return true;
 
-        const CUSTOMER_SUPPORT_FILENAME = 'vendor-customer-support.ini';
         const CUSTOMER_SUPPORT_GROUP_NAME = 'Customer Support';
         const CUSTOMER_SUPPORT_KEY_EMAIL = 'Email';
```
mariospr: I think you should move these other constants to the top as well, for consistency (and to have everything in one place)

js/gdm/authPrompt.js
```diff
@@ -600,18 +608,36 @@ const AuthPrompt = new Lang.Class({
...
 
+        let keyFile = this._loadCustomerSupportFile();
+        if (!keyFile)
+            return false;
```
mariospr: Please `logError()` here as well

js/gdm/authPrompt.js
```diff
@@ -660,7 +660,26 @@ const AuthPrompt = new Lang.Class({
     },
 
     _computeUnlockCode: function(resetCode) {
+        const PASSWORD_RESET_GROUP_NAME = 'Password Reset';
+        const PASSWORD_RESET_KEY_SALT = 'Salt';
+
+        let keyFile = this._loadCustomerSupportFile(),
+            salt = '';
```
mariospr: Please use `let` here as well (don't combine declaration of 2 variables under the same `let`)

js/gdm/authPrompt.js
```diff
@@ -660,7 +660,26 @@ const AuthPrompt = new Lang.Class({
...
+        if (keyFile) {
+            try {
+                salt = keyFile.get_locale_string(PASSWORD_RESET_GROUP_NAME,
+                                                 PASSWORD_RESET_KEY_SALT,
+                                                 null);
+            } catch (e) {
+                logError(e, 'Failed to read customer support data');
```
mariospr: This error should be more specific (e.g. mention that it's about the salt)

senko:
@mariospr thanks for the comments.

I’ve modified the code according to your suggestions (note that I’ve used `log` instead of `logError` in the case where we don’t have an error detail but still want to make it clear what failed).

I didn’t want to rebase the branch to capitalize the commit messages (wasn’t sure if that’s the standard practice here) - please let me know if I should do that (and/or if I should squash fixup commits to clean up the branch before the merge).

mariospr:
> I’ve modified the code according to your suggestions (note that I’ve used log instead of logError in the case where we don’t have an error detail but still want to make it clear what failed).

Thanks, looks good now. Also, good call on the order of the locations, that one I didn't realize on my previous review.

> I didn’t want to rebase the branch to capitalize the commit messages (wasn’t sure if that’s the standard practice here) - please let me know if I should do that (and/or if I should squash fixup commits to clean up the branch before the merge).

Sorry, should have mentioned it before. Yes, it is common practice here to rebase + force-push WIP branches before merging, so that git history is cleaner and shorter. Please do that here if you can too, which should leave this PR with only the 2 original commits.

Last, I still don't see a link to the Phabricator ticket. If the reason for that is that there's no ticket created yet, please go ahead and create a new one. Thanks!

senko:
@mariospr I’ve created the Phabricator task, rebased the commits here and linked the phab task in the related commit. Let me know if anything else is needed!

js/gdm/authPrompt.js
```diff
@@ -600,18 +611,35 @@ const AuthPrompt = new Lang.Class({
...
 
+    _loadCustomerSupportFile: function() {
+        if (this._customerSupportKeyFile)
+            return this._customerSupportKeyFile;
+
+        for (let path of CUSTOMER_SUPPORT_LOCATIONS) {
+            let keyFile = new GLib.KeyFile();
+            try {
+                keyFile.load_from_file(path, GLib.KeyFileFlags.NONE);
+                this._csutomerSupportKeyFile = keyFile;
```
mariospr: `_csutomerSupportKeyFile` -> `_customerSupportKeyFile`. Otherwise the early return above these lines will never be useful

js/gdm/authPrompt.js
```diff
@@ -600,18 +611,35 @@ const AuthPrompt = new Lang.Class({
...
+                return keyFile;
```
mariospr: I'd rather return `this._customerSupportKeyFile` here, so that mistakes like the one I pointed out before do not pass unattended.

js/gdm/authPrompt.js
cosimoc: There will be no error to log though... Since the above function already logs an error, I think we should simply return here if it fails.

senko:
Good catch, thanks!

Reworded & fixed as per your comments.

js/gdm/authPrompt.js
mariospr: That was the original code and I suggested logging here, but you're right it's logging anyway in the function so it's probably redundant. @senko Sorry for the back and forth, I think @cosimoc is right and we should not log anything here.

js/gdm/authPrompt.js
mariospr: Still pending to address

js/gdm/authPrompt.js
senko: Changed per suggestion, thanks @cosimoc (hi, btw, long time no see! :-)

js/gdm/authPrompt.js
```diff
@@ -600,18 +613,33 @@ const AuthPrompt = new Lang.Class({
         this.emit('cancelled');
     },
 
+    _loadCustomerSupportFile: function() {
```
cosimoc: Optional comment, but if this function was called e.g. `_ensureCustomerSupportKeyFile()`, and simply tried to create `this._customerSupportKeyFile` when it did not exist without returning anything, you could swap out the `if (keyFile)` checks in the callers for `if (this._customerSupportKeyFile)` checks, with a net reduction of lines of code :)

senko: Considering the search path involves presumably read-only locations (`PKGDATADIR` and `LOCALSTATEDIR` which I assume is reset on reboot?) does it make sense to try and create this file?

cosimoc: @senko With "create" I mean "create the instance", not the file on disk.
Basically since the callers of `_loadCustomerSupportFile()` already need to check its return value, and that function already checks the value of the internal `this._customerSupportKeyFile` state variable, I'd remove the latter in favor of checking the internal field in the callers.

mariospr: @senko Ping? I think this is the last pending review comment left to be addressed.

Also, I'm tagging @ramcq here in case he want to comment too, as he explicitly mentioned last week he'd like to take a look to the way the salt is being implemented with this PR.

senko:
Here is the refactored code based on the suggestion.

However, this time I’m not completely satisfied with it. The current (both in the existing code and in this PR) behaviour is to not show the password reset dialog if the support email can’t be loaded.

This is almost a non-issue (since the default v-c-s.ini is shipped with gnome-shell so it should always be found), *but* that shouldn’t stop us from handling its nonexistence carefully. As it is, the current behaviour looks broken (you click “forgot password” but don’t see anything).

My suggestion here would be to either not show the “forgot password” link if the support mail can’t be loaded, or to fallback to something (eg. just saying “contact your customer support). Or, if you think I’m overly complicating things, just ignore the issue and leave it as is (as mentioned, I don’t have an absolute conviction it must be dealt with).

ramcq:
Hey @senko - I think I mentioned in e-mail but maybe was lost in the noise - in order to prevent the wrong tool being used to generate codes for machines with salts, and never answering successfully (preventing errors on user and on the CS agent side), I'd like us to change the format/syntax of salted codes so the new salted code isn't valid in the old tool. Like passwd crypts can be $1$ or $2$ depending on the hash algo used - maybe let's generate an additional digit for the challenge code (prefix with 1) in the case that the machine is using a salt. The webapp can accept a 1 digit shorter code and prefix with 0 to indicate we have the old unsalted hash, otherwise a longer code with the prefix 1 means we're using this combination of salt and algorithm, etc.

cosimoc:
@senko Your suggestion of hiding the link when the email can't be loaded makes a lot of sense to me. Please go ahead and implement it if it's not too hard, but as you say, that is not really a case we care about (even more so considering that `/usr` is read-only on Endless, so if that file is missing, chances are we have bigger problems...).

cosimoc:
Otherwise this PR looks good to me as-is and I think can be merged. Leaving here in case @mariospr wants to give a last look.

senko:
@cosimoc I’ve implemented the additional logic in 2ccfa04, I believe it doesn’t complicate things too much - let me know what you think.

@ramcq sorry, I mis-parsed your earlier comment to mean you wouldn’t like to have the tool generate different output on same (legacy) input ie backwards compatibility. Thanks for clarifying, I’ve implemented this in c191969.

I’ve added these as separate commits for now so it’s easy to review, if/when you’re happy let me know and I’ll rebase and do other needed cleanup to get it ready for the merge again.

ramcq:
On 15/12/17 10:14, Senko Rašić wrote:
> @ramcq <https://github.com/ramcq> sorry, I mis-parsed your earlier
> comment to mean you wouldn’t like to have the tool generate different
> output on same (legacy) input ie backwards compatibility. Thanks for
> clarifying, I’ve implemented this in c191969
> <https://github.com/endlessm/gnome-shell/commit/c191969ec51ec8c55ab468a2123def3867df55f0>.

I was thinking more like:

_RESET_CODE_LENGTH = 7

...

code_length = _RESET_CODE_LENGTH;
if (we have a salt)
  {
    code_length++;
    resetCode = '1';
  }

...

So the code is 7 digits with no salt, or 8 digits (prefixed by 1) if we do.

js/gdm/authPrompt.js
```diff
@@ -669,18 +669,23 @@ const AuthPrompt = new Lang.Class({
...
 
-    _showPasswordResetPrompt: function() {
+    _getCustomerSupportEmail: function() {
         let keyFile = this._ensureCustomerSupportFile();
-        let customerSupportEmail = null;
 
         try {
-            customerSupportEmail = keyFile.get_locale_string(CUSTOMER_SUPPORT_GROUP_NAME,
+            return keyFile.get_locale_string(CUSTOMER_SUPPORT_GROUP_NAME,
                                                              CUSTOMER_SUPPORT_KEY_EMAIL,
```
mariospr: Indentation seems off for the parameters

mariospr:
> Otherwise this PR looks good to me as-is and I think can be merged. Leaving here in case @mariospr wants to give a last look.

Looks good to me now, just pointed out a minor styling issue and waiting now for the comment from @ramcq to be addressed. After that, I think it would be good to merge.

cosimoc:
@senko Looks great to me, thanks!

mariospr:
Looks good to me too, @senko. Please squash the last 2 commits in the first ones, as you mentioned before and this should be ready to merge. Thanks!

senko:
Squashed the two commits, ready for the merge.

js/gdm/authPrompt.js
mariospr: This indentation is still off

mariospr: Ah! It's fixed in the following commit. Nevermind then, I will probably "fix" while doing the next rebase. Let's not go for another review cycle because of this nit :-)

mariospr:
Looks good. Merged!

PR #188, opened 2017-12-06
Don't show the splash screen for an app that's already running

mariospr:
This was already being checked when activating and app over the
org.gnome.Shell.AppLauncher D-Bus interface, but surprisingly
enough not when activating an app from the search results.

To fix this and make it less error prone, incorporate the check
into _shouldShowSplash(), and remove it from anywhere else.

https://phabricator.endlessm.com/T20114

### DIFF
--- a/js/ui/appActivation.js
+++ b/js/ui/appActivation.js
@@ -33,6 +33,10 @@ const LAUNCH_MAXIMIZED_DESKTOP_KEY = 'X-Endless-LaunchMaximized';
 function _shouldShowSplash(app) {
     let info = app.get_app_info();
 
+    // Don't show the splash screen if the app is already running.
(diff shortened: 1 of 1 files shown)


js/ui/appActivation.js
```diff
@@ -155,7 +157,7 @@ const AppActivationContext = new Lang.Class({
             this._abort = true;
     },
 
-    showSplash: function() {
+    showSplashIfNeeded: function() {
```
cosimoc: I would not rename this method, since even before your changes it did not show the splash when not needed.

mariospr: Yes, I know. And I found that particularly confusing so I thought it would be nice to "fix" that.

PR #189, opened 2017-12-07
passwordMode, InputSourceManager: expose Enabled property over D-Bus

erikos:
This makes the passwordMode feature available outside of the shell and can be used from gnome-initial-setup and gnome-control-center to check if a kbd layout is latin and add the fallback layout accordingly.

https://phabricator.endlessm.com/T18598

### DIFF
--- a/js/ui/status/keyboard.js
+++ b/js/ui/status/keyboard.js
@@ -22,6 +22,12 @@ const Util = imports.misc.util;
 const INPUT_SOURCE_TYPE_XKB = 'xkb';
 const INPUT_SOURCE_TYPE_IBUS = 'ibus';
 
+const InputSourceManagerIface = '<node> \
+<interface name="org.gnome.Shell.InputSourceManager"> \
+<property name="PasswordModeEnabled" type="b" access="readwrite" /> \
+</interface> \
+</node>';
+
 const LayoutMenuItem = new Lang.Class({
     Name: 'LayoutMenuItem',
     Extends: PopupMenu.PopupBaseMenuItem,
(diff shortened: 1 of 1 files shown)


js/ui/status/keyboard.js
```diff
@@ -22,6 +22,15 @@ const Util = imports.misc.util;
 const INPUT_SOURCE_TYPE_XKB = 'xkb';
 const INPUT_SOURCE_TYPE_IBUS = 'ibus';
 
+const InputSourceManagerIface = '<node> \
+<interface name="org.gnome.Shell.InputSourceManager"> \
+<method name="SetPasswordModeEnabled"> \
+ <arg name="enabled" type="b" direction="in"/> \
+</method> \
+<property name="PasswordModeEnabled" type="b" access="read" /> \
```
cosimoc: This could be a `readwrite` property instead, which would save you from writing a separate method just to set it.

erikos: I had this read-write before. I tried to use g_dbus_proxy_set_cached_property like in other places for it, but that was not received by the shell. So I used code I found for Enabling Metrics 'SetEnabled' in the control center https://github.com/endlessm/gnome-control-center/blob/master/panels/privacy/cc-privacy-panel.c#L907.

cosimoc: `g_dbus_proxy_set_cached_property()` won't help, as it will only set the value of the local cached property of the proxy, not send it over the wire.
You will need to call the `Set` method on the `org.freedesktop.DBus.Properties` interface from the client side. (Or alternatively use `gdbus-codegen` to generate a class-specific proxy that actually has a property setter that will call that for you, but that's a lot of yak shaving.)

cosimoc:
LGTM

PR #190, opened 2017-12-08
T20325

GeorgesStavracas:
This PR adds three new CSS gradient-related properties to St:

 * background-gradient-radius
 * background-gradient-stop
 * background-gradient-position

And further adapts the CSS to use the new gradient properties to implement the curved Endless Toggle Button.

https://phabricator.endlessm.com/T20325

### DIFF
--- a/src/st/st-theme-node-drawing.c
+++ b/src/st/st-theme-node-drawing.c
@@ -590,11 +590,25 @@ create_cairo_pattern_of_background_gradient (StThemeNode *node,
     pattern = cairo_pattern_create_linear (0, 0, width, 0);
   else
     {
-      gdouble cx, cy;
+      gdouble cx, cy, radius;
 
-      cx = width / 2.;
-      cy = height / 2.;
-      pattern = cairo_pattern_create_radial (cx, cy, 0, cx, cy, MIN (cx, cy));
+      if (node->background_gradient_position_set)
+        {
+          cx = node->background_gradient_position_x;
+          cy = node->background_gradient_position_y;
+        }
+      else
+        {
+          cx = width / 2.;
+          cy = height / 2.;
+        }
+
+      if (node->background_gradient_radius != -1)
+        radius = node->background_gradient_radius;
+      else
+        radius = MIN (width / 2., height / 2.);
+
+      pattern = cairo_pattern_create_radial (cx, cy, 0, cx, cy, radius);
     }
 
   cairo_pattern_add_color_stop_rgba (pattern, 0,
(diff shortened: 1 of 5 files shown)


src/st/st-theme-node-drawing.c
```diff
@@ -590,18 +590,44 @@ create_cairo_pattern_of_background_gradient (StThemeNode *node,
...
+      else
+        {
+          cx = width / 2.;
+          cy = height / 2.;
+        }
+
+      if (node->background_gradient_radius != -1)
+        radius = node->background_gradient_radius;
+      else
+        radius = MIN (width / 2., height / 2.);
```
mariospr: Shouldn't this be `MIN (cx, cy)`  ?

src/st/st-theme-node-drawing.c
```diff
@@ -590,18 +590,44 @@ create_cairo_pattern_of_background_gradient (StThemeNode *node,
...
+      pattern = cairo_pattern_create_radial (cx, cy, 0, cx, cy, radius);
     }
 
   cairo_pattern_add_color_stop_rgba (pattern, 0,
                                      node->background_color.red / 255.,
                                      node->background_color.green / 255.,
                                      node->background_color.blue / 255.,
                                      node->background_color.alpha / 255.);
+
+  /* Add an intermediate stop when any is set */
```
mariospr: Is there a way you can generalize this to support more than one stop from the CSS?

src/st/st-theme-node-private.h
```diff
@@ -44,6 +44,12 @@ struct _StThemeNode {
   StGradientType background_gradient_type;
   ClutterColor background_gradient_end;
 
+  guint background_gradient_position_set : 1;
+  int background_gradient_position_x;
+  int background_gradient_position_y;
+  int background_gradient_stop;
```
mariospr: I would call this `_color_stop`, since that's what it is, and how both Cairo and CSS calls those types of stops. Also, as mentioned before, it would be nice if this could support more than one stop, since that might make the patch more upstream friendly (although I'm happy to leave that for a separate PR, if that works better)

mariospr: Hmm... actually my comment is not correct: this represents the position relative (between 0.0 and 1.0) to the width of the element it's being applied on top, so color-stop might not be a good name. Maybe stop-position?

Or even better, make it  as in [CSS's radial-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/radial-gradient) and accept the color + the position of the stop, instead of using `node->background_color`.

src/st/st-theme-node-drawing.c
GeorgesStavracas: No. If the `background_gradient_radius == -1`, the code strictly falls back to the previous behavior.

I did this solely to reduce the diff from the theme, since otherwise the theme itself would need to be adapted, and having bigger changes on the edge of a rebase feels like a bad decision.

mariospr: Ok

PR #191, opened 2017-12-12
DiscoveryFeed: design improvements for the reveal button

erikos:
Create a new button consistent of a bar (poor people window peek) and the tile.

https://phabricator.endlessm.com/T19354


js/ui/discoveryFeedButton.js
```diff
@@ -54,21 +50,42 @@ function maybeCreateButton() {
...
+        this._tile = new St.Button({ name: 'discovery-feed-tile',
+                                    child: new St.Icon({ style_class: 'discovery-feed-tile-icon' }),
+                                    style_class: 'discovery-feed-tile',
+                                    visible: _primaryMonitorWidthPassesThreshold() });
+        this.add(this._tile, { x_fill: false,
+                               x_align: St.Align.MIDDLE,
+                               expand: true });
+
+        this._bar.connect('event', Lang.bind(this, this._onEvent));
+        this._tile.connect('event', Lang.bind(this, this._onEvent));
```
smspillaz: We don't appear to re-use `this._bar` and `this._tile` anywhere outside this function, so prefer to just make them variables inside this function.

js/ui/discoveryFeedButton.js
```diff
@@ -54,21 +50,42 @@ function maybeCreateButton() {
...
+        this._bar = new St.Button({ name: 'discovery-feed-bar',
+                                    child: new St.Icon({ style_class: 'discovery-feed-bar-icon' }),
+                                    style_class: 'discovery-feed-bar',
+                                    visible: _primaryMonitorWidthPassesThreshold() });
+        this.add(this._bar);
+
+        this._tile = new St.Button({ name: 'discovery-feed-tile',
+                                    child: new St.Icon({ style_class: 'discovery-feed-tile-icon' }),
+                                    style_class: 'discovery-feed-tile',
+                                    visible: _primaryMonitorWidthPassesThreshold() });
```
smspillaz: Instead of controlling the visibility state of both elements separately with the same function, is it possible to move this visibility check into the parent?

js/ui/discoveryFeedButton.js
```diff
@@ -38,7 +38,10 @@ function _checkIfDiscoveryFeedEnabled() {
 function maybeCreateButton() {
     if (_checkIfDiscoveryFeedEnabled()) {
         let discoveryFeedButton = new DiscoveryFeedButton();
-        discoveryFeedButton.connect('clicked', Lang.bind(this, function() {
+        discoveryFeedButton.tile.connect('clicked', Lang.bind(this, function() {
```
mariospr: Not a big deal, but I think we should start arrow functions whenever possible instead of the -now no longer needed- Lang.bind() idiom

js/ui/discoveryFeedButton.js
```diff
@@ -63,25 +56,36 @@ const DiscoveryFeedButton = new Lang.Class({
         this.parent({ vertical: true,
                       style_class: 'discovery-feed-button' });
 
-        this.bar = new St.Button({ name: 'discovery-feed-bar',
-                                   child: new St.Icon({ style_class: 'discovery-feed-bar-icon' }),
-                                   style_class: 'discovery-feed-bar',
-                                   visible: _primaryMonitorWidthPassesThreshold() });
-        this.add(this.bar);
+        this._bar = new St.Button({ name: 'discovery-feed-bar',
```
mariospr: Renaming this.bar to this._bar here seems to unnecessarily pollute the diff, since you could have made it this._bar right from the start, in the previous commit.

js/ui/discoveryFeedButton.js
```diff
@@ -63,25 +56,36 @@ const DiscoveryFeedButton = new Lang.Class({
...
+                                    child: new St.Icon({ style_class: 'discovery-feed-bar-icon' }),
+                                    style_class: 'discovery-feed-bar',
+                                    visible: _primaryMonitorWidthPassesThreshold() });
```
mariospr: Some comment applies to this leading space added here, it could have been aded in the previous commit too

js/ui/discoveryFeedButton.js
```diff
@@ -63,25 +56,36 @@ const DiscoveryFeedButton = new Lang.Class({
...
+        this.add(this._bar);
 
-        this.tile = new St.Button({ name: 'discovery-feed-tile',
+        this._tile = new St.Button({ name: 'discovery-feed-tile',
```
mariospr: Same comment for this.title -> this._title

js/ui/discoveryFeedButton.js
mariospr: +1. Since the title and the bar are added as children of the box layout, there's no need to keep a reference to them

smspillaz: Did we come to a resolution on this @erikos  ?

smspillaz: Err, whoops, my bad, I see that we use them below.

js/ui/discoveryFeedButton.js
```diff
@@ -54,21 +50,54 @@ function maybeCreateButton() {
...
+                                     style_class: 'discovery-feed-tile' });
+        this.add(this._tile, { x_fill: false,
+                               x_align: St.Align.MIDDLE,
+                               expand: true });
+
+        this._bar.connect('event', Lang.bind(this, this._onEvent));
+        this._tile.connect('event', Lang.bind(this, this._onEvent));
+
+        this._bar.connect('notify::hover', Lang.bind(this, this._onHoverChanged));
+        this._tile.connect('notify::hover', Lang.bind(this, this._onHoverChanged));
```
smspillaz: For this event and the above, is it possible to connect to the signal on the parent container rather than on both the bar and the tile independently? It seems like clicking on either one should have the same behaviour.

js/ui/discoveryFeedButton.js
```diff
@@ -54,21 +50,54 @@ function maybeCreateButton() {
...
+
         Main.layoutManager.connect('monitors-changed', Lang.bind(this, function() {
             this.visible = _primaryMonitorWidthPassesThreshold();
         }));
     },
 
+    _onEvent: function(actor, event) {
+        if (event.type() == Clutter.EventType.TOUCH_BEGIN ||
+            event.type() == Clutter.EventType.BUTTON_PRESS)
```
smspillaz: I think you may want to use `TOUCH_END` and `BUTTON_RELEASE` here - a "click" or "touch" is only fully completed once one of those two things happens.

js/ui/discoveryFeedButton.js
erikos: Yes, that has changed :)

js/ui/discoveryFeedButton.js
erikos: Not really. We want the hover on those specific items, the container would include the to the user invisible parts as the hover area. That is why we have the separation between tile and bar actually.

js/ui/discoveryFeedButton.js
erikos: We have both behaviours in the shell. There was a specific request to make this on button_press see https://phabricator.endlessm.com/T19354#516229. This is the same behaviour as with the Endless button in the taskbar: https://github.com/endlessm/gnome-shell/blob/master/js/ui/endlessButton.js#L79

PR #192, opened 2017-12-13
Bring back DnD operations inside folders

mariospr:
Rebase of the previous work started in this branch in August, incorporating the changes done by @GeorgesStavracas in https://github.com/endlessm/gnome-shell/pull/165 and (hopefully) finishing it by simplifying it and fixing all the issues observed in that PR.

Main changes with regard to https://github.com/endlessm/gnome-shell/pull/165:
  * Removed the animations performed when "reshuffling" icons on the desktop, as a result of DnD. Those animations were still pretty confusing as implemented in the previous PR (source of the animation did not match where the icon was being dropped from), and on top of that were inconsistent since no animations were shown when DnD'ing inside folders, so I'll file a new ticket for those and keep them out of this one, which should be focused on re-enabling DnD inside folders.
  * Removed the hack to initialize the extra space taken by folders by making sure the estate is always reset after closing a popup. Still a workaround, but it seems to work better since it makes sure the state is correct right after closing a popup, without waiting until another folder is opened again.
  * Fixed the problem that was printing the backtrace mentioned above by making sure that any callback from which we would run a deferred job gets disconnected when the folder icon's actor is destroyed.
  * Removed the animation that was supposed to happen on startup, as it was causing trouble and was not related to this story anyway.
  * Cleaned up the commit history, by making smaller and clearer (hopefully) commits, removing everything that is not actually needed.

This PR replaces #165, so closing that one.

https://phabricator.endlessm.com/T18012

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -105,6 +105,9 @@ const EOS_DRAG_OVER_FOLDER_OPACITY = 128;
 
 const EOS_REPLACED_BY_KEY = 'X-Endless-Replaced-By';
 
+const EOS_NEW_ICON_ANIMATION_TIME = 0.5;
+const EOS_NEW_ICON_ANIMATION_DELAY = 0.7;
+
 function _getCategories(info) {
     let categoriesStr = info.get_categories();
     if (!categoriesStr)
@@ -218,6 +221,25 @@ const BaseAppView = new Lang.Class({
         return [movedList, removedList];
     },
 
+    _findAddedIcons: function() {
+        let oldItemLayout = this._allItems.map(function(icon) { return icon.getId(); });
+        if (oldItemLayout.length === 0)
+            return [];
+
+        let newItemLayout = this.getLayoutIds();
+        newItemLayout = this._trimInvisible(newItemLayout);
+
+        let addedIds = [];
+        for (let newItem of newItemLayout) {
+            let oldItemIdx = oldItemLayout.indexOf(newItem);
+
+            if (oldItemIdx < 0)
+                addedIds.push(newItem);
+        }
+
+        return addedIds;
+    },
+
     iconsNeedRedraw: function() {
         // Check if the icons moved around
         let [movedList, removedList] = this._findIconChanges();
@@ -266,6 +288,10 @@ const BaseAppView = new Lang.Class({
     },
 
     _loadApps: function() {
+        let addedIds = this._findAddedIcons();
+
+        this.removeAll();
+
         let ids = this.getLayoutIds();
 
         for (let itemId of ids) {
@@ -276,6 +302,9 @@ const BaseAppView = new Lang.Class({
             let icon = this._createItemIcon(item);
             if (!icon)
                 continue;
+
+            if (addedIds.indexOf(itemId) != -1)
+                icon.scheduleScaleIn();
 
             this.addItem(icon);
         }
@@ -293,7 +322,6 @@ const BaseAppView = new Lang.Class({
         if (!forceRedisplay && !this.iconsNeedRedraw())
             return;
 
-        this.removeAll();
         this._loadApps();
     },
 
@@ -1883,6 +1911,8 @@ const ViewIcon = new Lang.Class({
         this.canDrop = false;
         this.blockHandler = false;
 
+        this._scaleInId = 0;
+
         // Might be changed once the createIcon() method is called.
         this._iconSize = IconGrid.ICON_SIZE;
         this._iconState = ViewIconState.NORMAL;
@@ -1953,6 +1983,8 @@ const ViewIcon = new Lang.Class({
     },
 
     _onDestroy: function() {
+        this._unscheduleScaleIn();
+
         this.actor._delegate = null;
         this._removeMenuTimeout();
     },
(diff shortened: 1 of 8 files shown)


js/ui/appDisplay.js
```diff
@@ -173,13 +173,121 @@ const BaseAppView = new Lang.Class({
...
+            return IconGridLayout.layout.iconIsFolder(id) || appSystem.lookup_app(id) || (id == EOS_APP_CENTER_ID);
+        });
+    },
+
+    _findIconChanges: function() {
+        let oldItemLayout = this._allItems.map(function(icon) { return icon.getId(); });
+        let newItemLayout = this.getLayoutIds();
+        newItemLayout = this._trimInvisible(newItemLayout);
+
+        let movedList = {};
```
cosimoc: Using a `Hash.Map()` should make the code that uses the moved list a bit easier

js/ui/appDisplay.js
```diff
@@ -173,13 +173,121 @@ const BaseAppView = new Lang.Class({
...
+
+            // currentIcon is a ViewIcon (AppIcon or FolderIcon).
+            let oldIconInfo = currentIcon.getIcon();
+
+            // item is either a ShellApp or a  ShellDesktopDirInfo.
+            let newIconInfo = item.get_icon();
+
+            // The icon image changed
+            if (newIconInfo && !newIconInfo.equal(oldIconInfo))
+                return true;
```
cosimoc: The old code of this function (https://github.com/endlessm/eos-desktop/blob/master/js/ui/appDisplay.js#L351) also checked whether the name changed. I don't think we still need the bits that use the `customName` field, since that is not set anymore, but I believe we do still need the simple name check from here: https://github.com/endlessm/eos-desktop/blob/master/js/ui/appDisplay.js#L358

js/ui/appDisplay.js
```diff
@@ -609,8 +720,28 @@ const AllView = new Lang.Class({
...
+    },
+
+    getLayoutIds: function() {
+        let layoutIds = this.parent();
+
+        if (!global.settings.get_boolean(EOS_ENABLE_APP_CENTER_KEY))
+            return layoutIds;
+
+        let appSys = Shell.AppSystem.get_default();
+        if (appSys.lookup_app(EOS_APP_CENTER_ID)) {
```
cosimoc: I'd remove this check here tbh. If gnome-software somehow isn't installed, we have bigger problems already...

js/ui/appDisplay.js
```diff
@@ -2003,19 +2161,20 @@ const FolderIcon = new Lang.Class({
...
         let iconParams = { createIcon: Lang.bind(this, this._createIcon),
                            setSizeManually: false,
                            editable: true };
-        this.id = id;
+
         this.name = '';
         this._parentView = parentView;
 
-        this._dirInfo = Shell.DesktopDirInfo.new(id);
+        this.id = dirInfo.get_id();
```
cosimoc: We already have `getId()` methods on the icons, so this isn't really needed, right? I would change the users of this public field to use the method instead.

mariospr: It's not needed for us, but `this.id` for `FolderIcon` is something defined by upstream already, so I think I should keep this line (and keep any caller intact) to keep things as similar as possible.

js/ui/appDisplay.js
mariospr: I think it's a good idea to leave this check, since the App Center might not be installed in some images such as eosinstaller (see https://github.com/endlessm/gnome-shell/commit/1140d786a40085807da48e73f04714df2e83ca2b#diff-7dfb7548c6314bf99dd14ab6cb89bc5a)

js/ui/appDisplay.js
```diff
@@ -2444,10 +2578,18 @@ const AppIcon = new Lang.Class({
...
+    getId: function() {
+        return this.app.get_id();
+    },
+
     getName: function() {
         return this.name;
     },
 
+    getIcon: function() {
+        return this.app.get_app_info().get_icon();
```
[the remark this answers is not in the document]
mariospr: Sure!

js/ui/appDisplay.js
mariospr: Good catch. I'll put that extra check in, thanks

js/ui/appDisplay.js
mariospr: I think you mean JS's Map, right? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

js/ui/iconGrid.js
```diff
@@ -1208,14 +1210,26 @@ const PaginatedIconGrid = new Lang.Class({
...
                              });
         }
+
+        // This is not entirely correct since we should ideally do
+        // this on onComplete, but in our current implementation of
+        // folders and DnD it can happen that the actor of the icons
+        // we are moving back into their position get destroyed before
+        // the tween completes (e.g. dragging icons out of folders),
+        // meaning that the onComplete callback is never called, and
+        // leaving this in an inconsistent state.
```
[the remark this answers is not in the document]
cosimoc: This still needs addressing, at least the slow down factor part.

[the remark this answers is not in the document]
mariospr: It's very odd indeed. Sorry I didn't see this comment somehow, it's fixed now.

PR #193, opened 2017-12-13
appDisplay: Add a scale-in animation when adding new icons to the grid

mariospr:
This brings back the nice "bounce" animation we had in the old version
of the shell, that is run whenever a new icon is added to the icon
grid (e.g. Added to desktop from the App Center).

PS: Note that this PR goes against T18012 instead of master, as it depends on it to work.

https://phabricator.endlessm.com/T20087

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -105,6 +105,9 @@ const EOS_DRAG_OVER_FOLDER_OPACITY = 128;
 
 const EOS_REPLACED_BY_KEY = 'X-Endless-Replaced-By';
 
+const EOS_NEW_ICON_ANIMATION_TIME = 0.5;
+const EOS_NEW_ICON_ANIMATION_DELAY = 0.7;
+
 function _getCategories(info) {
     let categoriesStr = info.get_categories();
     if (!categoriesStr)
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -218,6 +221,24 @@ const BaseAppView = new Lang.Class({
         return [movedList, removedList];
     },
 
+    _findAddedIcons: function() {
+        let oldItemLayout = this._allItems.map(function(icon) { return icon.getId(); });
+        if (oldItemLayout.length === 0) return [];
```
cosimoc: Coding style: put the return on a new line.

js/ui/appDisplay.js
```diff
@@ -272,6 +297,9 @@ const BaseAppView = new Lang.Class({
             if (!icon)
                 continue;
 
+            if (addedIds && addedIds.indexOf(itemId) != -1)
```
cosimoc: Since `this._findAddedIcons()` always returns an array, you should not need the first condition here.

js/ui/appDisplay.js
mariospr: Oops

js/ui/appDisplay.js
mariospr: Good point

PR #194, opened 2017-12-15
Disable the background menu when in initial-setup mode

mariospr:
Otherwise it's possible to launch the App Center from it and,
with some additional effort, ending up executing anything even
before the FBE is finished.

https://phabricator.endlessm.com/T20471

### DIFF
--- a/js/ui/backgroundMenu.js
+++ b/js/ui/backgroundMenu.js
@@ -44,6 +44,11 @@ const BackgroundMenu = new Lang.Class({
 });
 
 function _addBackgroundMenuFull(actor, clickAction, layoutManager) {
+    // We don't want the background menu enabled on the desktop
+    // during the FBE, or in any mode without an overview, fwiw.
+    if (!Main.sessionMode.hasOverview)
+        return;
+
     // Either the actor or the action has to be defined
     if (!actor && !clickAction)
         return;


js/ui/layout.js
```diff
@@ -361,7 +361,11 @@ const LayoutManager = new Lang.Class({
                                                            monitorIndex: monitorIndex });
 
         bgManager.connect('changed', Lang.bind(this, this._addBackgroundMenu));
-        this._addBackgroundMenu(bgManager);
+
+        // We don't want the background menu enabled on the desktop
+        // during the FBE, or in any mode without an overview, fwiw.
+        if (Main.sessionMode.currentMode.hasOverview)
```
cosimoc: You can simply use `Main.sessionMode.hasOverview`, because the current mode properties are propagated to the object: https://github.com/endlessm/gnome-shell/blob/master/js/ui/sessionMode.js#L242

cosimoc: https://github.com/endlessm/gnome-shell/blob/fc24fb44a2f3b3c9862f6807e5ffd175dd65c1c7/js/ui/backgroundMenu.js#L46 
This is probably a better place to put this check; that's where it was in eos-desktop, and there's another caller for the method in `backgroundMenu`.

mariospr: Both comments make sense to me (and explain why this was not a problem in the previous shell).

Updated

PR #195, opened 2017-12-15
T20327

GeorgesStavracas:
This PR implements the bottom bar improvements from the proposed mockups.

Changes:
 * Swap clock and user avatar menu
 * Change the aggregate menu button style (only on user sessions; GDM looks the same)

https://phabricator.endlessm.com/T20327

### DIFF
--- a/data/theme/gnome-shell.css
+++ b/data/theme/gnome-shell.css
@@ -2054,6 +2054,11 @@ stage {
     -minimum-hpadding: 6px;
 }
 
+#panelRight .panel-button {
+    -natural-hpadding: 10px;
+    -minimum-hpadding: 10px;
+}
+
 .popup-menu.panel-menu {
     -boxpointer-gap: 0px;
     margin-bottom: 0px;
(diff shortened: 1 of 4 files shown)


data/theme/gnome-shell.css
```diff
@@ -690,8 +690,8 @@ StScrollBar {
       -panel-corner-background-color: transparent;
       -panel-corner-border-color: transparent; }
   #panel .panel-button {
-    -natural-hpadding: 12px;
-    -minimum-hpadding: 6px;
+    -natural-hpadding: 10px;
+    -minimum-hpadding: 10px;
```
mariospr: Please move these changes to an Endless specific section (after the `/* Endless-specific overrides below this point *` comment below), to help with future rebases.

data/theme/gnome-shell.css
```diff
@@ -2166,8 +2168,17 @@ stage {
...
+.aggregate-menu-indicators-box {
+    padding: 6px 4px;
+    background-color: rgba(255, 255, 255, 0.1);
+    border-radius: 3px;
+}
+
 #panel .endless-button,
 #panel.solid .endless-button {
+    -natural-hpadding: 12px;
+    -minimum-hpadding: 6px;
```
mariospr: I understand these changes are required here to explicitly state a different padding for the Endless button compared to the icons in the right side of the taskbar, however... 

Please consider either doing it the other way around (leaving the default padding for ` #panel .panel-button` -see my previous comment- unchanged + explicitly setting it to 10px for the right side of the panel), or doing it this way but then splitting the changes to `#panel.solid .endless-button` to a separate commit, so that we can squash it along with other Endless button-related changes in future rebases.

Either way, please do not modify the CSS above the `/* Endless-specific overrides below this point */` comment above.

data/theme/gnome-shell.css
GeorgesStavracas: Thanks, I didn't actually notice we had an Endless-specific section!

PR #201, opened 2018-01-05
fixup! shell-app-system: Report an event when an application opens/closes

wjt:
If a window opens for which the Shell can't find an app ID, Shell logs
the following message to the journal:

Jan 05 18:09:11 endless gnome-shell[3003]: g_variant_new_string: assertion 'string != NULL' failed

This is because g_variant_new ("s", NULL) ultimately calls
g_variant_new_string (NULL) which is illegal. The
SHELL_APP_STATE_STOPPED branch has this NULL check, but it was missing
in the SHELL_APP_STATE_RUNNING branch.

While we are here, let's call g_variant_new_string() directly.

### DIFF
--- a/src/shell-app-system.c
+++ b/src/shell-app-system.c
@@ -565,12 +565,13 @@ _shell_app_system_notify_app_state_changed (ShellAppSystem *self,
   switch (state)
     {
     case SHELL_APP_STATE_RUNNING:
-      {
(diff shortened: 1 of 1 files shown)


wjt:
@mariospr I started writing a ticket for this but given you've already started the rebase, and given there's no user-visible problem, we can just fix it in 3.4?

PR #204, opened 2018-01-12

mariospr:
Closing the speedwagon window won't stop an application from loading
unless it's already RUNNING and with at least one window visible, since
that's what shell_app_request_quit() relies on to close an application,
using meta_window_delete().

Thus, we need to defer quitting the application for which the splash
screen has been closed until it reaches that RUNNING state and
has at least its main window added, at which point it's safe to call
meta_window_delete() to terminate the process.

### DIFF
--- a/src/shell-app.c
+++ b/src/shell-app.c
@@ -85,6 +85,7 @@ struct _ShellApp
                           */
 
   ShellAppRunningState *running_state;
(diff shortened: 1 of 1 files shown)


mariospr:
Closed in favour of https://github.com/endlessm/gnome-shell/pull/205

PR #205, opened 2018-01-12

mariospr:
Stop relying on the SplashClosed D-Bus signal from speedwagon, which causes race conditions we'd need to deal with since such signal arrives normally after the application being launched moves to a STOPPED state, as per the speedwagon's window being explicitly closed.

Instead, keep the signal callback for 'app-state-changed' alive for a little longer so that we can defer quitting an application whose splash screen has been manually closed until the app reaches the RUNNING state.

### DIFF
--- a/js/ui/appActivation.js
+++ b/js/ui/appActivation.js
@@ -61,7 +61,6 @@ const AppActivationContext = new Lang.Class({
     _init: function(app) {
         this._app = app;
         this._abort = false;
-        this._cancelled = false;
 
         this._splash = null;
 
@@ -147,16 +146,6 @@ const AppActivationContext = new Lang.Class({
         Main.overview.hide();
     },
 
-    cancelSplash: function() {
-        this._cancelled = true;
-        this._clearSplash();
-        // If application doesn't quit very likely is because it
-        // didn't reach running state yet; so wait for it to
-        // finish
-        if (!this._app.request_quit())
-            this._abort = true;
-    },
-
     showSplash: function() {
         if (!_shouldShowSplash(this._app))
             return;
@@ -166,9 +155,7 @@ const AppActivationContext = new Lang.Class({
         if (Main.overview.visible)
             this._hideWindows();
 
-        this._cancelled = false;
         this._splash = new SpeedwagonSplash(this._app);
-        this._splash.connect('close-clicked', Lang.bind(this, this.cancelSplash));
         this._splash.show();
 
         // Scale the timeout by the slow down factor, because otherwise
@@ -191,6 +178,11 @@ const AppActivationContext = new Lang.Class({
     _clearSplash: function() {
         this._resetWindowsVisibility();
 
+        if (this._appStateId) {
+            this._appSystem.disconnect(this._appStateId);
+            this._appStateId = 0;
+        }
+
         if (this._splash) {
             this._splash.rampOut();
             this._splash = null;
@@ -200,7 +192,7 @@ const AppActivationContext = new Lang.Class({
     _maybeClearSplash: function() {
         // Clear the splash only when we've waited at least 700ms,
         // and when the app has transitioned to the running state...
-        if (this._appStateId == 0 && this._timeoutId == 0)
+        if (this._app.state == Shell.AppState.RUNNING && this._timeoutId == 0)
             this._clearSplash();
     },
 
@@ -275,17 +267,15 @@ const AppActivationContext = new Lang.Class({
         if (this._isBogusWindow(app))
             return;
 
-        appSystem.disconnect(this._appStateId);
-        this._appStateId = 0;
 
-        let aborted = this._abort;
-        this._abort = false;
-
-        if (aborted) {
+        if (this._abort) {
             this._app.request_quit();
             this._clearSplash();
+            this._abort = false;
         } else if (app.state == Shell.AppState.STOPPED) {
-            this._clearSplash();
+            // User cancelled the splash screen, but we need to wait
+            // for the running state before requesting to quit the app.
+            this._abort = true;
         } else {
             this._recordLaunchTime();
             this._maybeClearSplash();
@@ -303,8 +293,6 @@ const AppActivationContext = new Lang.Class({
         // The focused application changed and it is not the one that we are showing
         // the splash for, so clear the splash after it times out (because we don't
         // want to risk hiding too early)
-        this._appSystem.disconnect(this._appStateId);
-        this._appStateId = 0;
         this._maybeClearSplash();
     }
 });
@@ -317,9 +305,6 @@ const SpeedwagonIface = '<node> \
 <method name="HideSplash"> \
     <arg type="s" direction="in" name="desktopFile" /> \
 </method> \
-<signal name="SplashClosed"> \
-    <arg type="s" name="desktopFile" /> \
-</signal> \
 </interface> \
 </node>';
 const SpeedwagonProxy = Gio.DBusProxy.makeProxyWrapper(SpeedwagonIface);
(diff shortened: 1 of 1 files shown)


src/shell-app.c
```diff
@@ -1095,8 +1128,11 @@ _shell_app_add_window (ShellApp        *app,
 
   if (shell_app_is_interesting_window (window))
     app->running_state->interesting_windows++;
-  else if (shell_window_tracker_is_speedwagon_window (window))
+
+  if (shell_window_tracker_is_speedwagon_window (window)) {
```
cosimoc: Coding style: put curly brace on new line

src/shell-app.c
```diff
@@ -1095,8 +1128,11 @@ _shell_app_add_window (ShellApp        *app,
...
     app->running_state->speedwagon_windows++;
+    app->showing_splash = TRUE;
```
cosimoc: Why can't you check `speedwagon_windows > 0` instead of adding a new state?

src/shell-app.c
```diff
@@ -927,13 +928,45 @@ shell_app_on_user_time_changed (MetaWindow *window,
...
+}
+
 static void
 shell_app_sync_running_state (ShellApp *app)
 {
   g_return_if_fail (app->running_state != NULL);
 
   if (app->running_state->interesting_windows > 0)
     {
+      if (app->showing_splash && app->state == SHELL_APP_STATE_STOPPED) {
```
cosimoc: I'm not entirely sure why this code here doesn't already accomplish this: https://github.com/endlessm/gnome-shell/blob/master/js/ui/appActivation.js#L284

It seems to me that we're trying to solve a race condition between the `SplashClosed` signal and how this class uses the speedwagon window count to determine the window state. A better fix may be to simply remove the `SplashClosed` signal altogether then, and only rely on the window manager notifications to determine that, if the splash screen was indeed closed before the app became running, we should quit it.

mariospr: > I'm not entirely sure why this code here doesn't already accomplish this: https://github.com/endlessm/gnome-shell/blob/master/js/ui/appActivation.js#L284

That doesn't work right now since the app changes it state to STOPPED before the `SplashClosed` D-Bus signal sent by the speedwagon process reaches the shell, so the [code that is supposed to set `this._abort = true'](https://github.com/endlessm/gnome-shell/blob/master/js/ui/appActivation.js#L157) has not been called at this point, resulting in `this._app.request_quit()` never being executed. And even if got executed, `request_quit()` would be a no-op since no windows associated to the app are yet created at this point (just the speedwagon).

> It seems to me that we're trying to solve a race condition between the `SplashClosed` signal and how this class uses the speedwagon window count to determine the window state. A better fix may be to simply remove the `SplashClosed` signal altogether then, and only rely on the window manager notifications to determine that, if the splash screen was indeed closed before the app became running, we should quit it.

Yes, indeed there's a race condition and I also thought of removing that signal, but was not 100% sure about missing some use cases. I'll give it another thought then.

Also, it's entirely possible that I messed up during previous rebases when adapting the code to integrate with the speedwagon, so a more general audit/solution might be worth investigating as well.

src/shell-app.c
mariospr: Ok

src/shell-app.c
mariospr: Because by the time `shell_app_sync_running_state()` is called, from `_shell_app_remove_windo()`, `speedwagon_windows` will be `0` again.

That said, it's true this looks a bit awkward, I'll look for a better way to achieve a similar result, if I can find one.

js/ui/appActivation.js
```diff
@@ -275,17 +269,15 @@ const AppActivationContext = new Lang.Class({
...
-        if (aborted) {
+        if (this._abort) {
             this._app.request_quit();
             this._clearSplash();
+            this._abort = false;
         } else if (app.state == Shell.AppState.STOPPED) {
-            this._clearSplash();
+            // User cancelled the splash screen, but we need to wait
+            // for the running state before requesting to quit the app.
+            this._abort = true;
```
cosimoc: I don't completely follow the flow here... If we get `STOPPED` for a given app, how can we get another notification later for the `RUNNING` state?

If what you see is a transition like this `STARTING->(speedwagon close)->STOPPED->(app window is created)->RUNNING->(request quit)->STOPPED`, could we further clean that up and make it so that if `STOPPED` is reached from the `STARTING` state, we don't forward that out and handle the quit request directly `ShellAppSystem`?

From the outside, I think we should only see `STARTING->STOPPED` or `STARTING->RUNNING->STOPPED` state transitions sequences.

mariospr: > I don't completely follow the flow here... If we get STOPPED for a given app, how can we get another notification later for the RUNNING state?

As the code in shell-app.c is right now, you can get into this situation if you explicitly close the Speedwagon window, which 99% of the times will happen before any "interesting window" for the app itself is created, thus moving to STOPPED.

However, the app itself is still not killed, since `meta_window_delete()` could not be called yet (no windows => no-op), so the app will eventually get into the RUNNING state after that point, which is when we call `request_quit()`.

> If what you see is a transition like this STARTING->(speedwagon close)->STOPPED->(app window is created)->RUNNING->(request quit)->STOPPED, could we further clean that up and make it so that if STOPPED is reached from the STARTING state, we don't forward that out and handle the quit request directly ShellAppSystem?

That's the other possibility I mentioned in the ticket, which would involve killing (e.g. SIGTERM) the process being launched at the time the speedwagon is closed. That would involve 2 different code paths, though, once for apps launched via D-Bus (e.g. totem) and one for apps spawned/forked directly (e.g. Chromium), since in each case there are different ways to get the PID.

Since all this code was already quite messy I proposed to have a simpler solution for now which is based on what we were trying to attempt already (deferring quitting the app until it's RUNNING), as that is a single code path, so that this doesn't get much in the way of the rebase.

If you think it's important enough to do the right thing or do nothing, then I'd suggest waiting until the rebase is done and do it afterwards. Thoughts?

> From the outside, I think we should only see STARTING->STOPPED or STARTING->RUNNING->STOPPED state transitions sequences.

I agree with this as being the right thing to do for the long term.

cosimoc: >That's the other possibility I mentioned in the ticket, which would involve killing (e.g. SIGTERM) the process being launched at the time the speedwagon is closed. That would involve 2 different code paths, though, once for apps launched via D-Bus (e.g. totem) and one for apps spawned/forked directly (e.g. Chromium), since in each case there are different ways to get the PID.

>Since all this code was already quite messy I proposed to have a simpler solution for now which is based on what we were trying to attempt already (deferring quitting the app until it's RUNNING), as that is a single code path, so that this doesn't get much in the way of the rebase.

>If you think it's important enough to do the right thing or do nothing, then I'd suggest waiting until the rebase is done and do it afterwards. Thoughts?

I failed to properly convey my actual suggestion in my previous comment. I'm not suggesting that right now we actually kill the process without a window the moment the speedwagon is closed, even though that in theory would be the best case scenario.

What I'm suggesting is that we add logic directly in `shell-app.c` along the lines of "if we see the speedwagon window count dropping to zero, and the app would go to the STOPPED state, instead keep it in STARTING and then kill it as soon as the window appears". In other words, we'd "mask" the double transition from the outside world and completely handle internally in `shell-app.c`. Don't bother if it's too difficult though; in that case we can just merge this one.

mariospr: Well, that was what I attempted to do in the first version of this PR, maybe not in the best or most elegant way, but that was the intention: to do it self-contained in shell-app.c

That said, at that point I was still considering supporting the  -now gone- SplashClosed signal, so perhaps I can re-think it again.

PR #207, opened 2018-01-19

mariospr:
Stop relying on the SplashClosed D-Bus signal from speedwagon, which causes race conditions we'd need to deal with since such signal arrives normally after the application being launched moves to a STOPPED state, as per the speedwagon's window being explicitly closed.

Instead, rely on the new logic implemented now in shell-app.c, which will detect when the speedwagon window has been explicitly closed and wait until the app would reach the RUNNING state to kill it by using the meta_window_delete() function once the app's window is mapped.

### DIFF
--- a/src/shell-app.c
+++ b/src/shell-app.c
@@ -85,6 +85,7 @@ struct _ShellApp
                           */
 
   ShellAppRunningState *running_state;
+  gboolean showing_splash;
 
   char *window_id_string;
   char *name_collation_key;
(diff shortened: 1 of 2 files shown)


src/shell-app.c
```diff
@@ -927,13 +928,45 @@ shell_app_on_user_time_changed (MetaWindow *window,
...
+}
+
 static void
 shell_app_sync_running_state (ShellApp *app)
 {
   g_return_if_fail (app->running_state != NULL);
 
   if (app->running_state->interesting_windows > 0)
     {
+      if (app->showing_splash && app->state == SHELL_APP_STATE_STOPPED) {
```
cosimoc: Please put curly braces on their own line

src/shell-app.c
```diff
@@ -1095,8 +1128,11 @@ _shell_app_add_window (ShellApp        *app,
 
   if (shell_app_is_interesting_window (window))
     app->running_state->interesting_windows++;
-  else if (shell_window_tracker_is_speedwagon_window (window))
+
+  if (shell_window_tracker_is_speedwagon_window (window)) {
```
cosimoc: Here too

PR #208, opened 2018-01-24
remoteSearch: fix search order for apps on desktop

rshuler:
The search order properly put prioritized apps first, then apps
that are on the desktop, and finally apps that are installed
but not on the desktop.  The apps not on the desktop
were properly sorted alphabetically, but the apps that are on
the desktop were not sorted alphabetically with the old logic.

https://phabricator.endlessm.com/T15929


nedrichards:
Ideally this would be applied to the rebased 3.4 and 3.3 branches.

PR #210, opened 2018-02-15
T20178 stop background procs

treitter:
This branch automatically stops background processes to improve responsiveness. In my testing, this significantly helps on 1 GiB systems for `APP_FREEZING_CONCURRENT_PROCS_MAX <= 2` (with 1 being better than 2). It also helps on 2 GiB systems for `APP_FREEZING_CONCURRENT_PROCS_MAX` = 1 but actually worsens responsiveness when set to 2 in that case.

Because of the risk of regression for higher-memory systems and because these changes worsen the user experience if they aren't significantly improving responsiveness, I include a guard to only enable the feature on systems with <= 1 GiB of RAM.

This branch isn't 100% final since we still need to resolve the issue of `meta_window_get_child_pid()` from Mutter being private. I've contacted upstream and it seems likely we can get it made public but we should resolve that (at least by patching downstream and starting upstream review) before merging this.

I also need to rebase this upon master but I'll do so before proposing for final merge.


mariospr:
I've reviewed the PR, but I'm not entirely sure the approach followed to determine the most recently used apps is the right one. That said, I might be missing considerations you have probably made, so take this review with a grain of salt (I could be completely wrong after all)...

If I'm understand things right, in `update_app_freeze_states()` you build a hash table of `<pid, newest_timestamp>` for each process with "interesting" windows" (to discard things such as dialogs or splash screens), which you then use to create a balanced tree with the most recently used apps at the top, which you will then traverse applying `signal_stop_cont()` over its values, so that the first `APP_FREEZING_CONCURRENT_PROCS_MAX` processes get a `SIGCONT` and the remaining ones a `SIGSTOP`.

If this is correct, then the approach matches the description of the PR (and associated tickets) and I don't have any problem with it from a high level perspective (other than it looks more like a starting point to find the right heuristic, but I think that's the idea anyway), but I do find the implementation a bit too low-level and convoluted, and I wonder if there's a simpler approach that does not involve creating & destroying hash tables and tress each time the focus changes, as well as converting integers to pointers (and back) all the time, which seems to me like an approach that can be easily go out of control, as well as bug-prone (and kind of hard to follow reading the code). Also you rely on `meta_window_get_client_pid()` being public API, which is not at the moment, and I wonder if that's really needed if you use a different approach...

And as for that I had a (maybe naive) idea that would be as follows:

Instead of manually traversing the list of workspaces and windows in each of them, and then manually dealing with timestamps, interesting windows, pids and other things... why not simply using the APIs the shell currently exposes in ShellAppSystem and ShellApp?

In my mind, you could get the default `ShellAppSystem` instance and simply call `shell_app_system_get_running()`, which returns a `GSList*` of `ShellApp` instances, already ordered as I think you want them. According to the documentation:
```
 * Returns the set of applications which currently have at least one
 * open window in the given context.  The returned list will be sorted
 * by shell_app_compare().
```

And the documentation for `shell_app_compare()` says this:
```
 * Compare one #ShellApp instance to another, in the following way:
 *   - Running applications sort before not-running applications.
 *   - If one of them has non-minimized windows and the other does not,
 *     the one with visible windows is first.
 *   - Finally, the application which the user interacted with most recently
 *     compares earlier.
```

With that `GSList*` at hand, unless I'm missing something, all you'd have to do would be call `g_slist_foreach()` over its elements passing `signal_stop_cont()`, which would only need a `value` parameter now (the `key` parameter was being ignored in your PR anyway), send `SIGCONT` for the first `APP_FREEZING_CONCURRENT_PROCS_MAX` and `SIGSTOP` for the rest.

You would need to adapt `signal_stop_cont()` so that `value` is a `ShellApp`, and the you'd have to call `shell_app_get_pids()` to get the list of PIDs you need to "kill", but other than that it would be pretty much the same thing, I believe.

And if you ever need to get access to the MetaWindows from the apps and their timestamps, you could use `shell_app_get_windows()` and `meta_window_get_user_time()`, I guess, but perhaps that would not be needed. 

Anyway, as I said at the beginning, I might be missing things here so feel free to comment and push back against my comments. But before merging anything I'd like to understand better why this is implemented this way and whether a simpler approach (like the one proposed) is not possible, if that's the case.

treitter:
@mariospr Don't worry about feeling like you're pointing out the obvious -- I wasn't aware of `shell_app_system_get_running()` and `shell_app_compare()` and they do sound like they do what I want in a cleaner way.

I'll create a new version of my patch based on that.

This new approach should fulfill our immediate needs in a cleaner way. We may need a bit more flexibility for a more-nuanced heuristic in the future (which might factor in other details beyond recency) which is part of the reason I structured my patch the way I did but it's better to not be too anticipatory and only include pieces as they're needed.

Using both a hash and a tree wasn't my first choice and I agree it's clunky. I was originally just using a tree until I realized out `GTree` doesn't let you use a traverse function that factors in the value so I needed a parallel data structure. An alternative I briefly considered was just populating a `GList` and sorting it instead of using the `GTree` but I would have had to create a `struct`, `new()`/`free()` functions just to fit them into the `GList` so it seemed like it'd be similar enough. The joys of GLib/C :)

I've been trying to get feedback from Mutter upstream about making `meta_window_get_client_pid()` public but don't have a final answer yet.

treitter:
I think we need to abandon this approach since it breaks copy-and-paste.

See my comments on https://phabricator.endlessm.com/T20178

PR #211, opened 2018-03-06
trayArea: De-emphasize icons, matching the style of system indicators

mariospr:
Also, rename ICON_SATURATION into ICON_DESATURATION, since that's what
the values are really about (and otherwise it's confusing).

https://phabricator.endlessm.com/T21427

### DIFF
--- a/js/ui/components/trayArea.js
+++ b/js/ui/components/trayArea.js
@@ -47,10 +47,10 @@ const Me = ExtensionUtils.getCurrentExtension();
 
 // The original TopIcons Plus extension allowed configuring these values via a
 // GSettings schema, but we don't need that for now, let's hardcode those values.
-const ICON_BRIGHTNESS = 0.0;
-const ICON_CONTRAST = 0.0;
-const ICON_OPACITY = 240;
-const ICON_SATURATION = 0.4;
+const ICON_BRIGHTNESS = 0.0;   // range: [-1.0, 1.0]
+const ICON_CONTRAST = 0.0;     // range: [-1.0, 1.0]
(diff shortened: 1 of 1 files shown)


js/ui/components/trayArea.js
```diff
@@ -118,15 +118,15 @@ var TrayArea = new Lang.Class({
 
         // Saturation
         if (arguments.length == 1) {
-            let sat_effect = new Clutter.DesaturateEffect({ factor : ICON_SATURATION });
-            sat_effect.set_factor(ICON_SATURATION);
-            sat_effect.set_factor(ICON_SATURATION);
+            let sat_effect = new Clutter.DesaturateEffect({ factor : ICON_DESATURATION });
```
rshuler: For consistency, should probably call this variable "desat_effect"

js/ui/components/trayArea.js
```diff
@@ -118,15 +118,15 @@ var TrayArea = new Lang.Class({
...
+            sat_effect.set_factor(ICON_DESATURATION);
+            sat_effect.set_factor(ICON_DESATURATION);
```
rshuler: Why is this line duplicated?

js/ui/components/trayArea.js
mariospr: I agree

js/ui/components/trayArea.js
mariospr: Looks like an error we inherited from the original extension:
https://github.com/phocean/TopIcons-plus/blob/master/extension.js#L328

PR #217, opened 2018-03-16
payg: Initial implementation of the PAYG manager component

mariospr:
It simply proxies all the calls, signals and properties changes
between the shell and the eos-payg daemon.

https://phabricator.endlessm.com/T21607


mariospr:
This was not meant to be proposed for merge just yet, sorry

PR #218, opened 2018-03-16
payg: Initial implementation of the PAYG manager component

mariospr:
It simply proxies all the calls, signals and properties changes
between the shell and the eos-payg daemon.

https://phabricator.endlessm.com/T21607

### DIFF
--- a/js/ui/components/paygManager.js
+++ b/js/ui/components/paygManager.js
@@ -1,0 +1,154 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
+//
+// Copyright (C) 2018 Endless Mobile, Inc.
+//
+// This is a GNOME Shell component to wrap the interactions over
+// D-Bus with the eos-payg system daemon.
+//
+// Licensed under the GNU General Public License Version 2
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+const Gio = imports.gi.Gio;
+const GLib = imports.gi.GLib;
+
+const Main = imports.ui.main;
+const Lang = imports.lang;
+const Signals = imports.signals;
+
+const EOS_PAYG_NAME = 'com.endlessm.Payg1';
+const EOS_PAYG_PATH = '/com/endlessm/Payg1';
+
+const EOS_PAYG_IFACE = '<node> \
+<interface name="com.endlessm.Payg1"> \
+<method name="AddCode"> \
+  <arg type="ay" direction="in" name="code"/> \
+</method> \
+<method name="ClearCode" /> \
+<signal name="Expired" /> \
+<property name="ExpiryTime" type="t" access="read"/> \
+<property name="Enabled" type="b" access="read"/> \
+</interface> \
+</node>';
+
+var PaygManager = new Lang.Class({
+    Name: 'PaygManager',
+
+    _init: function() {
+        this._proxy = null;
+        this._proxyInfo = Gio.DBusInterfaceInfo.new_for_xml(EOS_PAYG_IFACE);
+
+        this._enabled = false;
+        this._expiryTime = 0;
+
+        this._codeExpiredId = 0;
+        this._propertiesChangedId = 0;
+    },
+
+    _onPropertiesChanged: function(proxy, changedProps, invalidatedProps) {
+        let propsDict = changedProps.deep_unpack();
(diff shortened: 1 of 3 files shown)


js/ui/components/paygManager.js
```diff
@@ -0,0 +1,154 @@
...
+
+    _onCodeExpired: function(proxy) {
+        this.emit('payg-code-expired');
+    },
+
+    _onProxyConstructed: function(object, res) {
+        try {
+            object.init_finish (res);
+        } catch (e) {
+            logError(e, "Error while constructing DBus proxy for " + EOS_PAYG_NAME);
```
pwithnall: Nitpick: s/DBus/D-Bus/

js/ui/components/paygManager.js
```diff
@@ -0,0 +1,154 @@
...
+    _setExpiryTime: function(value) {
+        if (this._expiryTime === value)
+            return;
+
+        this._expiryTime = value;
+        this.emit('payg-expiry-time-changed', this._expiryTime);
+    },
+
+    _onCodeExpired: function(proxy) {
+        this.emit('payg-code-expired');
```
pwithnall: Is ‘payg-’ needed in the name? This is the `PaygManager` class after all.

js/ui/components/paygManager.js
```diff
@@ -0,0 +1,154 @@
...
+        if (propsDict.hasOwnProperty('ExpiryTime'))
+            this._setExpiryTime(this._proxy.ExpiryTime);
+    },
+
+    _setEnabled: function(value) {
+        if (this._paygEnabled === value)
+            return;
+
+        this._paygEnabled = value;
+        this.emit('payg-enabled', this._paygEnabled);
```
pwithnall: Is ’payg-’ needed in the name? This is the `PaygManager` class after all. Also, should it be `enabled` or `enabled-changed`? The `expiry-time-changed` signal has a `-changed` suffix, which I think makes slightly more sense.

js/ui/components/paygManager.js
```diff
@@ -0,0 +1,154 @@
...
+    },
+
+    _setExpiryTime: function(value) {
+        if (this._expiryTime === value)
+            return;
+
+        this._expiryTime = value;
+        this.emit('payg-expiry-time-changed', this._expiryTime);
```
pwithnall: Not `notify::expiry-time`? I guess `expiry-time` isn’t actually a property on this class, is it?

js/ui/components/paygManager.js
```diff
@@ -0,0 +1,154 @@
...
+            this._propertiesChangedId = 0;
+        }
+
+        if (this._codeExpiredId > 0) {
+            this._proxy.disconnectSignal(this._codeExpiredId);
+            this._codeExpiredId = 0;
+        }
+
+        this._paygEnabled = false;
+        this._expiryTime = 0;
```
pwithnall: Should these property value changes cause signal emissions (of `enabled` and `expiry-time-changed`)?

js/ui/components/paygManager.js
mariospr: Correct

js/ui/components/paygManager.js
mariospr: I appreciate you taking revenge from my comment on the PaygEnabled signal, in the design document :-). I think it's a fair point, though, I'll remove the prefix.

js/ui/components/paygManager.js
mariospr: Yes, it actually did bother me a bit too. I think `enabled-changed` is probably the best option.

js/ui/components/paygManager.js
mariospr: Hmm... I guess it could be `notify::expiry-time`, yes. Probably good to make `notify::enabled` too?

js/ui/components/paygManager.js
mariospr: Hmmm... I suppose so, although it's a bit weird to emit that signal when you're disabling the component, no? In which case, maybe I should not re-set the values at all, and simply let `_onProxyConstructed()` to set them again when re-enabling the component (emitting the signal there if the value for any of those values changed compared to when it was previously enabled).

js/ui/components/paygManager.js
pwithnall: Depends whether those are actually exposed as GObject properties. If not, I think it would be incorrect and confusing to use the `notify` signal.

js/ui/components/paygManager.js
pwithnall: `disable()` is just another method call; it’s not like this is the `dispose` handler for the object and the object will no longer exist afterwards. Just in case someone calls `disable()` and then various UI/property bindings continue to use `paygEnabled` and `expiryTime`, I think it makes sense to reset the values (as you have done) and to emit the change signals (because all changes to the values should be notified). It makes sense to emit the change signals from `_onProxyConstructed()` too, for the same reasons.

js/ui/components/paygManager.js
mariospr: Well, that's a good point. In that case I think `enabled-changed` and `expiry-time-changed` makes more sense then, since PaygManager is not really a GObject, but a GJS class.

To be fair, I've also thought that would not be right if not emitting from a GObject, but the case of calendar.js made me hesitate since it emits a `notify::has-calendars` from `DBusEventSource` even if that's not a GObject either: https://github.com/endlessm/gnome-shell/blob/master/js/ui/calendar.js#L232

I'll change it then

js/ui/components/paygManager.js
mariospr: Fair point. I think that makes sense, I'll change it back.

PR #222, opened 2018-03-26
payg: Integrate the PAYG unlock dialog with the login screen

mariospr:
This integrates the PAYG unlock dialog in the login screen, making sure that such dialog is shown on boot & logout when the machine is locked.

https://phabricator.endlessm.com/T21609

### DIFF
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -525,6 +525,10 @@ var LoginDialog = new Lang.Class({
         // focus later
         this._startupCompleteId = Main.layoutManager.connect('startup-complete',
                                                              Lang.bind(this, this._updateDisableUserList));
+
+        // With PAYG machines this class can be created AFTER the startup process.
+        if (!Main.layoutManager.startingUp)
+            this._updateDisableUserList();
     },
 
     _getBannerAllocation: function (dialogBox) {
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -600,15 +600,16 @@ var ScreenShield = new Lang.Class({
                 return;
             }
 
-            if (Main.sessionMode.currentMode == 'unlock-dialog-payg') {
+            let paygExpectedMode = this._isGreeter ? 'gdm-unlock-dialog-payg' : 'unlock-dialog-payg';
+            if (Main.sessionMode.currentMode == paygExpectedMode) {
                 // This is the most common case.
-                Main.sessionMode.popMode('unlock-dialog-payg');
+                Main.sessionMode.popMode(paygExpectedMode);
             } else if (Main.sessionMode.currentMode == 'lock-screen') {
                 // There's a chance we could be in the screen shield, instead of the
                 // unlock dialog, if the user wend back manually to it (e.g. pressed ESC)
                 // between introducing the code and the actual verification happening.
                 Main.sessionMode.popMode('lock-screen');
-                Main.sessionMode.popMode('unlock-dialog-payg');
+                Main.sessionMode.popMode(paygExpectedMode);
                 Main.sessionMode.pushMode('lock-screen');
             }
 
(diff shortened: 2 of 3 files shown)


js/gdm/realmd.js
```diff
@@ -127,6 +127,9 @@ var Manager = new Lang.Class({
             }
         }
 
+        if (!this._loginFormat && !newLoginFormat)
+            return;
```
pwithnall: This early return will be taken if both `this._loginFormat` and `newLoginFormat` evaluate to `false`. In that case, we most likely have `this._loginFormat == newLoginFormat`, right? In which case, the branch below shouldn’t be taken and the function should exit without doing anything else anyway. Is that not the case?

mariospr:
> Review of commit “payg: Make sure to initialize the users list after startup has completed”: +1
>
> Would it make sense to upstream this? I realise it’s not useful upstream, but I don’t think it would be harmful or incorrect, right?

It could be, but my feeling is that upstream will reject it if nothing actually uses that. Maybe it's worth waiting until 3.30, since the unlock / login screens are going to be re-designed anyway.

mariospr:
> Review of commit “payg: Integrate the PAYG unlock screen with the login screen”: 100
>
> This is pleasingly small. As with the similar commit which introduced unlock-dialog-payg, I can’t be 100% sure that all the right state management is in all the right places. This is definitely one to test once it’s in an image.

Thanks. I still feel like it might be possible to get rid of the extra session added to handle the unlock dialog from the greeter (vs from a running session), but at the same time I kind of like how simple and "orthogonal" the patch is, so for now I think it's ok.

js/gdm/realmd.js
mariospr: Now I look at this patch, I realize is completely wrong, since the problem I'm trying to avoid is a reference to a non-initialized `this._loginFormat` attribute, which would throw an error. But that can be achieved by properly initializing that attribute in the constructor, so I think I will do that instead.

mariospr:
> Review of commit “payg: Fix the signature of the com.endlessm.Payg1.AddCode method”: fireworks
>
> (I’ve already mentioned this, but:) a similar change will need to be made to change ExpiryTime from usec to sec.

Isn't still a uint64 (`t` in GVariant slang)?

pwithnall:
> Isn't still a uint64 (`t` in GVariant slang)?

It is, but you’re going to have to interpret the number in seconds, rather than microseconds.

mariospr:
> > Isn't still a uint64 (t in GVariant slang)?
>
> It is, but you’re going to have to interpret the number in seconds, rather than microseconds.

I am already doing that, see https://github.com/endlessm/gnome-shell/blob/T21609/js/misc/paygManager.js#L166

js/gdm/realmd.js
```diff
@@ -68,6 +68,7 @@ var Manager = new Lang.Class({
                                            '/org/freedesktop/realmd',
                                            Lang.bind(this, this._reloadRealms))
         this._realms = {};
+        this._loginFormat = null;
```
pwithnall: There’s a check for `(this._loginFormat !== undefined)` on line 138. Will this change break that?

mariospr: I haven't observed any issue with this change but this is a fairly good point. I will double check the need of this patch then once again, since it is possible this was hit by an earlier version of my branch and maybe now it's not needed. But if it is, I agree I would probably need to be more careful since that can potentially break things.

mariospr: Actually, this patch **is not needed anymore** as I confirmed by reverting it and running my local tests once again. Removing it

pwithnall: Wheeee

pwithnall: The patch is still here in the branch on github, but presumably you’ve dropped it locally. This all looks good for you to merge.

PR #224, opened 2018-03-29
payg: Show an explicit error message when timing out

mariospr:
The 'Unknown error' message would cover this, but I think it's
interesting in this case to show something else for debugging
purposes, since an user pointing to us that a time out happened
will immediately direct us into a D-Bus communication issue.

https://phabricator.endlessm.com/T21806


pwithnall:
Ah, reading the [task](https://phabricator.endlessm.com/T21806) it seems this is defensive, rather than expected. Carry on. :-)

PR #226, opened 2018-03-29
Use the right text strings for the PAYG unlock dialog

mariospr:
https://phabricator.endlessm.com/T21841

### DIFF
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -172,12 +172,12 @@ var PaygUnlockDialog = new Lang.Class({
         this.actor.add_child(promptBox);
 
         let titleLabel = new St.Label({ style_class: 'unlock-dialog-payg-title',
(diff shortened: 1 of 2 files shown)


js/ui/paygUnlockDialog.js
```diff
@@ -193,6 +201,16 @@ var PaygUnlockDialog = new Lang.Class({
         this._buttonBox = this._createButtonsArea();
         promptBox.add_child(this._buttonBox);
 
+        let helpLineMain = new St.Label({ style_class: 'unlock-dialog-payg-help-main',
+                                          text: _("Dont have an unlock code? That's ok!"),
```
pwithnall: s/Dont/Don’t/

s/That's/That’s/ (Unicode apostrophe)

s/ok/OK/

mariospr: All addressed

PR #227, opened 2018-03-30
Gracefully handle com.endlessm.Payg1.Error.TooManyAttempts

mariospr:
https://phabricator.endlessm.com/T21808

### DIFF
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -130,6 +130,9 @@ var PaygUnlockCodeEntry = new Lang.Class({
     },
 
     addCharacter: function(character) {
+        if (!GLib.unichar_isdigit(character))
+            return;
+
         this.clutter_text.insert_unichar(character);
     },
 
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -74,10 +74,28 @@ var PaygUnlockCodeEntry = new Lang.Class({
         this.clutter_text.x_align = Clutter.ActorAlign.CENTER;
 
         this._enabled = false;
-        this.connect('button-press-event', this._onButtonPressEvent.bind(this));
+        this._buttonPressEventId = this.connect('button-press-event', this._onButtonPressEvent.bind(this));
+        this._capturedEventId = this.clutter_text.connect('captured-event', this._onCapturedEvent.bind(this));
+        this._textChangedId = this.clutter_text.connect('text-changed', this._onTextChanged.bind(this));
 
-        this.clutter_text.connect('captured-event', this._onCapturedEvent.bind(this));
-        this.clutter_text.connect('text-changed', this._onTextChanged.bind(this));
+        this.connect('destroy', this._onDestroy.bind(this));
+    },
+
+    _onDestroy: function() {
+        if (this._buttonPressEventId > 0) {
+            this.disconnect(this._buttonPressEventId);
+            this._buttonPressEventId = 0;
+        }
+
+        if (this._capturedEventId > 0) {
+            this.clutter_text.disconnect(this._capturedEventId);
+            this._capturedEventId = 0;
+        }
+
+        if (this._textChangedId > 0) {
+            this.clutter_text.disconnect(this._textChangedId);
+            this._textChangedId = 0;
+        }
     },
 
     _onCapturedEvent: function(textActor, event) {
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -82,17 +82,37 @@ var PaygUnlockCodeEntry = new Lang.Class({
         if (event.type() != Clutter.EventType.KEY_PRESS)
             return Clutter.EVENT_PROPAGATE;
 
-        let character = event.get_key_unicode();
+        let keysym = event.get_key_symbol();
+        let isDeleteKey =
+            keysym == Clutter.KEY_Delete ||
+            keysym == Clutter.KEY_KP_Delete ||
+            keysym == Clutter.KEY_BackSpace;
+        let isEnterKey =
+            keysym == Clutter.KEY_Return ||
+            keysym == Clutter.KEY_KP_Enter ||
+            keysym == Clutter.KEY_ISO_Enter;
+        let isExitKey =
+            keysym == Clutter.KEY_Escape ||
+            keysym == Clutter.KEY_Tab;
+        let isMovementKey =
+            keysym == Clutter.KEY_Left ||
+            keysym == Clutter.KEY_Right ||
+            keysym == Clutter.KEY_Home ||
+            keysym == Clutter.KEY_KP_Home ||
+            keysym == Clutter.KEY_End ||
+            keysym == Clutter.KEY_KP_End;
 
-        // We only support printable characters.
-        if (!GLib.unichar_isprint(character))
-            return Clutter.EVENT_PROPAGATE;
+        // Make sure we can leave the entry and delete and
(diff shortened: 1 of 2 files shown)


js/ui/paygUnlockDialog.js
```diff
@@ -81,17 +81,29 @@ var PaygUnlockCodeEntry = new Lang.Class({
...
+        let isDeleteKey = keysym == Clutter.KEY_Delete ||
+            keysym == Clutter.KEY_KP_Delete ||
+            keysym == Clutter.KEY_BackSpace;
+        let isEnterKey = keysym == Clutter.KEY_Return ||
+            keysym == Clutter.KEY_KP_Enter ||
+            keysym == Clutter.KEY_ISO_Enter;
+        let isExitKey = keysym == Clutter.KEY_Escape ||
+            keysym == Clutter.KEY_Tab;
+        let isArrowKey = keysym == Clutter.Left ||
+            keysym == Clutter.Right;
```
pwithnall: Up and Down normally go to the beginning and end of an entry. See also: Home, End, PageUp, PageDown.

js/ui/paygUnlockDialog.js
```diff
@@ -53,7 +53,8 @@ var UnlockStatus = {
     NOT_VERIFYING: 0,
     VERIFYING: 1,
     FAILED: 2,
-    SUCCEEDED: 3,
+    TOO_MANY_ERRORS: 3,
```
pwithnall: `TOO_MANY_ATTEMPTS`?

js/ui/paygUnlockDialog.js
```diff
@@ -277,13 +278,23 @@ var PaygUnlockDialog = new Lang.Class({
             this._startVerifyingCode();
         });
 
+        this._clearTooManyErrorsId = 0;
```
pwithnall: `_clearTooManyAttemptsId`?

js/ui/paygUnlockDialog.js
```diff
@@ -405,6 +415,26 @@ var PaygUnlockDialog = new Lang.Class({
...
             else
                 this._setErrorMessage(_("Too many attempts. Try again in a few seconds."));
+
+            // Make sure to clean the status once the time is up (if this dialog is still alive)
+            // and make sure that we install this callback at some point in the future (+1 sec).
+            this._clearTooManyErrorsId = Mainloop.timeout_add_seconds(secondsLeft + 1, () => {
+                this._verificationStatus = UnlockStatus.NOT_VERIFYING;
+                this._clearError();
+                this._updateSensitivity();
+                this._entry.grab_key_focus()
```
pwithnall: Missing trailing semicolon.

js/ui/paygUnlockDialog.js
```diff
@@ -342,7 +342,14 @@ var PaygUnlockDialog = new Lang.Class({
         } else if (error.matches(PaygManager.PaygErrorDomain, PaygManager.PaygError.CODE_ALREADY_USED)) {
             this._setErrorMessage(_("Code already used. Please enter a new code."));
         } else if (error.matches(PaygManager.PaygErrorDomain, PaygManager.PaygError.TOO_MANY_ATTEMPTS)) {
-            this._setErrorMessage(_("Too many attempts. Please wait and try again."));
+            let currentTime = GLib.get_real_time() / GLib.USEC_PER_SEC;
+            let secondsLeft = Main.paygManager.rateLimitEndTime - currentTime;
+            let minutesLeft = Math.max(0, Math.round(secondsLeft / 60))
+            if (minutesLeft > 0)
+                this._setErrorMessage(_("Too many attempts. Try again in %s %s.")
+                                      .format(minutesLeft, (minutesLeft > 1) ? _("minutes") : _("minute")));
```
pwithnall: Is there no `ngettext()` in JS? This method of doing plurals will not work for many languages.

What about if `Main.paygManager.rateLimitEndTime < currentTime`? `secondsLeft` will be negative then. I guess the `(minutesLeft > 0)` condition will fail then, but that seems a bit unsatisfying.

ptomato: It's available as `Gettext.ngettext()`

js/ui/paygUnlockDialog.js
mariospr: I was trying to keep it simple and FWIW, up & down and PgUp & PgDown do not always behave that way (e.g. path bar at the top of Nautilus) so I'd rather keep those ones out.

I'll add HOME and END, though

js/ui/paygUnlockDialog.js
mariospr: Can't believe this... ok!

js/ui/paygUnlockDialog.js
mariospr: Indeed

js/ui/paygUnlockDialog.js
mariospr: Right

js/ui/paygUnlockDialog.js
mariospr: > Is there no ngettext() in JS? This method of doing plurals will not work for many languages.

Oops! My bad... I didn't remember the way to do it and I put this as a placeholder, then I forgot. TL;DR: Yes, we can use ngettext(), I'll change that.

> What about if Main.paygManager.rateLimitEndTime < currentTime? secondsLeft will be negative then. I guess the (minutesLeft > 0) condition will fail then, but that seems a bit unsatisfying.

That should be handled by `Math.max(0, Math.round(secondsLeft / 60))` no?

mariospr: I'll change the condition to `minutesLeft >= 0` to make it more clear that the `else` branch handles the `[0, 1)` interval

js/ui/paygUnlockDialog.js
```diff
@@ -336,13 +411,41 @@ var PaygUnlockDialog = new Lang.Class({
...
 
-    _showError: function(error) {
+    _processError: function(error) {
+        logError(error, 'Error adding PAYG code');
+
+        // The 'too many errors' case is a bit special, and sets a different state.
+        if (error.matches(PaygManager.PaygErrorDomain, PaygManager.PaygError.TOO_MANY_ATTEMPTS)) {
+            let currentTime = GLib.get_real_time() / GLib.USEC_PER_SEC;
+            let secondsLeft = Main.paygManager.rateLimitEndTime - currentTime;
+            let minutesLeft = Math.max(0, Math.round(secondsLeft / 60))
```
pwithnall: I think this should be `Math.ceil()` rather than `Math.round()`, or the user might try again 30 seconds too early. Sorry I didn’t catch this before.

mariospr: I thought that, but if I do ceil I won't ever show the "Try in a few seconds message", since ' ceil(0.0000001) == 1`, so I thought being a little bit "imprecise" wouldn't be that bad.

pwithnall: You could be safely imprecise by always adding 1 minute to `minutesLeft` if `(secondsLeft > 0)` or something like that.

mariospr: I think I'll do this:
```
            let minutesLeft = Math.max(0, Math.ceil(secondsLeft / 60))
            if (minutesLeft >= 1 || secondsLeft > 30) {
                this._setErrorMessage(Gettext.ngettext("Too many attempts. Try again in %s minute.",
                                                       "Too many attempts. Try again in %s minutes.", minutesLeft)
                                      .format(minutesLeft));
            } else {
                this._setErrorMessage(_("Too many attempts. Try again in a few seconds."));
            }
```

PR #228, opened 2018-03-30
payg: Don't start the eos-paygd daemon if PAYG is not enabled

mariospr:
This will prevent the daemon for running at all, since the Shell
will no longer need it to check whether PAYG is enabled in a machine.

https://phabricator.endlessm.com/T21844

### DIFF
--- a/js/misc/paygManager.js
+++ b/js/misc/paygManager.js
@@ -60,18 +60,55 @@ const DBusErrorsMapping = {
     DISABLED          : 'com.endlessm.Payg1.Error.Disabled',
 }
 
+// This function checks the configuration file of PAYG directly
+// from the expected locations on disk, on an attempt to figure
+// out whether the feature is enabled, so that we don't wake up
+// the D-Bus service and keep it running when it's disabled.
+function _isPaygEnabled() {
+    // See man page eos-payg.conf(5)
+    let searchDirs = [
+        '/etc/eos-payg',
+        '/usr/local/share/eos-payg',
+        '/usr/share/eos-payg',
+    ];
+
+    let configFileName = 'eos-payg.conf'
+    let keyfile = new GLib.KeyFile();
+    try {
+        keyfile.load_from_dirs(configFileName,
+                               searchDirs,
+                               GLib.KeyFileFlags.NONE);
+        return keyfile.get_boolean('PAYG', 'Enabled');
+    } catch (e) {
+        // A non-existent file is a perfectly normal use case.
+        if (!e.matches(GLib.KeyFileError, GLib.KeyFileError.NOT_FOUND))
+            logError(e, "Error reading PAYG configuration file from " + configFileName);
+    }
+
+    return false;
+}
+
 var PaygManager = new Lang.Class({
     Name: 'PaygManager',
 
     _init: function() {
         this._initialized = false;
-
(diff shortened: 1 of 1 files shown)


js/misc/paygManager.js
```diff
@@ -60,33 +60,64 @@ const DBusErrorsMapping = {
...
         this._expiryTime = 0;
         this._rateLimitEndTime = 0;
 
-        this._codeExpiredId = 0;
-        this._propertiesChangedId = 0;
+        // Consider this manager initialized if PAYG is not
+        // enabled, and skip all the D-Bus related bits.
+        this._initialized = !_isPaygEnabled();
+        if (!this._initialized) {
+            this._proxy = null;
```
rshuler: I think you always want to explicitly initialize this._proxy to null, even if payg is not enabled.

js/misc/paygManager.js
```diff
@@ -60,33 +60,64 @@ const DBusErrorsMapping = {
...
-
         this._enabled = false;
         this._expiryTime = 0;
         this._rateLimitEndTime = 0;
 
-        this._codeExpiredId = 0;
-        this._propertiesChangedId = 0;
+        // Consider this manager initialized if PAYG is not
+        // enabled, and skip all the D-Bus related bits.
+        this._initialized = !_isPaygEnabled();
```
rshuler: Possibly confusing to say it is initialized if it is disabled, but I suppose I understand why you are doing this.  I did not track down all the places where you read this variable, but as long as they are basically checking whether initialization work is needed (as opposed to checking that the service is up and running), I can understand why you just want one variable.

Not a strong preference one way or the other (in fact, in my past life early returns were a big no-no), but it might be more consistent with how we tend to do things / easier to read using an early return:

```
this._initialized = false;
this._proxy = null;

if (!_isPaygEnabled()) {
    this._initialized = true;
    return;
}
```

js/misc/paygManager.js
mariospr: Not necessarily. The only two places where the proxy is needed (addCode() and clearCode()) are protected with `if (!this._proxy)`, which will be `false` if `this._proxy == undefined`.

js/misc/paygManager.js
mariospr: > Possibly confusing to say it is initialized if it is disabled, but I suppose I understand why you are doing this. I did not track down all the places where you read this variable, but as long as they are basically checking whether initialization work is needed (as opposed to checking that the service is up and running), I can understand why you just want one variable.

The only purpose of `this._initialized` is to check whether we're ready to call `screenShield.lock()` from main.js, during the initialization process. As the D-Bus proxy is asynchronously created, it might take some time, and so I have this flag to make sure we can query the state of PaygManager at any point.

Now, in the case where we don't really create the D-Bus proxy we can assume that PaygManager is immediately initialized, since there's no asynchronous operation to perform on construction, so that's where this comes from.

> Not a strong preference one way or the other (in fact, in my past life early returns were a big no-no), but it might be more consistent with how we tend to do things / easier to read using an early return:

I actually **love** early returns, but in this particular case I kind of liked how all the "optional" code that relates to the D-Bus proxy got indented inside the if, as that made it more clear to my eyes that is something optional. 

That said, I like your proposal too, so I'll change it.

PR #245, opened 2018-04-23
T22190

GeorgesStavracas:
When the user switches from a connection that had automatic
updates disabled to a connection that has it enabled (and
vice-versa too, enabled → disabled), the proposed behavior
is to send a notification about it.

This commit implements such behavior in a particularly
sophisticated way. We now store a new setting in each
network that we notify, so if you connect to a network
that you already connected in the past, before the new
automatic updates feature landed, you'll still get a
first-time notification. This is to make sure users
properly learn and understand about the feature.

In addition to that, a 4s timeout was added to the
notification. Showing the notification immediately
after connecting to a network is a drastic cognitive
load, and it's pretty invasive.


js/ui/status/automaticUpdates.js
```diff
@@ -151,6 +158,27 @@ var Indicator = new Lang.Class({
...
+            this._updateStatus();
+            return;
+        }
+
+        // Use a timeout to avoid instantly throwing the notification at
+        // the users' face, and to avoid a series of unecessary updates
+        // that happen when NetworkManager is still figuring out details.
+        this._updateTimeoutId = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT,
+                                                         4,
+                                                         this._updateStatus.bind(this));
```
mariospr: Instead of reusing `_updateStatus` here directly, could you instead use an arrow function that calls `_updateStatus() and then remove the source and set `_updateTimeoutId` to `0`? I don't like adding those extra lines at the end of `_updateStatus` that are only needed because of this use here in `timeout_add_seconds()`.

js/ui/status/automaticUpdates.js
```diff
@@ -171,6 +199,11 @@ var Indicator = new Lang.Class({
 
         // The status label
         this._item.label.text = _("Automatic Updates");
+
+        this._state = this._getState();
```
mariospr: This line looks like it could be kept here, though (instead of moving it into the arrow function requested above)

js/ui/status/automaticUpdates.js
```diff
@@ -171,6 +199,11 @@ var Indicator = new Lang.Class({
...
+
+        this._updateTimeoutId = 0;
+        return GLib.SOURCE_REMOVE;
```
mariospr: Move these too lines into the arrow function, this function shouldn't become a GSourceFunc just because it's now used with `timeout_add_seconds`

js/ui/status/automaticUpdates.js
```diff
@@ -187,9 +220,78 @@ var Indicator = new Lang.Class({
...
+        }
+
+        let gicon = new Gio.FileIcon({ file: Gio.File.new_for_uri(iconFile) });
+
+        // Create the notification
+        this._notification = new MessageTray.Notification(source, title, subtitle, { gicon: gicon });
+        this._notification.setUrgency(MessageTray.Urgency.NORMAL);
+        this._notification.setTransient(false);
+
+        this._notification.addAction(_("Change Settings…"), function() {
```
mariospr: Use an arrow function

js/ui/status/automaticUpdates.js
```diff
@@ -187,9 +220,78 @@ var Indicator = new Lang.Class({
...
+            let app = Shell.AppSystem.get_default().lookup_app('gnome-updates-panel.desktop');
+            Main.overview.hide();
+            app.activate();
+        });
+
+        source.notify(this._notification);
+
+        this._notification.connect('destroy', function() {
```
mariospr: Use an arrow function (otherwise `this` below will be unbound)

js/ui/status/automaticUpdates.js
```diff
@@ -286,12 +387,22 @@ var Indicator = new Lang.Class({
...
+
+        if (!connection)
+            return false;
 
-        if (!activeConnection)
-            activeConnection = this._client.get_activating_connection();
+        let userSetting = connection.get_setting(NM.SettingUser.$gtype);
+
+        if (!userSetting)
+            return true;
```
mariospr: Shouldn't this return `false`? If you can't get a valid `userSetting` here you can't really tell whether this is a new connection or not, right?

js/ui/status/automaticUpdates.js
```diff
@@ -187,9 +220,78 @@ var Indicator = new Lang.Class({
...
+
+        source.notify(this._notification);
+
+        this._notification.connect('destroy', function() {
+            this._notification = null;
+        });
+
+        // Now that we first detected this connection, mark it as such
+        let userSetting = this._ensureUserSetting(this._activeConnection);
+        userSetting.set_data(NM_SETTING_IS_NEW_CONNECTION, '1');
```
mariospr: As mentioned in Slack, I wonder if storing something more useful than just a boolean makes sense. Once added this `is-new-connection` data, we'll have to maintain it forever, so I wonder if perhaps storing something like a timestamp instead (and renaming to something like `last-notification-time`, for instance) could be equally helpful for this purpose, while perhaps being useful for something else in the future.

js/ui/status/automaticUpdates.js
GeorgesStavracas: The code assumes that, if there is no `NMSettingUser`, it is probably that the user connected to it in the past with Endless OS, __before the automatic updates feature__, and we didn't touch the connection. So, it must be notified.

js/ui/status/automaticUpdates.js
```diff
@@ -36,9 +36,9 @@ const MessageTray = imports.ui.messageTray;
 const PanelMenu = imports.ui.panelMenu;
 const PopupMenu = imports.ui.popupMenu;
 
+const EOS_SETTING_AUTOMATIC_UPDATES_NOTIFICATION_TIME = "eos.automatic-updates-notification-time";
```
mariospr: Why the sudden change to `eos` here? Does that mean that `connection.is-new-connection` was supported by NM upstream? (If that's the case that would be preferred, of course)

js/ui/status/automaticUpdates.js
```diff
@@ -174,7 +174,11 @@ var Indicator = new Lang.Class({
         // that happen when NetworkManager is still figuring out details.
         this._updateTimeoutId = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT,
                                                          4,
-                                                         this._updateStatus.bind(this));
+                                                         (data) => {
```
mariospr: You don't use `data` here

js/ui/status/automaticUpdates.js
```diff
@@ -225,15 +226,25 @@ var Indicator = new Lang.Class({
         this._updateNotification();
 
         // Connect from the current active connection
-        if (currentActiveConnection)
-            this._settingChangedSignalId = currentActiveConnection.connect('changed', this._updateStatus.bind(this));
+        if (currentActiveConnection) {
+            this._settingChangedSignalId = currentActiveConnection.connect('changed', () => {
```
mariospr: you can use `this._updateStatus.bind(this))` here, no need to change anything

js/ui/status/automaticUpdates.js
```diff
@@ -387,18 +399,18 @@ var Indicator = new Lang.Class({
             return AutomaticUpdatesState.IDLE;
     },
 
-    _isNewConnection: function() {
+    _alreadySentNotification: function() {
         let connection = this._getActiveConnection();
 
         if (!connection)
-            return false;
+            return true;
```
mariospr: I'm confused... why do you return `false` here now? Following the same reasoning from yesterday, if you can't get an active connection, you can't get the userSetting and thus you can't tell whether you already sent a notification, so you should return `false` (as you were returning before), no?

GeorgesStavracas: The semantics of the function changed. `true` now means "already notified; don't sent a notification". This, however, is just a defensive measure - if there is no active connection, this result is ignored.

js/ui/status/automaticUpdates.js
```diff
@@ -151,6 +158,31 @@ var Indicator = new Lang.Class({
...
+            this._updateStatus();
+            return;
+        }
+
+        // Use a timeout to avoid instantly throwing the notification at
+        // the users' face, and to avoid a series of unecessary updates
+        // that happen when NetworkManager is still figuring out details.
+        this._updateTimeoutId = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT,
+                                                         4,
+                                                         (data) => {
```
mariospr: You don't use data here

js/ui/status/automaticUpdates.js
```diff
@@ -187,9 +221,89 @@ var Indicator = new Lang.Class({
...
 
+        // Maybe send a notification if the connection changed and
+        // with it, the automatic updates setting
+        this._updateNotification();
+
         // Connect from the current active connection
-        if (currentActiveConnection)
-            this._settingChangedSignalId = currentActiveConnection.connect('changed', this._sync.bind(this));
+        if (currentActiveConnection) {
+            this._settingChangedSignalId = currentActiveConnection.connect('changed', () => {
```
mariospr: you can use this._updateStatus.bind(this)) here, no need to change anything

js/ui/status/automaticUpdates.js
```diff
@@ -286,12 +399,22 @@ var Indicator = new Lang.Class({
...
 
-    _getActiveConnection: function() {
-        let activeConnection = this._client.get_primary_connection();
+    _alreadySentNotification: function() {
+        let connection = this._getActiveConnection();
 
-        if (!activeConnection)
-            activeConnection = this._client.get_activating_connection();
+        if (!connection)
+            return true;
```
mariospr: I'm confused... why do you return false here now? Following the same reasoning from yesterday, if you can't get an active connection, you can't get the userSetting and thus you can't tell whether you already sent a notification, so you should return false (as you were returning before), no?

js/ui/status/automaticUpdates.js
```diff
@@ -32,9 +32,11 @@ const Shell = imports.gi.Shell;
 const St = imports.gi.St;
 
 const Main = imports.ui.main;
+const MessageTray = imports.ui.messageTray;
 const PanelMenu = imports.ui.panelMenu;
 const PopupMenu = imports.ui.popupMenu;
 
+const NM_SETTING_AUTOMATIC_UPDATES_NOTIFICATION_TIME = "connection.automatic-updates-notification-time";
```
mariospr: Can we use a shorter name? Actually, this makes me wonder whether we need a different namespace, or even an additional one. I'm thinking we might need either:
```
connection.automatic-updates.latest-notification-time
connection.automatic-updates.allow-downloads
connection.automatic-updates.tariff-enabled
```

Or simply:
```
automatic-updates.latest-notification-time
automatic-updates.allow-downloads
automatic-updates.tariff-enabled
```
... but I don't know whether there other components relying on those names. What do you think?

js/ui/status/automaticUpdates.js
mariospr: Still needs addressing

js/ui/status/automaticUpdates.js
mariospr: Or... if the `connection.` namespace actually implies "automatic updates", maybe we can simply use `connection.latest-notification-time`?

mariospr: Looks like this long name was agreed already with other people working in the metered data project. Ignore my comment then

js/ui/status/automaticUpdates.js
mariospr: Oh! I see you replied in https://github.com/endlessm/gnome-shell/pull/245#discussion_r184092691, but is not showing up here... anyway, I get what you said about the changing semantics but still I need whatever gets returned here needs to be consistent with what's returned in the early return below...

GeorgesStavracas: The semantics of the function changed. `true` now means "already notified; don't sent a notification". This, however, is just a defensive measure - if there is no active connection, this result is ignored.

js/ui/status/automaticUpdates.js
mariospr: This is still unaddressed

js/ui/status/automaticUpdates.js
mariospr: I know, I already replied to that in my previous comment (see https://github.com/endlessm/gnome-shell/pull/245/files#r184104687), and I still think this needs to be consistent with what whatever gets returned below, for `if (!userSetting)`

js/ui/status/automaticUpdates.js
mariospr: (to be clear, I mean using an arrow function here)

mariospr: (to be clear, I mean using an arrow function here)

js/ui/status/automaticUpdates.js
mariospr: As I said yesterday in the meeting, and mentioned to @GeorgesStavracas already a few times, I'm fine with whatever you agreed with. I simply didn't know where this came from, but you'll be the ones using it so :+1:

PR #254, opened 2018-05-02
T22190

mariospr:
This should make sure that the notification is not easily dismissed, by
requiring a explicit action from the user, and that it's shown expanded
so that the full text can be read right from the popup.

### DIFF
--- a/js/ui/status/automaticUpdates.js
+++ b/js/ui/status/automaticUpdates.js
@@ -173,7 +173,7 @@ var Indicator = new Lang.Class({
         // the user's face, and to avoid a series of unecessary updates
         // that happen when NetworkManager is still figuring out details.
         this._updateTimeoutId = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT,
-                                                         4,
+                                                         2,
                                                          () => {
                                                             this._updateStatus();
                                                             this._updateTimeoutId = 0;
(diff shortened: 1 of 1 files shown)


mariospr:
Ok. Closing for now to prevent "accidental" merges

js/ui/status/automaticUpdates.js
```diff
@@ -280,7 +280,7 @@ var Indicator = new Lang.Class({
 
         // Create the notification
         this._notification = new MessageTray.Notification(source, title, subtitle, { gicon: gicon });
-        this._notification.setUrgency(MessageTray.Urgency.NORMAL);
+        this._notification.setUrgency(alreadySentNotification ? MessageTray.Urgency.NORMAL : MessageTray.Urgency.CRITICAL);
```
nedrichards: I'd like to see a comment explaining our intent here.

rshuler: Oops -- your comment came in a minute after I merged.  I'll push a follow-up commit with a comment.

rshuler: Here's the updated comment:
```
        // Create the notification.
        // The first time we notify the user for a given connection,
        // we set the urgency to critical so that we make sure the
        // user understands how we may be changing their settings.
        // On subsequent notifications for the given connection,
        // for instance if the user regularly switches between
        // metered and unmetered connections, we set the urgency
        // to normal so as not to be too obtrusive.
```

PR #255, opened 2018-05-03
appIconBar: use the right method to specify packing properties

cosimoc:
ClutterContainer.add() is patched in js/ui/environment.js to allow passing
child properties. We use ClutterActor.add_child() though, which doesn't
allow that, and causes warnings on the terminal when that code is
reached.

https://phabricator.endlessm.com/T21741

### DIFF
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -77,7 +77,7 @@ const WindowMenuItem = new Lang.Class({
 
         this.cloneBin = new St.Bin({ child: clone,
                                      style_class: 'panel-window-menu-item-clone' });
-        this.actor.add_child(this.cloneBin, { align: St.Align.MIDDLE });
+        this.actor.add(this.cloneBin, { align: St.Align.MIDDLE });
 
         this.label = new St.Label({ text: window.title,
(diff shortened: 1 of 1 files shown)


js/ui/appIconBar.js
```diff
@@ -77,7 +77,7 @@ const WindowMenuItem = new Lang.Class({
 
         this.cloneBin = new St.Bin({ child: clone,
                                      style_class: 'panel-window-menu-item-clone' });
-        this.actor.add_child(this.cloneBin, { align: St.Align.MIDDLE });
+        this.actor.add(this.cloneBin, { align: St.Align.MIDDLE });
```
mariospr: I'd rather keep using `clutter_actor_add_child()` instead of the deprecated `clutter_container_add()` function (even if it's overriden, which is again a pretty weird thing to do IMHO), and set the alignment for the `cloneBin` actor using `x-align` and `y-align` properties during its creation.

That's actually how we've been fixing these kind of issues so far, e.g. https://github.com/endlessm/gnome-shell/commit/8582b2a14f22a31e45de5442b5731fd533af1243

mariospr: Merging as per the standup

PR #256, opened 2018-05-08
T22190

GeorgesStavracas:
This commit fixes the notifications, and adds a "Close" button to the notification.

Even though a translatable string was added, there are translations available for it -- no new translations were introduced here.

### DIFF
--- a/js/ui/status/automaticUpdates.js
+++ b/js/ui/status/automaticUpdates.js
@@ -290,6 +290,10 @@ var Indicator = new Lang.Class({
         this._notification.setUrgency(alreadySentNotification ? MessageTray.Urgency.NORMAL : MessageTray.Urgency.CRITICAL);
         this._notification.setTransient(false);
 
+        this._notification.addAction(_("Close"), () => {
+            this._notification.destroy();
+        });
+
         this._notification.addAction(_("Change Settings…"), () => {
             let app = Shell.AppSystem.get_default().lookup_app('gnome-updates-panel.desktop');
             Main.overview.hide();
--- a/js/ui/status/automaticUpdates.js
+++ b/js/ui/status/automaticUpdates.js
@@ -128,6 +128,7 @@ var Indicator = new Lang.Class({
         this._client = NM.Client.new_finish(result);
 
         this._client.connect('notify::primary-connection', this._sync.bind(this));
+        this._client.connect('notify::state', this._sync.bind(this));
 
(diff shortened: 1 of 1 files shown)


js/ui/status/automaticUpdates.js
```diff
@@ -163,11 +164,10 @@ var Indicator = new Lang.Class({
             this._updateTimeoutId = 0;
         }
 
-        // When disconnected, don't timeout before updating
-        if (!this._getActiveConnection()) {
-            this._updateStatus();
+        // Intermediate states (connecting or disconnecting) must not trigger
+        // any kind of state change.
+        if (this._client.state == NM.State.CONNECTING || this._client.state == NM.State.DISCONNECTING)
```
mariospr: Just a thought here: What about doing it the other way around and only proceeding if the state is either `NM.State.CONNECTED_GLOBAL` (not only connected to the network, but also to the Internet), NM.State.DISCONNECTED` or `NM.State.UNKNOWN` (I presume that would behave as `DISCONNECTED`)?

GeorgesStavracas: This is a conscious code style practice of mine. I sistematicaly prefer early returns rather than chained conditions.

mariospr: My comment is not about "early returns vs chained blocks" (I also prefer early returns), but about explicitly considering the subset of all possible states we're interested in, since I have the feeling we might want to ignore other states such as CONNECTED_LOCAL, or CONNECTED_SITE.

In other words, what I'm proposing is not the exact opposite (or negation) of the condition you have in there to early return, but something slightly different.

See https://developer.gnome.org/libnm/stable/libnm-nm-dbus-interface.html#NMState

GeorgesStavracas: Got it. In this case, then, I still wouldn't change this, because the rest of the code should be able to deal with any other possible `NMState` value.

As for `CONNECTED_GLOBAL` x `CONNECTED_SITE` x `CONNECTED`, it shouldn't change how the Automatic Updates indicator behave. This difference is not relevant at this level (but it might be for lower levels, such as Mogwai or GNOME Software).

`UNKNOWN` and `DISCONNECTED` are handled equally in the code.

mariospr: Ok. Sounds reasonable to me, thanks. Merging...

PR #257, opened 2018-05-08
fixup! networkAgent: Add switch button for metered connections

GeorgesStavracas:
Per new design direction, we want the dialog labels to be centered
and use the High-Contrast toggle asset instead of the on|off switcher.

https://phabricator.endlessm.com/T22420

### DIFF
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -923,3 +923,9 @@ popup-separator-menu-item {
   }
 }
 
+.metered-data-switch {
+  width: 48px;
+  background-image: url("resource:///org/gnome/shell/theme/toggle-off-hc.svg");
+  &:checked { background-image: url("resource:///org/gnome/shell/theme/toggle-on-hc.svg"); }
+}
+
--- a/js/ui/components/networkAgent.js
+++ b/js/ui/components/networkAgent.js
@@ -50,7 +50,8 @@ var NetworkSecretDialog = new Lang.Class({
         this._inputSourceManager.passwordModeEnabled = true;
 
         let contentParams = { title: this._content.title,
-                              body: this._content.message };
+                              body: this._content.message,
+                              center_title: true };
         let contentBox = new Dialog.MessageDialogContent(contentParams);
         this.contentLayout.add_actor(contentBox);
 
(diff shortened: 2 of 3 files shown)


data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -923,3 +923,8 @@ popup-separator-menu-item {
   }
 }
 
+.metered-data-switch { width: 48px; }
+.metered-data-switch {
+  background-image: url("resource:///org/gnome/shell/theme/toggle-off-hc.svg");
+  &:checked { background-image: url("resource:///org/gnome/shell/theme/toggle-on-hc.svg"); }
+}
```
mariospr: Nit: Missing extra line, sorry!

PS: @pwithnall will approve

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -923,3 +923,8 @@ popup-separator-menu-item {
   }
 }
 
+.metered-data-switch { width: 48px; }
+.metered-data-switch {
```
mariospr: This is SASS, an so we should take advantage of its capabilities to define things hierarchical: please only declare one `.metered-data-switch` block in this file (i.e. move the `width` statement in the first block to the second one)

js/ui/components/networkAgent.js
```diff
@@ -219,7 +226,7 @@ var NetworkSecretDialog = new Lang.Class({
         if (userSetting) {
             let allowDownloads = userSetting.get_data(NM_SETTING_ALLOW_DOWNLOADS);
             if (allowDownloads)
-                defaultValue = (allowDownloads === '0');
+                defaultValue = (allowDownloads === '1');
```
mariospr: This PR also changes what's supposed to be the default from "not allowing downloads" to "allowing downloads", which I don't see documented neither in the commit message nor in the originating ticket. Where is this coming from?

If it's not  a mistake, please add a brief comment to the commit message.

js/ui/components/networkAgent.js
```diff
@@ -229,9 +236,9 @@ var NetworkSecretDialog = new Lang.Class({
 
             if (connectionSetting) {
                 let metered = connectionSetting.get_metered();
-                defaultValue = metered == NM.Metered.YES || metered == NM.Metered.GUESS_YES;
+                defaultValue = metered != NM.Metered.YES && metered != NM.Metered.GUESS_YES;
             } else {
-                defaultValue = false;
+                defaultValue = true;
```
mariospr: Same comment here (about changing defaults)

js/ui/components/networkAgent.js
```diff
@@ -265,7 +272,7 @@ var NetworkSecretDialog = new Lang.Class({
 
             // Update the Automatic Updates values
             if (this._settingName != 'vpn')
-                this._updateMeteredSetting(this._updatesToggle.checked);
+                this._updateMeteredSetting(!this._updatesToggle.checked);
```
mariospr: Similar comment here about flipping a boolean in this PR without explaining why

PR #259, opened 2018-05-10
discoveryFeed: Tell the Discovery Feed when its window is fully hidden

smspillaz:
The discovery feed window animates in and out and the overview is
only shown once the discovery feed window is fully hidden (if the
overview was showing before). The Discovery Feed needs to know
when its window is fully hidden, since then it can launch the
relevant app (which will ensure that the overview is not shown
part-way through a speedwagon window being shown).

https://phabricator.endlessm.com/T22448

### DIFF
--- a/js/ui/components/discoveryFeed.js
+++ b/js/ui/components/discoveryFeed.js
@@ -10,6 +10,7 @@ const DISCOVERY_FEED_PATH = '/com/endlessm/DiscoveryFeed';
 
 const DiscoveryFeedIface = '<node> \
 <interface name="' + DISCOVERY_FEED_NAME + '"> \
+<method name="notifyHideAnimationCompleted" /> \
 <method name="show"> \
   <arg type="u" direction="in" name="timestamp"/> \
 </method> \
@@ -38,12 +39,16 @@ var DiscoveryFeed = new Lang.Class({
         Main.discoveryFeed = null;
     },
 
+    notifyHideAnimationCompleted: function() {
+        this.proxy.notifyHideAnimationCompletedRemote();
+    },
+
     callShow: function(timestamp) {
         this.proxy.showRemote(timestamp);
     },
 
-    callHide: function(timestamp) {
-        this.proxy.hideRemote(timestamp);
+    callHide: function() {
+        this.proxy.hideRemote();
     }
 });
 var Component = DiscoveryFeed;
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -2232,6 +2232,11 @@ var WindowManager = new Lang.Class({
                     this._showOtherWindows(actor, false);
             }
 
+            /* If this is the Discovery Feed, notify it that it has
+             * finished closing now */
+            if (SideComponent.isDiscoveryFeedWindow(actor.meta_window))
+                Main.discoveryFeed.notifyHideAnimationCompleted();
+
             shellwm.completed_destroy(actor);
         }
     },


js/ui/windowManager.js
```diff
@@ -2232,6 +2232,12 @@ var WindowManager = new Lang.Class({
                     this._showOtherWindows(actor, false);
             }
 
+            /* If this is the Discovery Feed, notify it that it has
+             * finished closing now */
+            if (SideComponent.isDiscoveryFeedWindow(actor.meta_window)) {
```
mariospr: No need to use brackets here

js/ui/components/discoveryFeed.js
```diff
@@ -16,6 +16,7 @@ const DiscoveryFeedIface = '<node> \
 <method name="hide"> \
   <arg type="u" direction="in" name="timestamp"/> \
 </method> \
+<method name="hidden" /> \
```
mariospr: This method might dissapear from here or be renamed at least, depending on the way the comment in https://github.com/endlessm/eos-discovery-feed/pull/108#discussion_r187405937 gets addressed.

cosimoc:
I just realized that the `launchedFromDesktop` property available to side components, and in `js/ui/windowManager.js` IIRC was trying to accomplish just this.
I believe that the issue is it's broken for the feed at the moment, and I will look to see if it's possible to fix it there instead.

Otherwise this approach is OK, but I second Mario's suggestion about using a different naming.

smspillaz:
> I just realized that the launchedFromDesktop property available to side components, and in js/ui/windowManager.js IIRC was trying to accomplish just this.
I believe that the issue is it's broken for the feed at the moment, and I will look to see if it's possible to fix it there instead.

I'll push a revised PR here with the changes suggested by Mario and then give this one a try.

smspillaz:
Hmmm, @cosimoc it seems to me that the `launchedFromDesktop` logic was already there for the discovery feed:

```
/**
 * launchedFromDesktop:
 * @metaWindow: an instance of #Meta.Window
 * @return: whether the side component was launched from the desktop
 */
function launchedFromDesktop (metaWindow) {
    return isSideComponentWindow(metaWindow) &&
        ((metaWindow.get_wm_class == 'Eos-app-store' && Main.appStore.launchedFromDesktop) ||
         (isDiscoveryFeedWindow(metaWindow) && Main.discoveryFeed.launchedFromDesktop));
};
```

I think that before we did all this "hide immediately once an entry is clicked" business that logic would have been used, since the speedwagon window would have appeared and taken focus away from the feed (and thus, we would not have returned back to the overview). However, now that we immediately close the discovery feed window as soon as an entry is clicked and there isn't necessarily some other window taking focus, we'll attempt to return to the overview ... only for the speedwagon for that other window to show up a little bit later (then when the discovery feed window is *finished* closing in `_destroyWindowDone`, we'll end up showing the desktop there, even though the speedwagon appeared in the interim:

```
            if (SideComponent.isSideComponentWindow(actor.meta_window) && this._showDesktopOnDestroyDone) {
                Main.overview.showApps();

                if (SideComponent.shouldHideOtherWindows(actor.meta_window))
                    this._showOtherWindows(actor, false);
            }
```

So I don't think we can use the `launchedFromDesktop` method.

Side note: `WindowManager` is a singleton and setting a property on it (`_showDesktopOnDestroyDone`) when a side component starts destroying and expecting it to be consistent when that same side component finishes destroying seems awfully fragile to me....

js/ui/components/discoveryFeed.js
```diff
@@ -38,12 +39,16 @@ var DiscoveryFeed = new Lang.Class({
...
+    },
+
     callShow: function(timestamp) {
         this.proxy.showRemote(timestamp);
     },
 
-    callHide: function(timestamp) {
-        this.proxy.hideRemote(timestamp);
+    callHide: function() {
+        this.proxy.hideRemote();
```
cosimoc: Why did you remove this timestamp?

smspillaz: Must have bee a mistake when I removed the timestamp from the old `hidden` method. I'll create a new PR to revert this change - thanks for spotting

PR #260, opened 2018-05-11
Finish startup sequence for process that exit too early

mariospr:
Add a new public method in ShellWindowTracker to wrap sn_startup_sequence_complete() and use that to make sure that these kind of applications get their startup sequences completed, and then properly moved to STOPPED.

https://phabricator.endlessm.com/T22416

### DIFF
--- a/src/shell-app.c
+++ b/src/shell-app.c
@@ -1254,11 +1254,36 @@ app_child_setup (gpointer user_data)
 #endif
 
 static void
+_shell_app_watch_callback (GPid      pid,
+                           gint      status,
+                           ShellApp *app)
+{
+  if (app->state == SHELL_APP_STATE_STARTING)
+    {
+      ShellWindowTracker *tracker = shell_window_tracker_get_default ();
+      GSList *startup_sequences = shell_window_tracker_get_startup_sequences (tracker);
+      GSList *iter = NULL;
+
+      for (iter = startup_sequences; iter; iter = g_slist_next (iter))
+        {
+          ShellStartupSequence *sequence = (ShellStartupSequence*) iter->data;
+          ShellApp *startup_app = shell_startup_sequence_get_app (sequence);
+          if (startup_app == app)
+            shell_startup_sequence_complete (sequence);
+        }
+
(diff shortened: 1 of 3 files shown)


src/shell-window-tracker.c
```diff
@@ -1008,6 +1008,11 @@ shell_startup_sequence_create_icon (ShellStartupSequence *sequence, guint size)
   return texture;
```
smspillaz: Typo in the commit message s/publich/public/

src/shell-app.c
```diff
@@ -1253,12 +1253,37 @@ app_child_setup (gpointer user_data)
...
+static void
+_shell_app_watch_callback (GPid      pid,
+                           gint      status,
+                           ShellApp *app)
+{
+  if (app->state == SHELL_APP_STATE_STARTING)
+    {
+      ShellWindowTracker *tracker = shell_window_tracker_get_default ();
+      GSList *startup_sequences = shell_window_tracker_get_startup_sequences (tracker);
+      GSList *iter = NULL;
```
smspillaz: `iter` can probably assigned with `startup_sequences` directly here

src/shell-app.c
```diff
@@ -1305,7 +1330,8 @@ shell_app_launch (ShellApp     *app,
 #else
                                                    NULL, NULL,
 #endif
-                                                   wait_pid, NULL,
+                                                   (GDesktopAppLaunchCallback) wait_pid,
```
smspillaz: Is this change relevant to this commit?

smspillaz:
Looks good - just left a few drive-by nitpicks

mariospr:
> Looks good - just left a few drive-by nitpicks

Thanks. I've addresed the typo in the commit message but I don't think there' nothing that needs to be done with regard the other two comments, see my answers inline. Please let me know what you think.

mariospr:
As mentioned in the ticket, we will be fixing this in GTK after all, so closing this MR now.

Let's continue in https://gitlab.gnome.org/GNOME/gtk/issues/1084

PR #265, opened 2018-06-15
endSessionDialog: Don't ever speak to PackageKit (not available on Endless)

mariospr:
GNOME Shell is spitting out some errors in the journal due to its attempts
to speak to PackageKit, which is not present on Endless OS, so remove all
the related code to silence those errors.

https://phabricator.endlessm.com/T22975


js/ui/endSessionDialog.js
```diff
@@ -286,14 +286,34 @@ var EndSessionDialog = new Lang.Class({
...
+                                                          return;
+                                                      }
+
+                                                      // It only makes sense to check for this permission if PackageKit is available.
+                                                      try {
+                                                          this._updatesPermission =
+                                                              Polkit.Permission.new_sync("org.freedesktop.packagekit.trigger-offline-update", null, null);
+                                                      } catch(e) {
+                                                          log('No permission to trigger offline updates: %s'.format(e.toString()));
+                                                      }
```
GeorgesStavracas: I personally would rather move this into a method of its own. It's too much code in a far indentation.

mariospr: I kind of agree, but at the same time I like having the code that initializes `this._updatesPermission` all in th e`_init()` function, so I'm not sure it's worth it.

Maybe comment the same thing in the upstream MR, and see what others (e.g. Florian) say about it?

PR #267, opened 2018-07-02
wobbly: Only query the untransformed paint box when resizing model

smspillaz:
We want the model size to be based on the size of the window
including its shadows, but we do not want that to include any
affine transformations currently done on the window, which
clutter_actor_get_paint_box will perform on the volume before
projecting it and getting its bounding box.

This commit assumes that the paint volume without any effects applied
is two dimensional and orthogonal.

https://phabricator.endlessm.com/T23034

### DIFF
--- a/src/wobbly-effect.cpp
+++ b/src/wobbly-effect.cpp
@@ -327,6 +327,36 @@ endless_shell_fx_wobbly_move_by (EndlessShellFXWobbly *effect,
     }
 }
 
+/* This constant is used to deal with rounding error in computing
+ * paint boxes. See also https://gitlab.gnome.org/GNOME/mutter/blob/master/clutter/clutter/clutter-paint-volume.c#L1212 */
+#define PAINT_BOX_OFFSET 1
+
+static gboolean
+endless_shell_fx_get_untransformed_paint_box (ClutterActor    *actor,
+                                              ClutterActorBox *box)
+{
+  /* Get the actor's paint volume, bypassing affine
+   * transformations which would usually be applied
+   * if we just queried the paint box. */
+  const ClutterPaintVolume *volume = clutter_actor_get_paint_volume (actor);
+  ClutterVertex origin;
+
+  if (volume == NULL)
+    return FALSE;
+
+  /* We don't have access to the stage projection matrix
+   * so the best we can do is hope here that the volume is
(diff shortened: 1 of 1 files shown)


src/wobbly-effect.cpp
```diff
@@ -327,6 +327,33 @@ endless_shell_fx_wobbly_move_by (EndlessShellFXWobbly *effect,
     }
 }
 
+#define PAINT_BOX_OFFSET 1
+
+static gboolean
+endless_shell_fx_get_untransformed_paint_box (ClutterActor    *actor,
+                                              ClutterActorBox *box)
+{
+  /* Get the actor's parent class paint volume, bypassing affine
```
cosimoc: This comment confuses me... There does not seem to be any involvement of the parent class.

src/wobbly-effect.cpp
```diff
@@ -327,6 +327,33 @@ endless_shell_fx_wobbly_move_by (EndlessShellFXWobbly *effect,
...
+   * transformations (except translation). */
+  const ClutterPaintVolume *volume = clutter_actor_get_paint_volume (actor);
+  ClutterVertex origin;
+
+  if (volume == NULL)
+    return FALSE;
+
+  /* We don't have access to the stage projection matrix
+   * so the best we can do is hope here that the volume is
+   * two dimensional and orthogonal. */
```
cosimoc: I believe that we do... see `clutter_actor_get_transform()` and `clutter_paint_volume_get_depth()`

src/wobbly-effect.cpp
```diff
@@ -327,6 +327,33 @@ endless_shell_fx_wobbly_move_by (EndlessShellFXWobbly *effect,
     }
 }
 
+#define PAINT_BOX_OFFSET 1
```
cosimoc: It surprises me to see magic numbers here. What is this for?

src/wobbly-effect.cpp
smspillaz: Hmm, I guess I meant that in the sense that the actor needs to be a camera-facing plane for this to work. If it was three-dimensional this probably wouldn't work - in that case we would use the modelview matrix and premultiply it by the projection matrix, then postmultiply each of its vertices, but we don't have access to the projection matrix here.

PR #268, opened 2018-07-02
wobbly: Use the paint box offsets to compute the right paint volume

smspillaz:
Since the model is computed from the actor only paint box which
will be in the [0, model_geometry] space, this doesn't include
the top and left shadow offsets, so we need to apply them after
the fact. This corrects the computed paint volume so that there are
no shadow artifacts.

https://phabricator.endlessm.com/T23039

### DIFF
--- a/src/wobbly-effect.c
+++ b/src/wobbly-effect.c
@@ -64,6 +64,12 @@ G_DEFINE_TYPE_WITH_PRIVATE (EndlessShellFXWobbly,
                             endless_shell_fx_wobbly,
                             CLUTTER_TYPE_DEFORM_EFFECT)
 
+/* Forward declaration so that we can set the function pointer
+ * in enforce_no_effects_paint_box_cleanup below */
+static gboolean
+endless_shell_fx_wobbly_get_paint_volume (ClutterEffect    *effect,
+                                          ClutterPaintVolume *volume);
+
 static gboolean
 endless_shell_fx_wobbly_lie_about_paint_volume (ClutterEffect    *effect,
                                                 ClutterPaintVolume *volume)
--- a/src/wobbly-effect.c
+++ b/src/wobbly-effect.c
@@ -167,7 +167,8 @@ static gboolean
 endless_shell_fx_wobbly_get_paint_volume (ClutterEffect    *effect,
                                           ClutterPaintVolume *volume)
 {
-  ClutterActor *actor = clutter_actor_meta_get_actor (CLUTTER_ACTOR_META (effect));
+  ClutterActorMeta *meta = CLUTTER_ACTOR_META (effect);
+  ClutterActor *actor = clutter_actor_meta_get_actor (meta);
   EndlessShellFXWobbly *wobbly_effect = ENDLESS_SHELL_FX_WOBBLY (effect);
   EndlessShellFXWobblyPrivate *priv =
     endless_shell_fx_wobbly_get_instance_private (wobbly_effect);
@@ -176,7 +177,7 @@ endless_shell_fx_wobbly_get_paint_volume (ClutterEffect    *effect,
    * TRUE here. */
   CLUTTER_EFFECT_CLASS (endless_shell_fx_wobbly_parent_class)->get_paint_volume (effect, volume);
 
-  if (priv->model)
+  if (priv->model && clutter_actor_meta_get_enabled (meta))
     {
       float actor_x, actor_y;
       float actor_paint_box_x, actor_paint_box_y;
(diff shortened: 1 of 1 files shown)


src/wobbly-effect.cpp
```diff
@@ -77,6 +77,8 @@ endless_shell_fx_wobbly_lie_about_paint_volume (ClutterEffect    *effect,
                                                                                         volume);
 }
 
+#define DAMAGE_RECTANGLE_EXPAND_FACTOR 8
```
cosimoc: Ouch, this feels a bit like stabbing in the dark. Are you sure that we're not papering over some other issue with this, such as using the wrong actor to calculate the paint volume, or a bug in the wobbly model calculations?
I would at least like to know where those 8px come from, and have a clear understanding, encoded in a comment, of why we can't trust the numbers we get from Clutter/libwobbly.

js/ui/windowManager.js
```diff
@@ -158,6 +159,17 @@ var EOSShellWobbly = new Lang.Class({
...
     },
 
+    _removeFromActorOnDisabled: function(actor) {
+        if (!this.enabled) {
+            if (this._wobblyEffectDisabledId !== -1) {
+                this.disconnect(this._wobblyEffectDisabledId);
+                this._wobblyEffectDisabledId = -1;
+            }
+
+            actor.remove_effect(this);
```
cosimoc: This strikes me as an odd way of achieving the goal of the commit.
What ends up actually disabling the effect? The effect is created and added in `windowManager.js` before calling into `grabbedByMouse()`. If the idea is to remove it when it's ungrabbed, why not just do it from there as well?

js/ui/windowManager.js
```diff
@@ -189,6 +206,20 @@ var EOSShellWobbly = new Lang.Class({
...
+        // Remove the effect from the actor if necessary
+        if (!this.enabled) {
+            this._removeFromActorOnDisabled(actor);
+        }
+
+        // Watch for the effect to be disabled and remove if it
+        // it does get disabled - saves some CPU cycles computing
+        // things like boundings boxes
+        if (this._wobblyEffectDisabledId === -1) {
+            this._wobblyEffectDisabledId = this.connect('notify::enabled', Lang.bind(this, function() {
```
cosimoc: You could also do this with a vfunc, which would save you the whole signal connect/disconnect boilerplate.

js/ui/windowManager.js
```diff
@@ -189,6 +206,20 @@ var EOSShellWobbly = new Lang.Class({
 
         actor.disconnect(this._positionChangedId);
         this._positionChangedId = null;
+
+        // Remove the effect from the actor if necessary
+        if (!this.enabled) {
```
cosimoc: Remove curly braces here

src/wobbly-effect.c
```diff
@@ -184,10 +192,10 @@ endless_shell_fx_wobbly_get_paint_volume (ClutterEffect    *effect,
...
       ClutterActorBox const extremesBox =
         {
-          floor (x1),
-          floor (y1),
-          ceil (x2),
-          ceil (y2)
+          floor (x1 + offset.x),
+          floor (y1 + offset.y),
+          ceil (x2 + offset.x),
+          ceil (y2 + offset.x)
```
cosimoc: This snippet belongs to the next commit

js/ui/windowManager.js
smspillaz: So, ClutterActorMeta has an 'enabled' property which for a ClutterEffect, will be flipped to FALSE when there is no more animation to do.

We cant remove it immediately on ungrab - there might still be force/velocity/deformation remaining in the model that has yet to play out.

js/ui/windowManager.js
smspillaz: Which vfunc would this be?

src/wobbly-effect.c
smspillaz: Will fix

js/ui/windowManager.js
cosimoc: Presumably though, the `enabled` property is designed to achieve the same that you're trying to achieve here, no? Meaning to avoid doing any work when the animation is not running.
Would it be possible to adjust the effect so that it turns into a no-op when it's disabled, instead of forcibly removing it like this? It looks like that's the usual pattern for clutter effects.

js/ui/windowManager.js
cosimoc: `vfunc_notify` I believe

js/ui/windowManager.js
smspillaz: I could certainly rework the effect to just do nothing when `enabled` is not set.

The current approach is derived from what is typically done in compiz, which is that hooks on paint operations are removed if they are inactive to save CPU cycles on function call overhead. But its probably not a big deal in this case.

PR #269, opened 2018-07-05
T23173: Port to libwobbly-0.3

smspillaz:
This one drops the boost dependency, amongst other things.

https://phabricator.endlessm.com/T23173


smspillaz:
Obviously don't merge until we have libwobbly-0.3 in the distro.

src/meson.build
```diff
@@ -198,7 +198,7 @@ libeos_shell_fx = library('eos-shell-fx',
   sources: libeos_shell_fx_sources,
   dependencies: libeos_shell_fx_deps,
   include_directories: conf_inc,
-  cpp_args: ['-std=c++11'],
+  cpp_args: ['-std=c++14'],
```
joaquimrocha: Don't we need C++ 17?

smspillaz: This is a good point. That said ... I was planning on doing a C API for libwobbly so that we can do away with a lot of this C++ business in the shell. I'll close this PR for now and open a new one on the subtask.

PR #270, opened 2018-07-07
T23208: Drop libwobbly dependency, switch to libwobbly-glib-0.3

smspillaz:
This PR drops the dependency on libwobbly's C++ API and moves to using libwobbly-glib (the C API). Amongst other things, using this API means that we no longer have a single C++ file cluttering up the build.

https://phabricator.endlessm.com/T23208

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -110,8 +110,8 @@ m_dep = cc.find_library('m', required: false)
 eosmetrics_dep = dependency('eosmetrics-0')
 
 # Endless-specific: Wobbly windows
-libwobbly_dep = dependency('libwindowfx_wobbly')
-libeos_shell_fx_deps = [mutter_dep, gdk_pixbuf_dep, libwobbly_dep]
+libwobbly_glib_dep = dependency('libwindowfx_wobbly_glib-0.3')
+libeos_shell_fx_deps = [mutter_dep, gdk_pixbuf_dep, libwobbly_glib_dep, m_dep]
 
 # Endless-specific: Required for keyboard layout switcher in password entries
 xkbcommon_dep = dependency('xkbcommon')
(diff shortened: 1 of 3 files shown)


joaquimrocha:
Jenkins: retest this please.

smspillaz:
Ah, we'll need to include libwobbly-glib-0.3 in the distro . I'll chat to @cosimoc about that, the easiest way is probably to migrate it to `endlessm` and create a jenkins job

cosimoc:
Update: I created the repository now in endlessm as a fork of the original smspillaz one, and added the libwobbly job to the eos-build template. I can't run the jenkins-jobs command to update it at the moment. I will try again in a bit.

meson.build
```diff
@@ -102,6 +102,10 @@ if enable_recorder
   recorder_deps += [gst_dep, gst_base_dep, gtk_dep, x11_dep]
 endif
 
+cc = meson.get_compiler('c')
+
+m_dep = cc.find_library('m', required: false)
```
cosimoc: Why is this required? I'd avoid diverging from the upstream build file if possible.

src/wobbly-effect.c
```diff
@@ -298,12 +297,11 @@ endless_shell_fx_wobbly_ungrab (EndlessShellFXWobbly *effect)
    * of the animation */
   if (priv->timeout_id)
     {
```
cosimoc: Remove curly braces here and below, now that they only have one line

meson.build
smspillaz: Its required since EndlesShellFX depends on `libm`. I can probably get rid of EndlessShellFX now that we dont need to compile it separately but would rather do that in another story since it's a lot of churn.

PR #271, opened 2018-07-10
T23224: Skip maximization effect when synchronizing actor state

smspillaz:
https://phabricator.endlessm.com/T23224


js/ui/appActivation.js
```diff
@@ -303,7 +303,7 @@ const SpeedwagonIface = '<node> \
 </node>';
 const SpeedwagonProxy = Gio.DBusProxy.makeProxyWrapper(SpeedwagonIface);
 
-const SpeedwagonSplash = new Lang.Class({
+var SpeedwagonSplash = new Lang.Class({
```
cosimoc: The commit message is wrong. It should say `instead of 'const'`

cosimoc:
Changed the commit message and pushed to master.

PR #274, opened 2018-07-11
T23236: Start builder once the animation is finished

smspillaz:
https://phabricator.endlessm.com/T23236

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -360,6 +360,11 @@ var WindowTrackingButton = new Lang.Class({
     }
 });
 
+function _constructLoadFlatpakValue(app, appManifest) {
+    // add an app_id_override to the manifest to load
+    return appManifest.get_path() + '+' + app.meta_window.get_flatpak_id() + '.Coding';
+}
+
 var CodingSession = new Lang.Class({
     Name: 'CodingSession',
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -360,6 +360,11 @@ var WindowTrackingButton = new Lang.Class({
     }
 });
 
+function _constructLoadFlatpakValue(app, appManifest) {
```
cosimoc: The function was renamed (now it has an underscore in front) in this commit, but the call site was not updated to use the new name.

js/ui/codeView.js
```diff
@@ -643,8 +644,8 @@ var CodingSession = new Lang.Class({
         if (!this.builder) {
             // get the manifest of the application
             // return early before we setup anything
-            let appManifest = _getAppManifest(this.app.meta_window.get_flatpak_id());
-            if (!appManifest) {
+            this.appManifest = _getAppManifest(this.app.meta_window.get_flatpak_id());
```
cosimoc: There should be no reason for this to be "public". You can use `this._appManifest`

js/ui/codeView.js
erikos: Fixed.

PR #273, opened 2018-07-10
codeview: Don't show button while lock screen is active

smspillaz:
https://phabricator.endlessm.com/T23226

PR #275, opened 2018-07-11
codeview: Don't show button while lock screen is active

erikos:
https://phabricator.endlessm.com/T23226


erikos:
This superseeds https://github.com/endlessm/gnome-shell/pull/273 and removes the stacking on top of #270 as it is blocked on finishing libwobbly: Export GObject API https://phabricator.endlessm.com/T23208.

Also a small indentation fix.

PR #279, opened 2018-07-31
Implement scheduled updates warning dialog

uajain:
Currently, if you try to test this PR, it will open up warning's dialog no matter what. I am stuck to resolve that problem (mentioned below in  PR comment) and need help with that.

Also TODO:
* Icons
* More rigorous DevQA 

https://phabricator.endlessm.com/T22888


js/ui/endSessionDialog.js
```diff
@@ -417,6 +461,13 @@ var EndSessionDialog = new Lang.Class({
         this._user.disconnect(this._userChangedId);
     },
 
+    _getUpdatesStatus: function()  {
+        if (this._proxy.EntryCount > 0)
+            this._isUpdatesScheduled = true;
+        else
+            this._isUpdatesScheduled = true; /* FIXME: should be false */
+    },
+
```
uajain: For some reason the `EntryCount` parameter is returning `0` even if there are pending updates scheduled. I don't know why but I can't get past this issue/
As anyone can figure out, I am using `_isUpdatesScheduled` as a flag to know whether there are updates scheduled or not? Is there something wrong with that logic?

pwithnall: Capture a D-Bus log of the relevant traffic when `new SchedulerProxy` is called, and see if `GetAll` is called to get the property values from the scheduler. Check there are no error responses in the log. It’s probably easiest to capture the log using [Bustle](https://www.freedesktop.org/wiki/Software/Bustle/).

uajain: I  figured out what was happening.

I was initially using a single VM for this work and my ongoing work on  gnome-software. More often I used to `killall gnome-software` for `gnome-software`-related work and that used to stop mogwai too (and resetting the properties I think).

Now that if I start mogwai using `service mogwai-scheduled start` and wait for ~1 min, those scheduled downloads objects are spawned (checked using `d-feet`) and `EntryCount` is updated correctly.

pwithnall: Was Mogwai crashing, or was it just reaching its inactivity timeout and exiting gracefully?

js/ui/endSessionDialog.js
```diff
@@ -80,6 +80,24 @@ const logoutDialogContent = {
...
+    descriptionscheduledUpdates: function() {
+        return _("You need to leave the computer on to get scheduled updates");
+    },
+    showBatteryWarning: true,
+    confirmButtons: [{ signal: 'ContinueToShutdown',
+                       label:  C_("button", "Continue to shutdown") },
+                     { signal: 'changeSettings',
+                       label:  C_("button", "Change Settings...") },
+                     ],
+    iconName: 'system-shutdown-symbolic', /* FIXME: Use endless-auto-updates-scheduled-symbolic.svg */
```
uajain: Hey @GeorgesStavracas 

Need a bit of help here. I couldn't figure out how to use ` endless-auto-updates-scheduled-symbolic.svg` as the icon here. Any ideas?

js/ui/endSessionDialog.js
```diff
@@ -327,6 +356,7 @@ var EndSessionDialog = new Lang.Class({
                                                this._sync();
                                            }));
 
+
```
GeorgesStavracas: Unrelated change

js/ui/endSessionDialog.js
```diff
@@ -417,6 +447,18 @@ var EndSessionDialog = new Lang.Class({
...
+    _syncUpdateStatus: function()  {
+        // Start retrieving the Mogwai proxy
+        this._proxy = new SchedulerProxy(Gio.DBus.system,
+                                         'com.endlessm.MogwaiSchedule1',
+                                         '/com/endlessm/DownloadManager1');
+
+        if (this._proxy.EntryCount > 0)
+            this._hasScheduledUpdates = true;
+        else
+            this._hasScheduledUpdates = false;
```
GeorgesStavracas: Simpler: `this._hasScheduledUpdates = this._proxy.EntryCount > 0;`

js/ui/endSessionDialog.js
```diff
@@ -417,6 +447,18 @@ var EndSessionDialog = new Lang.Class({
         this._user.disconnect(this._userChangedId);
     },
 
+    _syncUpdateStatus: function()  {
+        // Start retrieving the Mogwai proxy
+        this._proxy = new SchedulerProxy(Gio.DBus.system,
+                                         'com.endlessm.MogwaiSchedule1',
+                                         '/com/endlessm/DownloadManager1');
```
GeorgesStavracas: The proxy should be initialized in `_init()`, together with the other proxies, and asynchronously.

js/ui/endSessionDialog.js
```diff
@@ -417,6 +447,18 @@ var EndSessionDialog = new Lang.Class({
         this._user.disconnect(this._userChangedId);
     },
 
+    _syncUpdateStatus: function()  {
```
GeorgesStavracas: Hm, now that I'm looking at it again, I believe `_syncAutomaticUpdateStatus` is a better name, even if more verbose. This dialog already has to deal with updates coming from PackageKit, so `_syncUpdateStatus` is ambiguous in this context.

js/ui/endSessionDialog.js
```diff
@@ -80,6 +80,24 @@ const logoutDialogContent = {
     showOtherSessions: false,
 };
 
+
+const scheduledUpdatesDialogContent = {
+    subject: C_("title", "Power Off"),
+    subjectscheduledUpdates: ("title", "You have updates scheduled to happen"),
+    descriptionscheduledUpdates: function() {
```
GeorgesStavracas: `scheduledUpdatesDescription`

js/ui/endSessionDialog.js
```diff
@@ -80,6 +80,24 @@ const logoutDialogContent = {
...
+        return _("You need to leave the computer on to get scheduled updates");
+    },
+    showBatteryWarning: true,
+    confirmButtons: [{ signal: 'ContinueToShutdown',
+                       label:  C_("button", "Continue to shutdown") },
+                     { signal: 'changeSettings',
```
GeorgesStavracas: `ChangeSettings` (capitalize first letter)

js/ui/endSessionDialog.js
```diff
@@ -80,6 +80,24 @@ const logoutDialogContent = {
...
+                       label:  C_("button", "Change Settings...") },
```
GeorgesStavracas: Please use the proper character `…` instead of the three dots.

js/ui/endSessionDialog.js
```diff
@@ -769,6 +843,21 @@ var EndSessionDialog = new Lang.Class({
...
+         * show regular shutdown dialog.
+         */
+        if (this._hasScheduledUpdates && !this._parameters) {
+            // Show the UPDATES_SCHEDULED dialog before the regular shutdown dialog
+            this._type = DialogType.UPDATES_SCHEDULED;
+
+            // Save the parameters in case the user chooses to "Continue to shutdown"
+            this._parameters = parameters;
+            this._invocation = invocation;
+        }
```
GeorgesStavracas: Hm, I think this approach is fundamentally flawed. You are injecting `DialogType.UPDATES_SCHEDULED` on top of the regular shutdown type, but the more correct approach would be sending the correct type anyway. It's a laborious path, sure, since it'd involve adding patches to `gnome-session` and hacking other parts of the code; it's also not feasible now, since it'd require refactoring the Automatic Updates management code to an isolated component in GNOME Shell; but this kind of hack makes me nervous about the amount of rework that will have to be done during the rebase.

It's up to you whether you want to chase a more correct approach now, or later, or never, but this trade-off is something we absolutely must keep in mind from now on.

@nedrichards we agreed on having a smaller, less correct patchset, but please take note of this   ↑↑↑

js/ui/endSessionDialog.js
```diff
@@ -769,6 +843,21 @@ var EndSessionDialog = new Lang.Class({
                 this._type = DialogType.UPGRADE_RESTART;
         }
 
+        this._syncUpdateStatus();
+
+        /* If this._parameters has already saved parameters, it means that we have shown
+         * warning dialog(UPDATES_SCHEDULED DialogType) to the user. So, now skip it and
+         * show regular shutdown dialog.
+         */
```
GeorgesStavracas: Either rework this comment to be more descriptive and shorter, or remove it whatsoever. It doesn't really add to the codebase.

js/ui/endSessionDialog.js
```diff
@@ -80,6 +80,24 @@ const logoutDialogContent = {
     showOtherSessions: false,
 };
 
+
+const scheduledUpdatesDialogContent = {
+    subject: C_("title", "Power Off"),
+    subjectscheduledUpdates: ("title", "You have updates scheduled to happen"),
```
GeorgesStavracas: You can make this field `subject` and remove the `if (dialogContent.subjectscheduleUpdates ...)` condition below. Also, this needs to have the `C_` otherwise it's not marked for translation.

js/ui/endSessionDialog.js
GeorgesStavracas: I had problems in the past too with icons, and ultimately failed to figure out what was going on. Sorry, I don't think I have any valuable tip or comment to say :slightly_frowning_face:

uajain:
Let's close this and see if we can implement this via upstream-first approach.

Thanks for the comments Greg.

PR #288, opened 2018-08-29
wobbly: Remove paint box override hack

smspillaz:
Now that clutter correctly computes the paint volume during
effects as the paint volume *up to* the current effect, there's
no need to override the get_paint_volume vfunc and lie about
the paint box.

https://phabricator.endlessm.com/T23692

### DIFF
--- a/src/wobbly-effect.c
+++ b/src/wobbly-effect.c
@@ -64,58 +64,16 @@ G_DEFINE_TYPE_WITH_PRIVATE (EndlessShellFXWobbly,
                             endless_shell_fx_wobbly,
                             CLUTTER_TYPE_DEFORM_EFFECT)
 
-/* Forward declaration so that we can set the function pointer
- * in enforce_no_effects_paint_box_cleanup below */
-static gboolean
-endless_shell_fx_wobbly_get_paint_volume (ClutterEffect    *effect,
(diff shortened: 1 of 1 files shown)


smspillaz:
> I have a question though: if this removes a change we introduced, should revert that patch instead? 

In this case, the relevant "change" was introduced when we merged in the Wobbly Windows branch to the shell a few years ago in 1e38f :)

joaquimrocha:
> In this case, the relevant "change" was introduced when we merged in the Wobbly Windows branch to the shell a few years ago in 1e38f :)

Okay, I still feel maybe this should be squashed on a rebase, but the changes are approved, so feel free to merge (after solving the current conflicts).

GeorgesStavracas:
@smspillaz I agree that it would be better to rename the title of the commit as `fixup! <title of the commit that introduced it>`. Keep the commit message though, it'll help a lot at the next rebase.

smspillaz:
> I agree that it would be better to rename the title of the commit as fixup! <title of the commit that introduced it>

Hmm, it would be tricky to do that, I suppose, since the relevant libwobbly API changed a little bit in the meantime and (and the file changed from C++ to C). I think it'd make sense on the next rebase to just squash all the wobbly stuff together.

PR #292, opened 2018-08-29
debian: Add gir1.2-animations-dbus dependency to control.in

smspillaz:
https://phabricator.endlessm.com/T23314


smspillaz:
> Needs to be added to control as well?

Ah, its already in `control` (hence the reason why the dependency didn't work the first time around): https://github.com/endlessm/gnome-shell/commit/53bf874c36784a3db536d860f686f4ae84fe0d0e

PR #297, opened 2018-09-20
shell-app-system: obey X-Flatpak-RenamedFrom

wjt:
Tested as follows:

* Installed Pidgin from Flathub (it has `X-Flatpak-RenamedFrom=pidgin.desktop;`)
* Manually edited the `icon-grid-layout` property to replace `org.libreoffice.LibreOffice-writer.desktop` with `libreoffice-writer.desktop` (this has `X-Endless-Alias`) and `im.pidgin.Pidgin.desktop` with `pidgin.desktop`
* Restarted the unmodified shell: Writer was still shown on the desktop but Pidgin disappeared
* Started the modified shell: both were shown on the desktop

https://phabricator.endlessm.com/T22596

### DIFF
--- a/src/shell-app-system.c
+++ b/src/shell-app-system.c
@@ -25,11 +25,26 @@
  */
 #define SHELL_APP_IS_OPEN_EVENT "b5e11a3d-13f8-4219-84fd-c9ba0bf3d1f0"
 
-/* Additional key used to map a renamed desktop file to its previous name;
- * for instance, org.gnome.Totem.desktop would use this key to point to
- * 'totem.desktop'
+/* Additional key used to map a renamed desktop file to its previous name.
+ * For instance, org.gnome.Totem.desktop contains:
+ *
+ *   X-Endless-Alias=totem
+ *
+ * (without the .desktop suffix).
  */
 #define X_ENDLESS_ALIAS_KEY     "X-Endless-Alias"
+
+/* Additional key listing 0 or more previous names for an application. This is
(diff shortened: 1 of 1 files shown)


wjt:
Actually I think we need to write back the file with the alias swapped back – unless our modification to this key will survive Flatpak update?

src/shell-app-system.c
```diff
@@ -113,13 +128,49 @@ add_aliases (ShellAppSystem  *self,
...
+   * be) ;-terminated, not just ;-delimited. This is so you can express an
+   * empty trailing element: "a;;" represents the list ["a", ""].
+   *
+   * It is possible to escape a semicolon in a list: "a\;b;" represents ["a;",
+   * "b"]. We cannot support this here, because g_key_file_get_string(), used
+   * internally by g_desktop_app_info_get_string(), rejects the "\;" sequence
+   * and returns NULL.
+   *
+   * https://gitlab.gnome.org/GNOME/glib/merge_requests/339
+   */
```
GeorgesStavracas: This is a brilliant commit message, but not so much as a comment.

wjt: I can move the bulk of it to the commit message, but I do really want to make a note of it in the code, in the hope that when we have GLib 2.60 with the function I wrote in that MR, we can update this code

wjt:
Pushed a new version of the last patch which moves most of the mega-comment to the commit message, adds a few more notes in the commit message, and adds the missing trailing `;` to the example in the comment.

PR #300, opened 2018-09-24
Fix sorting App Center below other apps when searching for "app"

wjt:
The app center is special, since it's always pinned to the end of the grid
and is not part of the IconGridLayout.

The visible symptom of this is that searching for 'app' would rank other
apps (on my system, Steam, D-Feet and KeepAssXC, whose descriptions happen
to contain 'app') above the app center, which is correctly returned in the
highest-ranked group of results from g_desktop_app_info_search() since its
name matches the search.

I also noticed that this comparison function is poorly behaved; while that
wasn't the cause of this bug, it is worth fixing.

https://phabricator.endlessm.com/T23669

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1599,8 +1599,8 @@ var AppSearchProvider = new Lang.Class({
 
         // resort to keep results on the desktop grid before the others
         results = results.sort(function(a, b) {
-            let hasA = IconGridLayout.layout.hasIcon(a);
-            let hasB = IconGridLayout.layout.hasIcon(b);
+            let hasA = a === EOS_APP_CENTER_ID || IconGridLayout.layout.hasIcon(a);
+            let hasB = b === EOS_APP_CENTER_ID || IconGridLayout.layout.hasIcon(b);
 
             if (hasA)
                 return -1;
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -1602,9 +1602,9 @@ var AppSearchProvider = new Lang.Class({
             let hasA = a === EOS_APP_CENTER_ID || IconGridLayout.layout.hasIcon(a);
             let hasB = b === EOS_APP_CENTER_ID || IconGridLayout.layout.hasIcon(b);
 
-            if (hasA)
+            if (hasA && !hasB)
                 return -1;
-            if (hasB)
+            if (hasB && !hasA)
```
GeorgesStavracas: This can be simplified as `return hasA - hasB;`

wjt: Do you really think treating the booleans as integers and doing arithmetic on them is clearer?

GeorgesStavracas: I am totally biased because C, so yes I do. And I only see benefits there:

 * Less lines of code
 * Sorting functions actually looks like a sorting function, rather than a table test
 * Three less branches to mentally track

GeorgesStavracas: Okay

wjt: (testing the new patch now)

wjt:
Pushed the `hasB - hasA` version, which I have tested.

PR #302, opened 2018-09-26
Improve .desktop file change detection

wjt:
Superset of #303 with a fix for this bug:

* Foo is on the desktop; `org.foo.Foo.desktop` contains the line `Exec=zenity --info --text Foo`
* During the session, the user (or automatic updates) installs an update for Foo; `org.foo.Foo.desktop` now contains the line `Exec=zenity --info --text FooBar`
* User clicks the Foo icon on the desktop; the old `Exec=` line is still used

In practice, this bug manifests itself when migrating from the KDE 4 version of ktuberling in eos-apps (which uses `-caption %c` in its desktop file) to the KDE 5 version on Flathub (which uses `-qwindowtitle %c`). The `-caption` syntax is not supported by the Flathub version; as a result, the app can't be launched from the desktop 

**TODO**:
* [x] the same stale-.desktop-file bug for apps pinned to the taskbar (this is because it doesn't monitor ::installed-changed; the upstream logic in dash.js does) (fixed in this branch)
* [x] the taskbar not updating when the favorite-apps gsetting is modified externally (fixed in this branch)
* [x] if an app is in a folder, and its .desktop file's Icon= key changes while the folder is collapsed, the preview will not be updated until the folder is opened (deferred https://phabricator.endlessm.com/T23905)
* [x] if an app is in a folder, and it's updated while the folder is open, changes to `Exec=` will be ignored until the folder is closed (deferred https://phabricator.endlessm.com/T23905)
* [x] Test this code on master

https://phabricator.endlessm.com/T23877

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -282,6 +282,20 @@ var BaseAppView = new Lang.Class({
             // The icon image changed
             if (newIconInfo && !newIconInfo.equal(oldIconInfo))
                 return true;
+
+            if (item instanceof Shell.App) {
+                if (!(currentIcon instanceof AppIcon) || !currentIcon.app) {
+                    // Should not happen: the IDs are the same so if the new
+                    // item is an app, so should be the existing icon.
+                    return true;
+                }
+
+                if (!Shell.AppSystem.app_info_equal(item.get_app_info(),
+                                                    currentIcon.app.get_app_info()))
+                    return true;
+            }
+            // TODO: deep equality for folders; at present the miniature icons
+            // will not update until the folder is next opened.
         }
 
         return false;
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -627,7 +627,9 @@ const ScrolledIconList = new Lang.Class({
                              app: app });
         }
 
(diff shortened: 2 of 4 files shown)


wjt:
Actually, I'll split the regression & warning fixes into a separate PR…

js/ui/appDisplay.js
```diff
@@ -282,6 +282,20 @@ var BaseAppView = new Lang.Class({
             // The icon image changed
             if (newIconInfo && !newIconInfo.equal(oldIconInfo))
                 return true;
+
+            if (item instanceof Shell.App) {
```
wjt: This is … not pleasant, right? One could put the comparison function onto the `ShellApp` but then you still need a typecheck to discover whether we have one of those or a ShellDesktopDirInfo. Or we could construct the new ViewIcon (probably an expensive operation for what's meant to be a reasonably quick check…), and add the comparison function to its two subclasses.

GeorgesStavracas: I think this approach is fine for now.

GeorgesStavracas: But not in the long run, of course.

wjt:
Since the branches are so similar, I'm taking r+ on #305 as r+ on this too.

PR #303, opened 2018-09-26
Fix unbounded recursion bug & some unrelated warnings

wjt:
The first patch fixes a regression introduced in #299. My (weak) claim is that I missed it in testing because the shell routinely prints warnings, including during startup where the bug would be triggered; as penance, I fixed a few "undefined property" warnings upstream and cherry-picked the fixes here as the middle 2 patches.

https://phabricator.endlessm.com/T22596

### DIFF
--- a/js/ui/endSessionDialog.js
+++ b/js/ui/endSessionDialog.js
@@ -799,7 +799,7 @@ var EndSessionDialog = new Lang.Class({
         let updatePrepared = this._pkOfflineProxy ? this._pkOfflineProxy.UpdatePrepared : false;
         let updatesAllowed = this._updatesPermission && this._updatesPermission.allowed;
 
-        _setCheckBoxLabel(this._checkBox, dialogContent.checkBoxText);
+        _setCheckBoxLabel(this._checkBox, dialogContent.checkBoxText || '');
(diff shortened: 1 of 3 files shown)


js/ui/endSessionDialog.js
```diff
@@ -799,7 +799,7 @@ var EndSessionDialog = new Lang.Class({
         let updatePrepared = this._pkOfflineProxy ? this._pkOfflineProxy.UpdatePrepared : false;
         let updatesAllowed = this._updatesPermission && this._updatesPermission.allowed;
 
-        _setCheckBoxLabel(this._checkBox, dialogContent.checkBoxText);
+        _setCheckBoxLabel(this._checkBox, dialogContent.checkBoxText || '');
```
GeorgesStavracas: Nitpick: I believe this would be better if all entries of `DialogContent` had the `checkBoxText` property set to `null` or empty string.

wjt: In general I prefer that style too. I wrote the patch this way because there are quite a few properties like this one (only defined in a few of the entries, with existence checks wherever they are used) and I didn't really want to get into changing that.

This patch was [merged upstream](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/245) in this form, so let's take it as-is.

PR #305, opened 2018-09-27
Improve .desktop file change detection (eos3.4 version)

wjt:
eos3.4 version of #302. This is the code I tested in situ on a machine running eos3.4. The differences are in these two patches:

* appIconBar: react to external changes to favorites
* appIconBar: remove redundant lookup

On master, Shell.GenericContainer has been removed; on eos3.4, `.actor` is needed in a few places to get from the JS-land object to the Clutter actor. Here's the diff between the two pairs of changed patches: https://gist.github.com/wjt/e8d41854e9492f19fd2fa6f3cda093e1

https://phabricator.endlessm.com/T23877

### DIFF
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -637,6 +637,7 @@ const ScrolledIconList = new Lang.Class({
             this._addButtonAnimated(app);
         }
 
+        appSys.connect('installed-changed', this._onInstalledChanged.bind(this));
         appSys.connect('app-state-changed', this._onAppStateChanged.bind(this));
     },
 
@@ -863,6 +864,12 @@ const ScrolledIconList = new Lang.Class({
 
         if (changed)
             this._updatePage();
+    },
+
+    _onInstalledChanged(appSys) {
+        let appFavorites = AppFavorites.getAppFavorites();
+        appFavorites.reload();
+        this._onAppFavoritesChanged(appFavorites);
     },
 
     _onAppStateChanged: function(appSys, app) {
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -621,7 +621,9 @@ const ScrolledIconList = new Lang.Class({
                              app: app });
         }
 
-        let favorites = AppFavorites.getAppFavorites().getFavorites();
+        let appFavorites = AppFavorites.getAppFavorites();
+        appFavorites.connect('changed', this._onAppFavoritesChanged.bind(this));
+        let favorites = appFavorites.getFavorites();
         for (let i = 0; i < favorites.length; i++) {
             this._addButtonAnimated(favorites[i]);
         }
@@ -794,7 +796,7 @@ const ScrolledIconList = new Lang.Class({
         return -1;
     },
 
-    _addButtonAnimated: function(app) {
+    _addButtonAnimated: function(app, oldChild) {
         if (this._taskbarApps.has(app) || !this._isAppInteresting(app))
             return;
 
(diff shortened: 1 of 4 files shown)


src/shell-app-system.c
```diff
@@ -213,11 +213,41 @@ scan_startup_wm_class_to_id (ShellAppSystem *self)
...
+    g_strcmp0 (g_app_info_get_executable (one_info),
+               g_app_info_get_executable (two_info)) == 0 &&
+    g_strcmp0 (g_app_info_get_commandline (one_info),
+               g_app_info_get_commandline (two_info)) == 0 &&
+    strcmp (g_app_info_get_name (one_info),
+            g_app_info_get_name (two_info)) == 0 &&
+    g_strcmp0 (g_app_info_get_description (one_info),
+               g_app_info_get_description (two_info)) == 0 &&
+    strcmp (g_app_info_get_display_name (one_info),
+            g_app_info_get_display_name (two_info)) == 0 &&
```
GeorgesStavracas: In understand this is just copy-pasted code from the section below, but this apparent random use of `strcmp()` and `g_strcmp0()` looks quite inconsistent. If there's an assumption that `g_app_info_get_display_name()` and `g_app_info_get_name()` are never `NULL`, there should be an assertion in those functions.

Anyway, not something to fix here, but probably worth looking at fixing upstream. Since moving these checks to `g_app_info_equal()` was [actually suggested](https://bugzilla.gnome.org/show_bug.cgi?id=773636#c8) when this code was initially introduced.

src/shell-app-system.c
```diff
@@ -213,7 +213,17 @@ scan_startup_wm_class_to_id (ShellAppSystem *self)
...
-static gboolean
+/**
+ * shell_app_system_app_info_equal:
+ * @one: (transfer none): a #GDesktopAppInfo
+ * @two: (transfer none): a possibly-different #GDesktopAppInfo
+ *
+ * Returns %TRUE if @one and @two can be treated as equal. Compared to
+ * g_app_info_equal(), which just compares app IDs, this function also compares
+ * fields of interest to the shell: icon, name, description, executable, and
+ * should_show().
```
GeorgesStavracas: Missing the final `Returns` annotation, as in:

```
Returns: %TRUE if @one and @two are equal, %FALSE otherwise
```

js/ui/appIconBar.js
```diff
@@ -810,21 +812,59 @@ const ScrolledIconList = new Lang.Class({
...
-            }
         });
         this._taskbarApps.set(app, newChild);
 
-        this._container.add_actor(newActor);
+        if (oldChild) {
+            this._container.replace_child(oldChild.actor, newActor);
+        } else {
+            this._container.add_actor(newActor);
+        }
```
GeorgesStavracas: Nitpick: does not need the curly braces

js/ui/appIconBar.js
```diff
@@ -810,21 +812,59 @@ const ScrolledIconList = new Lang.Class({
...
+                    childToReplace.actor.destroy();
+                changed = true;
+            }
+        }
+
+        // Get rid of any removed favorites
+        for (let [oldApp, oldChild] of this._taskbarApps) {
+            if (!this._isAppInteresting(oldApp)) {
+                oldChild.actor.destroy();
+                this._taskbarApps.delete(oldApp);
```
GeorgesStavracas: This looks suspicious... don't you need to set `chaged = true;` here too?

src/shell-app-system.c
wjt: I agree that the inconsistency makes it harder to read; I'll just make them all use `g_strcmp0()`. No harm in a few bonus `NULL` checks.

I read through those threads and I agree with desrt that changing `g_app_info_equal()`s behaviour to match this is not really an option. If one were adding a deeper equality function to GLib, it would have to be more exhaustive than this handful of checks: anything other than a full, deep comparison on the internal `GKeyFile` (expensive), or perhaps the inode number & mtime of the source files for the two (captured at construction time) would be misleading. So… I also don't like this ad-hoc list here in the shell but it's not clear exactly what should be pushed to GLib. As you say: let's think about this when we have a bit more breathing room!

js/ui/appIconBar.js
wjt: Ah yes, this style is even spelled out in HACKING! Will fix.

js/ui/appIconBar.js
wjt: Yes! Good catch.

src/shell-app-system.c
```diff
@@ -213,11 +213,53 @@ scan_startup_wm_class_to_id (ShellAppSystem *self)
...
+    g_app_info_equal (one_info, two_info) &&
+    g_app_info_should_show (one_info) == g_app_info_should_show (two_info) &&
+    g_strcmp0 (g_desktop_app_info_get_filename (one),
+               g_desktop_app_info_get_filename (two)) == 0 &&
+    g_strcmp0 (g_app_info_get_executable (one_info),
+               g_app_info_get_executable (two_info)) == 0 &&
+    g_strcmp0 (g_app_info_get_commandline (one_info),
+               g_app_info_get_commandline (two_info)) == 0 &&
+    g_strcmp0 (g_app_info_get_name (one_info),
+            g_app_info_get_name (two_info)) == 0 &&
```
GeorgesStavracas: Nitpick: misaligned line

wjt: Argh! Fixed this while squashing the fixups.

wjt:
Squashed and merged. Thanks!

PR #309, opened 2018-10-09
Add Clubhouse component

joaquimrocha:
This component will "catch" notifications coming from the Clubhouse app and display them in a different way than the regular ones, in fact implementing the Clubhouse's Quest View.

https://phabricator.endlessm.com/T23931

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -91,10 +91,10 @@ var ClubhouseNotification = new Lang.Class({
         // doesn't take any new "buttons" parameter into account.
         let buttons = params.buttons;
         if (buttons) {
-            buttons.deep_unpack().forEach(Lang.bind(this, function(button) {
+            buttons.deep_unpack().forEach((button) => {
                 this.addAction(button.label.unpack(),
                                Lang.bind(this, this._onButtonClicked, button));
-            }));
+            });
         }
         this.emit('updated', clear);
     },
@@ -149,12 +149,12 @@ var ClubhouseComponent = new Lang.Class({
                                                           CLUBHOUSE_ID,
                                                           null);
         this._clubhouseProxyHandler =
-            this._clubhouseProxy.connect('notify::g-name-owner', Lang.bind(this, function() {
+            this._clubhouseProxy.connect('notify::g-name-owner', () => {
                 if (!this._clubhouseProxy.g_name_owner) {
                     log('Nothing owning D-Bus name %s, so dismiss the Clubhouse banner'.format(CLUBHOUSE_ID));
                     this._dismissQuestCb();
                 }
-            }));
+            });
 
         this.actor = new St.Widget({ visible: true,
                                      clip_to_allocation: true,
@@ -205,23 +205,17 @@ var ClubhouseComponent = new Lang.Class({
     _overrideAddNotification: function() {
         this._oldAddNotificationFuncPrototype =
             NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync;
-        NotificationDaemon.GtkNotificationDaemon.prototype.ClubhouseData = [CLUBHOUSE_ID,
-                                                                            this._clubhouseSource,
-                                                                            this._oldAddNotificationFuncPrototype];
-        NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync = function(params, invocation) {
+        NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync = (params, invocation) => {
             let [appId, notificationId, notification] = params;
 
-            let clubhouseId = this.ClubhouseData[0];
-
             // If the app sending the notification is the Clubhouse, then use our own source
-            if (appId == clubhouseId) {
-                let source = this.ClubhouseData[1];
-                source.addNotification(notificationId, notification, true);
+            if (appId == CLUBHOUSE_ID) {
+                this._clubhouseSource.addNotification(notificationId, notification, true);
                 invocation.return_value(null);
                 return;
             }
 
-            this.ClubhouseData[2].apply(Main.notificationDaemon._gtkNotificationDaemon, arguments);
+            this._oldAddNotificationFuncPrototype.apply(Main.notificationDaemon._gtkNotificationDaemon, [params, invocation]);
         }
     },
 
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -76,46 +76,14 @@ var ClubhouseNotification = new Lang.Class({
     createBanner: function() {
         return new ClubhouseNotificationBanner(this);
     },
-
-    updateFromParams: function(params) {
-        // Update icon
-        let newGicon = null;
-        let gicon = params.icon;
-        if (gicon)
-            newGicon = Gio.icon_deserialize(gicon);
-
-        let clear = true;
-        this.update('', params.body.unpack(), {gicon: newGicon, clear: clear});
-
-        // We need to set up the actions here because the update method above unfortunately
-        // doesn't take any new "buttons" parameter into account.
-        let buttons = params.buttons;
-        if (buttons) {
-            buttons.deep_unpack().forEach((button) => {
-                this.addAction(button.label.unpack(),
-                               Lang.bind(this, this._onButtonClicked, button));
-            });
-        }
-        this.emit('updated', clear);
-    },
 });
 
 var ClubhouseNotificationSource = new Lang.Class({
     Name: 'ClubhouseNotificationSource',
     Extends: NotificationDaemon.GtkNotificationDaemonAppSource,
 
-    _init: function(appId) {
-        this.parent(appId);
-        this._notification = null;
-    },
-
-    addNotification: function(notificationId, notificationParams, showBanner) {
-        if (this._notification == null)
-            this._notification = new ClubhouseNotification(this, notificationParams);
-        else
-            this._notification.updateFromParams(notificationParams);
-
-        this.notify(this._notification);
+    _createNotification: function(params) {
+        return new ClubhouseNotification(this, params);
     },
 });
 
@@ -152,7 +120,7 @@ var ClubhouseComponent = new Lang.Class({
             this._clubhouseProxy.connect('notify::g-name-owner', () => {
                 if (!this._clubhouseProxy.g_name_owner) {
                     log('Nothing owning D-Bus name %s, so dismiss the Clubhouse banner'.format(CLUBHOUSE_ID));
-                    this._dismissQuestCb();
+                    this._clearBanner();
                 }
             });
 
@@ -192,6 +160,14 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseProxy = null;
     },
 
+    _clearBanner: function() {
+        this.actor.visible = false;
+        if (this._banner) {
+            this.actor.remove_child(this._banner.actor);
+            this._banner = null;
+        }
+    },
+
     _imageUsesClubhouse: function() {
         if (GLib.getenv('CLUBHOUSE_DEBUG_ENABLED'))
             return true;
@@ -222,27 +198,24 @@ var ClubhouseComponent = new Lang.Class({
     _onNotify: function(source, notification) {
         this.actor.visible = true;
 
-        // Only create the banner if it doesn't yet exist, otherwise just rely on the notification
-        // object getting updated.
+        notification.connect('destroy', (notification, reason) => {
+            if (reason != MessageTray.NotificationDestroyedReason.REPLACED)
+                this._dismissQuest();
+
+            this._clearBanner();
+        });
+
         if (!this._banner) {
             this._banner = notification.createBanner();
-            this._bannerCloseHandler =
-                this._banner.connect('close', Lang.bind(this, this._dismissQuestCb));
             this.actor.add_child(this._banner.actor);
             this._reposition();
         }
     },
 
-    _dismissQuestCb: function() {
-        this.actor.visible = false;
-        if (this._banner) {
-            // Stop the quest since the banner has been dismissed
-            if (this._clubhouseProxy.g_name_owner)
-                this._clubhouseSource.activateAction('stop-quest', null);
-
-            this.actor.remove_child(this._banner.actor);
-            this._banner = null;
-        }
+    _dismissQuest: function() {
+        // Stop the quest since the banner has been dismissed
+        if (this._clubhouseProxy.g_name_owner)
+            this._clubhouseSource.activateAction('stop-quest', null);
     },
 
     _reposition: function() {
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -60,6 +60,16 @@ var ClubhouseNotificationBanner = new Lang.Class({
         // Always wrap the body's text
         this._expandedLabel.actor.add_style_class_name('clubhouse-quest-view-label');
     },
+
+    reposition: function() {
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+
+        let margin = 50;
+        this.actor.x = monitor.width - (this.actor.width + margin);
+        this.actor.y = Math.floor((monitor.height - this.actor.height) / 2.0) - margin;
+    },
 });
 
 var ClubhouseNotification = new Lang.Class({
@@ -96,7 +106,6 @@ var ClubhouseComponent = new Lang.Class({
             return;
 
         this._banner = null;
-        this.actor = null;
         this._clubhouseSource = null;
         this._clubhouseProxy = null;
         this._clubhouseProxyHandler = 0;
@@ -123,15 +132,6 @@ var ClubhouseComponent = new Lang.Class({
                 }
             });
 
-        this.actor = new St.Widget({ visible: true,
-                                     clip_to_allocation: true,
-                                     can_focus: true,
-                                     track_hover: true,
-                                     reactive: true,
-                                     layout_manager: new Clutter.BinLayout() });
-
-        Main.layoutManager.addChrome(this.actor);
-
         this._clubhouseSource = new ClubhouseNotificationSource(CLUBHOUSE_ID);
         this._clubhouseSource.connect('notify', Lang.bind(this, this._onNotify));
 
@@ -144,24 +144,21 @@ var ClubhouseComponent = new Lang.Class({
         if (!this._useClubhouse)
             return;
 
-        if (this.actor)
-            this.actor.destroy();
+        this._clearBanner();
 
         if (this._clubhouseProxyHandler > 0)
             this._clubhouseProxy.disconnect(this._clubhouseProxyHandler);
 
-        this.actor = null;
         this._clubhouseSource = null;
-        this._banner = null;
         this._clubhouseProxy = null;
     },
 
     _clearBanner: function() {
-        this.actor.visible = false;
-        if (this._banner) {
-            this.actor.remove_child(this._banner.actor);
-            this._banner = null;
-        }
+        if (!this._banner)
+            return;
+
+        this._banner.actor.destroy();
+        this._banner = null;
     },
 
     _imageUsesClubhouse: function() {
@@ -175,8 +172,6 @@ var ClubhouseComponent = new Lang.Class({
     },
 
     _onNotify: function(source, notification) {
-        this.actor.visible = true;
-
         notification.connect('destroy', (notification, reason) => {
             if (reason != MessageTray.NotificationDestroyedReason.REPLACED)
                 this._dismissQuest();
@@ -186,8 +181,8 @@ var ClubhouseComponent = new Lang.Class({
 
         if (!this._banner) {
             this._banner = notification.createBanner();
-            this.actor.add_child(this._banner.actor);
-            this._reposition();
+            Main.layoutManager.addChrome(this._banner.actor);
+            this._banner.reposition();
         }
     },
 
@@ -195,16 +190,6 @@ var ClubhouseComponent = new Lang.Class({
         // Stop the quest since the banner has been dismissed
         if (this._clubhouseProxy.g_name_owner)
             this._clubhouseSource.activateAction('stop-quest', null);
-    },
-
-    _reposition: function() {
-        let monitor = Main.layoutManager.primaryMonitor;
-        if (!monitor)
-            return;
-
-        let margin = 50;
-        this.actor.x = monitor.width - (this.actor.width + margin);
-        this.actor.y = Math.floor((monitor.height - this.actor.height) / 2.0) - margin;
     },
 });
 var Component = ClubhouseComponent;
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -100,7 +100,6 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseSource = null;
         this._clubhouseProxy = null;
         this._clubhouseProxyHandler = 0;
-        this._oldAddNotificationFuncPrototype = null;
     },
 
     enable: function() {
@@ -136,7 +135,9 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseSource = new ClubhouseNotificationSource(CLUBHOUSE_ID);
         this._clubhouseSource.connect('notify', Lang.bind(this, this._onNotify));
 
-        this._overrideAddNotification();
+        // Inject this source in GtkNotificationDaemon's lookup table
+        Main.notificationDaemon._gtkNotificationDaemon._sources[CLUBHOUSE_ID] =
+            this._clubhouseSource;
     },
 
     disable: function() {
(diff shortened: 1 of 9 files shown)


js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+    },
+
+    updateFromParams: function(params) {
+        // Update actions/buttons
+        this.actions = [];
+
+        if (this._buttonBox) {
+            this._buttonBox.destroy();
+            this._buttonBox = null;
+        }
```
cosimoc: It looks like we could call `this.update()` with `clear=true` to get this?

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+
+    _onNotify: function(source, notification) {
+        this.actor.visible = true;
+
+        // Only create the banner if it doesn't yet exist, otherwise just rely on the notification
+        // object getting updated.
+        if (!this._banner) {
+            this._banner = notification.createBanner();
+            this._bannerCloseHandler = this._banner.connect('dismiss',
+                                                            Lang.bind(this, this._dismiss_quest_cb))
```
cosimoc: Missing semicolon

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        log('Enabling Clubhouse component');
+
+        this._clubhouseProxy = new Gio.DBusProxy.new_sync(sessionBus,
+                                                          Gio.DBusProxyFlags.DO_NOT_AUTO_START_AT_CONSTRUCTION,
+                                                          null,
+                                                          CLUBHOUSE_ID,
+                                                          CLUBHOUSE_DBUS_OBJ_PATH,
+                                                          CLUBHOUSE_ID,
+                                                          null);
+        this._clubhouseProxy.connect('notify::g-name-owner', Lang.bind(this, function() {
```
cosimoc: We should use arrow functions instead of `Lang.bind` in new code.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        if (!this._banner) {
+            this._banner = notification.createBanner();
+            this._bannerCloseHandler = this._banner.connect('dismiss',
+                                                            Lang.bind(this, this._dismiss_quest_cb))
+            this.actor.add_child(this._banner.actor);
+            this._reposition();
+        }
+    },
+
+    _dismiss_quest_cb: function() {
```
cosimoc: Typically camel case is used as naming convention for functions in the shell.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        this.actor.visible = false;
+        if (this._banner) {
+            log('Dismissing Clubhouse quest banner');
+
+            // Stop the quest since the banner has been dismissed
+            if (this._clubhouseProxy.g_name_owner)
+                this._banner.notification._activateAction('app.stop_quest', null);
```
cosimoc: This ends up just calling `activateAction()` on the associated source object, which you have here too.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+            this._banner = null;
+        }
+    },
+
+    _reposition: function() {
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+
+        let margin = 50;
```
cosimoc: It would be better to use a CSS style property for this.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        this.actor.x = monitor.width - (this.actor.width + margin);
+        this.actor.y = monitor.height / 2.0 - this.actor.height / 2.0 - margin;
```
cosimoc: Does the quest not want to control the position? It's OK if we want to keep this as a future improvement.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        this.actor.remove_style_class_name('notification-banner');
+        this.actor.add_style_class_name('clubhouse-quest-view');
+        this._iconBin.add_style_class_name('clubhouse-quest-view-icon-bin');
+
+        // Always wrap the body's text
+        this.bodyLabel.actor.clutter_text.line_wrap = true;
+        this.bodyLabel.actor.add_style_class_name('clubhouse-quest-view-label');
+
+        this._closeButton.connect('clicked', Lang.bind(this, function() {
+            this.emit('dismiss');
```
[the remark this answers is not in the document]
joaquimrocha: I did this because I noticed that the base class destroys the notification when the closed signal is emitted, and in the beginning I was creating a new banner every time but wanted to keep only one notification. I have tried now and since I keep only one notification and one banner, I can reuse the `close` signal. Thanks!

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+
+        // We don't have an "unexpanded" state for now
+        this.expanded = true;
+        this._actionBin.visible = true;
+
+        this.actor.can_focus = true;
+        this.actor.track_hover = true;
+
+        // Override the style name because this is a not a regular notification
+        this.actor.remove_style_class_name('notification-banner');
```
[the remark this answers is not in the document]
joaquimrocha: The idea was to clear any notion of the `notification-banner`. Why should we keep the old style?

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+
+var ClubhouseNotificationBanner = new Lang.Class({
+    Name: 'ClubhouseNotificationBanner',
+    Extends: MessageTray.NotificationBanner,
+
+    _init: function(notification) {
+        this.parent(notification);
+
+        // We don't have an "unexpanded" state for now
+        this.expanded = true;
```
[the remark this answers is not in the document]
joaquimrocha: Thanks!

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+var ClubhouseNotification = new Lang.Class({
+    Name: 'ClubhouseNotification',
+    Extends: NotificationDaemon.GtkNotificationDaemonNotification,
+
+    _init: function(source, notification) {
+        this._currentIconPath = null;
+
+        let gicon = notification.icon;
+        if (gicon) {
+            let newGicon = this._getClubhouseIcon(Gio.icon_deserialize(gicon));
```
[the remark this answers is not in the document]
joaquimrocha: > I'm not entirely sure what this code is trying to do, but there are a few problems with this approach:

I am using a `GFileIcon` from the Clubhouse side. This [serializes the content as the URI](https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gfileicon.c#L259).
I implemented it this way because if we use use the real icon's data we will be sending that over D-Bus all the time (even if we cache stuff in the Shell side). So I thought it'd be more optimal to just send the icon URI, but since the icons are shipped with the Clubhouse Flatpak, I have to look into it.
I know probably the optimization is negligible here, so if you really prefer to send data all the time I can do it.

> * it feels a bit awkward in general to have the shell look into the clubhouse Flatpak to find these icons

I understand but this is how I managed to skip sending the icons' data all the time.


> * if we need to have the shell look at the clubhouse Flatpak, the current code looks into the Flatpak user installation to locate icons, but the clubhouse will be installed system-wide

That's a mistake, I intended to add the counterpart check but forget to do it. Thanks for spotting it.

> * `this._getClubhouseIcon()` can return `null` but the code here does not account for that

I will take care of that.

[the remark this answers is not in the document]
joaquimrocha: I will keep the current approach of using `FileIcons` and looking into the app for now. If you still disagree after my explanation, we can change it.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+    disable: function() {
+        if (!this._useClubhouse)
+            return;
+
+        log('Disabling Clubhouse component');
+
+        if (this.actor)
+            this.actor.destroy();
+
+        if (this._clubhouseProxyHandler > 0)
```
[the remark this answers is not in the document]
joaquimrocha: Sorry.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        // We are only using the Clubhouse component in Endless Hack images for now, so
+        // check if the prefix of the image version matches the mentioned flavor.
+        return eosImageVersion.startsWith('hack-');
+    },
+
+    _overrideAddNotification: function() {
+        this._oldAddNotificationFuncPrototype =
+            NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync;
+        NotificationDaemon.GtkNotificationDaemon.prototype.ClubhouseData = [CLUBHOUSE_ID,
+                                                                            this.source];
```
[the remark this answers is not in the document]
joaquimrocha: I tried that in the beginning but it was failing when destroying the existing source for the app ID (because the messageTray's code runs before the components one).
Maybe it had to do with the fact that I had saved notifications, but in any case it seemed to be a safer way to do this.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+    },
+});
+
+var ClubhouseNotificationSource = new Lang.Class({
+    Name: 'ClubhouseNotificationSource',
+    Extends: NotificationDaemon.GtkNotificationDaemonAppSource,
+
+    _init: function(appId) {
+        this.parent(appId);
+        this._notification = null;
```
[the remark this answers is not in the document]
joaquimrocha: The parent class always recreates a new notification, this means we'd have to always reset the notification in the banner anyway. So I preferred to always reuse the same notification object which will also update the banner.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+    updateFromParams: function(params) {
+        // Update actions/buttons
+        this.actions = [];
+
+        if (this._buttonBox) {
+            this._buttonBox.destroy();
+            this._buttonBox = null;
+        }
+
+        let buttons = params.buttons;
```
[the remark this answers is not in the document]
joaquimrocha: This is the `params.buttons` given for the notification, and since I am reusing an action, I still had to reset/clear actions when updating it.

js/ui/components/clubhouse.js
joaquimrocha: The problem with that is that the icon will also be cleared, which means that it'd always have to load the icon again. We can try to work around this by keeping the icons cached (if we care at all).

js/ui/components/clubhouse.js
joaquimrocha: I will have to check how that's done.

js/ui/components/clubhouse.js
joaquimrocha: Arrgh. I did fix a bunch of snake-case vars but apparently I missed this one. Thanks.

js/ui/components/clubhouse.js
joaquimrocha: We will change this value according to where the banner is (to unblock screen real estate), so I think maybe we keep this here?

js/ui/components/clubhouse.js
joaquimrocha: Yeah, maybe I got ahead of myself.

joaquimrocha:
@cosimoc I have added a fixup commit with all the fixes from the discussions I closed as resolved.
Let's talk about the remaining.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
cosimoc: I am sure it can be made to work either way, but it's a bit unusual to remove style classes, since it can cause unexpected issues when e.g. selectors exist that target a child through a parent->child chain.
Anyway, you can keep it like this if you prefer it this way.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
cosimoc: @joaquimrocha I understand why it was done like this, but I do feel that it would be cleaner to send the image data over the wire for now; as long as it's done once per-notification the overhead should not be too bad as these are not large images. We can of course revisit it if it turns out to be a problem. I'm also keen on avoiding a dependency between the Shell and Flatpak for now.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        this._oldAddNotificationFuncPrototype = null;
+    },
+
+    enable: function() {
+        if (!this._useClubhouse) {
+            log('Cannot enable Clubhouse in this image version');
+            return;
+        }
+
+        log('Enabling Clubhouse component');
```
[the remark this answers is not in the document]
cosimoc: Yeah, if everything followed that logic, we would have enormous logs! :-)
I wish we had a way to print opt-in debug statements in the shell, but until then let's remove these.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,334 @@
...
+const Clutter = imports.gi.Clutter;
+const Flatpak = imports.gi.Flatpak
+const Gio = imports.gi.Gio;
+const GLib = imports.gi.GLib;
+const GObject = imports.gi.GObject;
+const Shell = imports.gi.Shell;
+const Signals = imports.signals;
+const St = imports.gi.St;
+
+const GnomeSession = imports.misc.gnomeSession;
```
cosimoc: This is unused

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,334 @@
...
+const Lang = imports.lang;
+const Main = imports.ui.main;
+const MessageList = imports.ui.messageList;
```
cosimoc: This is unused

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,334 @@
...
+const MessageTray = imports.ui.messageTray;
+const Monitor = imports.ui.monitor;
```
cosimoc: This is unused

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,334 @@
...
+    _getClubhouseAppDeployDir: function(installation) {
+        let ref = installation.get_current_installed_app(CLUBHOUSE_ID, null);
+        if (ref)
+            return ref.get_deploy_dir()
+
+        return null;
+    },
+
+    _getClubhouseAsset: function(assetSubpath) {
+        let app_dir = this._getClubhouseAppDeployDir(Flatpak.Installation.new_user(null))
```
cosimoc: Use camel case

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,334 @@
...
+        if (!iconFile)
+            return null;
+
+        return Gio.FileIcon.new(iconFile);
+    },
+
+    _getClubhouseAppDeployDir: function(installation) {
+        let ref = installation.get_current_installed_app(CLUBHOUSE_ID, null);
+        if (ref)
+            return ref.get_deploy_dir()
```
cosimoc: Missing semicolon

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
cosimoc: I see, I had missed that code in `notificationDaemon.js`. I feel that the way this is setup is going against how GtkNotificationDaemon notifications are supposed to work.
In other words, I'd like this code to run as part of the regular constructor of `GtkNotificationDaemonNotification` instead of here. More on that below.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
cosimoc: Following up from the previous commit, I would like to see if we can take the path of least resistance here, and avoid overriding the behavior to always use a single notification.

`g_application_send_notification()` explicitly mentions that notifications with the same ID will be replaced when receiving an update, which is already the default behavior.
Moreover, your implementation will prevent us from having two clubhouse notifications at the same time, which can be an interesting use case.

js/ui/components/clubhouse.js
```diff
@@ -0,0 +1,323 @@
...
+        }));
+
+        this.actor = new St.Widget({ visible: true,
+                                     clip_to_allocation: true,
+                                     can_focus: true,
+                                     track_hover: true,
+                                     reactive: true,
+                                     layout_manager: new Clutter.BinLayout() });
+
+        Main.layoutManager.addChrome(this.actor);
```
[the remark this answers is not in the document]
cosimoc: Not sure if you saw this question.

js/ui/components/clubhouse.js
cosimoc: I'm OK keeping this whole part as it is until we have a better idea of the positioning requirements.

js/ui/components/clubhouse.js
joaquimrocha: Left overs from a few versions before.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: I prefer to keep it because in principle our notifications should have anything to do with the default ones in terms of styling.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: Okay. I have changed that.

js/ui/components/clubhouse.js
joaquimrocha: I have changed this too.

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: Okay. I have no time to change all that implies using the original code. I am happy if you want to change it, hopefully you will see my point better, or I will see yours :smile:

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: As in my previous comment, I cannot deliver this that way today. Feel free to change it if you have time. Or we can always do it later too...

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: Do you mean long as we have now? :trollface: 
Alright, I will change that :+1:

js/ui/components/clubhouse.js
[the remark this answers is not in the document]
joaquimrocha: Sorry, I did see it but forgot to reply.
I copied this from the old Game Manager I believe. You seemed to me as well that this would give us a better control if, e.g. we decide to show two banners. But if you want to change it, please go ahead (I can no longer stare at the screen today).

js/ui/components/clubhouse.js
```diff
@@ -209,23 +209,17 @@ var ClubhouseComponent = new Lang.Class({
     _overrideAddNotification: function() {
         this._oldAddNotificationFuncPrototype =
             NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync;
-        NotificationDaemon.GtkNotificationDaemon.prototype.ClubhouseData = [CLUBHOUSE_ID,
-                                                                            this.source,
-                                                                            this._oldAddNotificationFuncPrototype];
-        NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync = function(params, invocation) {
+        NotificationDaemon.GtkNotificationDaemon.prototype.AddNotificationAsync = (params, invocation) => {
```
joaquimrocha: Hmm, I had that whole hacky way of passing the data because I noticed that the `this` inside the prototype's body referred to the `GtkNotificationDaemon`. Maybe the using the arrow func changes that?

js/ui/components/clubhouse.js
```diff
@@ -67,10 +67,6 @@ var ClubhouseNotification = new Lang.Class({
     Extends: NotificationDaemon.GtkNotificationDaemonNotification,
 
     _init: function(source, notification) {
-        this._currentIconPath = null;
-
-        notification.title = GLib.Variant.new('s', '');
```
joaquimrocha: This is actually used. Because in the Clubhouse I was not setting any title (and without a title, it won't work).
Maybe we can change it so that we just set the empty title if none is coming in the notification...

js/ui/components/clubhouse.js
```diff
@@ -67,10 +67,6 @@ var ClubhouseNotification = new Lang.Class({
     Extends: NotificationDaemon.GtkNotificationDaemonNotification,
 
     _init: function(source, notification) {
-        this._currentIconPath = null;
```
joaquimrocha: Thanks.

js/ui/components/clubhouse.js
```diff
@@ -252,7 +252,7 @@ var ClubhouseComponent = new Lang.Class({
 
         let margin = 50;
         this.actor.x = monitor.width - (this.actor.width + margin);
-        this.actor.y = monitor.height / 2.0 - this.actor.height / 2.0 - margin;
+        this.actor.y = Math.floor((monitor.height - this.actor.height) / 2.0) - margin;
```
joaquimrocha: `ClutterActor.y` is a float. Do we really need this?

js/ui/components/clubhouse.js
```diff
@@ -127,13 +127,10 @@ var ClubhouseComponent = new Lang.Class({
     Name: 'ClubhouseComponent',
 
     _init: function() {
-        this._useClubhouse = false;
-
-        let eosImageVersion = Shell.util_get_eos_image_version();
-        if (eosImageVersion == null || !this._imageUsesClubhouse(eosImageVersion))
+        this._useClubhouse = this._imageUsesClubhouse();
+        if (!this._useClubhouse)
```
joaquimrocha: Awesome! I will squash this into the original commit (hope that's okay).

js/ui/notificationDaemon.js
```diff
@@ -713,6 +713,10 @@ var GtkNotificationDaemonAppSource = new Lang.Class({
         return new FdoApplicationProxy(Gio.DBus.session, this._appId, this._objectPath, callback);
     },
 
+    _createNotification: function(params) {
+        return new GtkNotificationDaemonNotification(this, params);
+    },
+
```
joaquimrocha: Yeah. I tried to avoid this, so why I when for the prototype solution. I guess it's fine, especially since you sent it upstream too.

js/ui/messageTray.js
```diff
@@ -50,12 +50,15 @@ var State = {
...
+// that were destroyed as a result of a user action, SOURCE_CLOSED for the
+// notifications that were requested to be destroyed by the associated source,
+// and REPLACED for notifications that were destroyed as a consequence of a
+// newer version having replaced them.
 var NotificationDestroyedReason = {
     EXPIRED: 1,
     DISMISSED: 2,
-    SOURCE_CLOSED: 3
+    SOURCE_CLOSED: 3,
+    REPLACED: 4
```
joaquimrocha: I understand you want to make the code as simple as possible, but IMO this is bit less safe for the purpose of the Shell Quest View, because it is not really a notification, so if we're relying on the original GTK behavior, we always have to make sure that changes from upstream won't break the Quest View.
Of course with my original changes, things could be affected in a rebase as well, but it was less dependent on the GTK notificatoins side.

src/shell-util.c
```diff
@@ -725,3 +730,47 @@ shell_util_check_cloexec_fds (void)
...
+    }
+
+  return image_version;
+}
+
+/**
+ * shell_util_get_eos_image_version:
+ *
+ * Convenience function to get the EOS version for the image.
+ */
```
GeorgesStavracas: This needs a `Returns: (transfer full): ` annotation

js/ui/messageTray.js
cosimoc: I consider this patch to be a needed improvement upstream regardless of the use we make of it in this PR.

I don't agree that this is less safe than relying on other behavior from a class not specifically written for this purpose. We'll have to make sure this feature keeps working one way or another after a rebase, so saying that relying on a specific behavior could be safer than relying on another is really a speculation right now. Either way, I will assume that we are OK proceeding with this change right now.

js/ui/components/clubhouse.js
cosimoc: Yeah, actors need to be grid-aligned, or you'll get artifacts when rendering. In the case of the ASUS E406MA laptop I am testing on, the text was blurry/segmented because the actor was not vertically aligned to the pixel grid.

js/ui/components/clubhouse.js
cosimoc: The simplest thing to do is to ensure the clubhouse conforms to the expectations the shell has, so I added a patch to set an empty title there. I don't think we need to change the shell expectations in this case.

js/ui/components/clubhouse.js
cosimoc: Yes, arrow functions are a better way of doing the Lang.bind() dance. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_separate_this

js/ui/components/clubhouse.js
joaquimrocha: Thanks for the explanation!

js/ui/messageTray.js
joaquimrocha: Yeah. I think both views have their pros and cons. But I agree with moving on with what we have. Thanks.

PR #311, opened 2018-10-12
windowManager: let codeview handle windows while speedwagon is up

cosimoc:
We have a second code path to account for windows mapped while the
speedwagon is up. Make sure CodeViewManager handles that case too.

https://phabricator.endlessm.com/T24057


erikos:
Thanks Georges for taking a look.

PR #313, opened 2018-10-13
codeView: enable for all applications

cosimoc:
Now that we have a default lockscreen implementation, we can remove
any special case from the shell.
All applications are going to have a FlipToHack button, and those
without a more specialized implementation, are going to get a
lockscreen by default.

https://phabricator.endlessm.com/T24015


cosimoc:
Let's get this in for playtest.

PR #314, opened 2018-10-15
T24053

joaquimrocha:
https://phabricator.endlessm.com/T24053


js/ui/components/clubhouse.js
```diff
@@ -27,6 +27,7 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
 const Shell = imports.gi.Shell;
+const SideComponent = imports.ui.sideComponent;
```
cosimoc: Please move this down together with the other "local" imports.

js/ui/components/clubhouse.js
```diff
@@ -38,6 +39,20 @@ const NotificationDaemon = imports.ui.notificationDaemon;
 const CLUBHOUSE_ID = 'com.endlessm.Clubhouse';
 const CLUBHOUSE_DBUS_OBJ_PATH = '/com/endlessm/Clubhouse';
 
+const ClubhouseIface =
+'<node> \
+  <interface name="com.endlessm.Clubhouse"> \
+    <method name="show"> \
+    <arg type="u" direction="in" name="timestamp"/> \
```
cosimoc: Indentation is off here

js/ui/components/clubhouse.js
```diff
@@ -147,10 +159,22 @@ var ClubhouseComponent = new Lang.Class({
...
         if (this._clubhouseProxyHandler > 0)
-            this._clubhouseProxy.disconnect(this._clubhouseProxyHandler);
+            this.proxy.disconnect(this._clubhouseProxyHandler);
 
+        this.parent();
         this._clubhouseSource = null;
-        this._clubhouseProxy = null;
+    },
+
+    notifyHideAnimationCompleted: function() {
```
cosimoc: Does not look like this method exists on the clubhouse interface.

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    Name: 'ClubhouseButton',
+    Extends: St.Button,
+
+    _init: function(clubhouseProxy) {
+        this.parent();
+        this._clubhouseProxy = clubhouseProxy;
+
+        this.add_style_class_name('clubhouse-shell-button');
+
+        this._icon_size = 64
```
cosimoc: Use camel-case

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+
+        this.x = monitor.width - Math.floor(this._icon_size / 2.0);
+        this.y = Math.floor(monitor.height / 2.0) - Math.floor(this.height / 2.0);
+    },
+
+    updateVisibility: function() {
+        let clubhouseShowing = this._clubhouseProxy && this._clubhouseProxy.g_name_owner &&
```
cosimoc: I don''t believe the proxy will ever be null; similarly I don't think `Visible` will ever be true if the proxy does not have an owner. So it should be possible to considerably simplify this condition.

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        });
+    },
+
+    reposition: function() {
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+
+        this.x = monitor.width - Math.floor(this._icon_size / 2.0);
+        this.y = Math.floor(monitor.height / 2.0) - Math.floor(this.height / 2.0);
```
cosimoc: It would be simpler to write this as `Math.floor((monitor.height - this.height) / 2.0)`

js/ui/components/clubhouse.js
```diff
@@ -141,9 +186,20 @@ var ClubhouseComponent = new Lang.Class({
...
                 }
             });
 
+        this._clubhouseButton = new ClubhouseButton(this.proxy);
+        this._clubhouseButton.connect('clicked', () => {
+            let time = GLib.DateTime.new_now_local().to_unix();
+            if (Main.overview.visible)
+                this.show(time);
+            else
+                this.toggle(time);
```
cosimoc: `global.get_current_time()` is what we should pass here.

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+
+        this.add_style_class_name('clubhouse-shell-button');
+
+        this._icon_size = 64
+        let icon = new St.Icon({icon_name: 'user-home-symbolic',
+                                icon_size: this._icon_size});
+        this.add_actor(icon);
+        this.reposition();
+
+        Main.overview.connect('shown', () => {
```
cosimoc: If we're only calling in a method without parameters, the simplest syntax for these callbacks is
```
Main.overview.connect('shown', this.updateVisibility.bind(this));
```

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+            this.updateVisibility();
+        });
+        Main.overview.connect('hidden', () => {
+            this.updateVisibility();
+        });
+        Main.sessionMode.connect('updated', () => {
```
cosimoc: It should not be necessary to connect to this signal. See below.

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,50 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+            return;
+
+        this.x = monitor.width - Math.floor(this._icon_size / 2.0);
+        this.y = Math.floor(monitor.height / 2.0) - Math.floor(this.height / 2.0);
+    },
+
+    updateVisibility: function() {
+        let clubhouseShowing = this._clubhouseProxy && this._clubhouseProxy.g_name_owner &&
+            this._clubhouseProxy.Visible;
+        this.visible = !Main.sessionMode.isLocked && !Main.sessionMode.isGreeter &&
```
cosimoc: The clubhouse component is disabled altogether by the `SessionMode` machinery when the session mode is switched to a locked or greeter session.
The way to handle it would be to call `removeChrome()` on the button, or destroy it, when the clubhouse component is disabled.

js/ui/components/clubhouse.js
joaquimrocha: It was a left over from C&P... Thanks.

joaquimrocha:
@cosimoc Here's a new version that has both the open and the close buttons.

When I dismiss the Clubhouse I am still getting this log, but I've seen it in other parts of the Shell before (and I can try to fix it tomorrow too).

As for the issue I mentioned to you on Slack (resetting the Shell while the Clubhouse is opened), I have added a "correction" logic that is also defensive even if we fix the real issue later. I actually have an idea for a solution but it's not too elegant and I can no longer work on it tonight. Since the issue is not very common, and gets corrected with the last commit, in case your review is positive you can also merge it and I open a task about the Shell restart tomorrow.

```
Oct 16 23:04:16 endless gnome-shell[22661]: Object Meta.WindowActor (0x55ad809d0c10), has been already finalized. Impossible to get any property from it.
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: == Stack trace for context 0x55ad7e996650 ==
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #0 0x55ad7ee9d358 i   resource:///org/gnome/shell/ui/tweener.js:73 (0x7f445c0c8cd0 @ 9)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #1 0x55ad7ee9d2d8 i   resource:///org/gnome/shell/ui/tweener.js:105 (0x7f445c0c8f78 @ 36)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #2 0x55ad7ee9d250 i   resource:///org/gnome/shell/ui/tweener.js:92 (0x7f445c0c8de0 @ 52)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #3 0x7ffcd9279ea0 b   resource:///org/gnome/gjs/modules/tweener/tweener.js:208 (0x7f445c0d3918 @ 54)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #4 0x7ffcd927a750 b   resource:///org/gnome/gjs/modules/tweener/tweener.js:337 (0x7f445c0d39a0 @ 1626)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #5 0x7ffcd927a800 b   resource:///org/gnome/gjs/modules/tweener/tweener.js:350 (0x7f445c0d3a28 @ 100)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #6 0x7ffcd927afd0 b   resource:///org/gnome/gjs/modules/tweener/tweener.js:365 (0x7f445c0d3ab0 @ 10)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #7 0x7ffcd927be10 b   resource:///org/gnome/gjs/modules/signals.js:128 (0x7f445c0cff78 @ 386)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #8 0x7ffcd927c600 b   resource:///org/gnome/shell/ui/tweener.js:208 (0x7f445c0cf5e8 @ 159)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #9 0x7ffcd927cd20 I   resource:///org/gnome/gjs/modules/_legacy.js:82 (0x7f445c0b5de0 @ 71)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #10 0x7ffcd927d4a0 b   resource:///org/gnome/shell/ui/tweener.js:183 (0x7f445c0cf560 @ 20)
Oct 16 23:04:16 endless org.gnome.Shell.desktop[22661]: #11 0x7ffcd927dcc0 b   self-hosted:918 (0x7f445c0d7ef0 @ 394)
```

js/ui/components/clubhouse.js
```diff
@@ -211,5 +367,9 @@ var ClubhouseComponent = new Lang.Class({
         if (this.proxy.g_name_owner)
             this._clubhouseSource.activateAction('stop-quest', null);
     },
+
+    _onVisibilityChanged: function() {
+        this.parent();
+    },
```
cosimoc: This does not do anything and can be removed

js/ui/components/clubhouse.js
```diff
@@ -123,6 +265,8 @@ var ClubhouseComponent = new Lang.Class({
 
         this.parent(ClubhouseIface, CLUBHOUSE_ID, CLUBHOUSE_DBUS_OBJ_PATH);
 
+        this._clubhouseButton = null;
```
cosimoc: `this._clubhouseButton` is never set

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
     },
 });
 
+var ClubhouseButtonManager = new Lang.Class({
+    Name: 'ClubhouseButtonManager',
+
+    _init: function() {
+        this.parent();
```
cosimoc: This does nothing, since there is no parent class

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+
+        this._iconSize = 64
+
+        this._openButton =
+            new St.Button({ name: 'clubhouse-open-button',
```
cosimoc: I don't think you need to name this actor (or the one below)

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -967,3 +967,20 @@ popup-separator-menu-item {
 .clubhouse-quest-view-icon-bin > StIcon {
   icon-size: 5em;
 }
+
+.clubhouse-shell-button {
+  border-radius: 20px;
+  background-color: $selected_bg_color;
+}
```
cosimoc: It does not look like this style is used anywhere

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+
+        this._overviewHiddenHandler = Main.overview.connect('hidden', () => {
+            this.updateVisibility();
+        });
+    },
+
+    _getClubhouseActorFromWM: function() {
+        for (let actor of global.get_window_actors()) {
+            let window = actor.meta_window;
+            if (window && window.get_gtk_application_id() == CLUBHOUSE_ID)
```
cosimoc: There should not be a case where you get an actor without an associated window

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    },
+
+    updateVisibility: function() {
+        this._closeButton.visible = !Main.overview.visible && this._clubhouseWindowActor;
+        this._openButton.visible = !this._closeButton.visible;
+    },
+
+    _mapWindow: function(shellwm, actor) {
+        let metaWindow = actor.meta_window;
+        if (metaWindow && metaWindow.get_gtk_application_id() == CLUBHOUSE_ID)
```
cosimoc: There should not be a case where you get an actor without an associated window

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        if (this._clubhouseWindowActor)
+            this._closeButton.x = this._clubhouseWindowActor.x - this._iconSize / 2;
+    },
+
+    reposition: function() {
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+
+        this._openButton.x = monitor.width - this._iconSize / 2;
```
cosimoc: If you use `this._iconSize` for this calculation, you're assuming that the button will not add any borders/margins/paddings that change its size. That may be true at the moment, but a better way to get this information is to look at the width of the button.

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        this._openButton.y = Math.floor(monitor.height / 2.0 - this._iconSize / 2.0);
+
+        this._updateCloseButtonPosition();
+    },
+
+    updateVisibility: function() {
```
cosimoc: This can be "private"

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        return null;
+    },
+
+    _updateCloseButtonPosition: function() {
+        this._closeButton.y = this._openButton.y;
+        if (this._clubhouseWindowActor)
+            this._closeButton.x = this._clubhouseWindowActor.x - this._iconSize / 2;
+    },
+
+    reposition: function() {
```
cosimoc: This can be "private" (start with an underscore)

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
     },
 });
 
+var ClubhouseButtonManager = new Lang.Class({
+    Name: 'ClubhouseButtonManager',
+
+    _init: function() {
+        this.parent();
+
+        this._iconSize = 64
```
cosimoc: Missing semicolon

js/ui/components/clubhouse.js
```diff
@@ -112,6 +112,148 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+            this._clubhouseDestroyHandler = 0;
+        }
+
+        this._clubhouseWindowActor = actor;
+
+        if (!actor)
+            return;
+
+        // Track new window actor
+        this._clubhouseNotifyHandler = actor.connect('notify::x', (actor, prop) => {
```
cosimoc: I am not sure why you need to keep track of the `x` property to determine what feels like the visibility of the clubhouse window. Don't we already have a visible property in the DBus interface that can be used for this?

js/ui/components/clubhouse.js
```diff
@@ -291,7 +291,15 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseButtonManager = new ClubhouseButtonManager();
         this._clubhouseButtonManager.connect('open-clubhouse', () => {
             let time = global.get_current_time();
-            this.show(time);
+
+            // When the Shell is restarted, we cannot really get the Clubhouse window actor
+            // if the Clubhouse was running, which means the close button won't be shown;
+            // thus, the 'toggle' action below allows users to at least hide the Clubhouse
+            // window and reopen it, which would fix the issue
```
cosimoc: It feels like we should spend some time trying to understand why there are no window actors yet by the time this component is initialized, and how to defer initialization until the windows are known to the compositor, instead of putting this hack here.

js/ui/components/clubhouse.js
joaquimrocha: Sorry. I thought I had removed that before pushing the last version...

js/ui/components/clubhouse.js
joaquimrocha: It's not for the visibility, it's for setting the `closeButton`'s position. I may have left those visibility bits after one of the many attempts to get it right. I will see if I can remove those bits and leave only the close button's position.

js/ui/components/clubhouse.js
joaquimrocha: Okay. I will try the solution I have in mind. Even if I fix the issue, this logic is harmless so we can consider keeping it to prevent eventual problems of having the Clubhouse button shown when the Clubhouse is also being shown (as if that happens there's no way to close the Clubhouse, making it a bigger issue than having the wrong button being shown).

joaquimrocha:
@cosimoc I have done another run with this and I found a signal to connect to that makes the buttons work on a shell restart and also allows to simplify the code a lot. It's in a fixup commit so you can more easily see the changes. Once approved I will squash + merge.

GeorgesStavracas:
Do not wait for my review here

PR #316, opened 2018-10-15
clubhouse: only enable component when clubhouse is installed

cosimoc:
Or we will segfault later in case it isn't.

https://phabricator.endlessm.com/T24097

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -101,7 +101,7 @@ var ClubhouseComponent = new Lang.Class({
     Name: 'ClubhouseComponent',
 
     _init: function() {
-        this._useClubhouse = this._imageUsesClubhouse();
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -161,6 +161,10 @@ var ClubhouseComponent = new Lang.Class({
         this._banner = null;
     },
 
+    _haveClubhouse: function() {
```
joaquimrocha: Nitpick: I'd prefer the method to be called `_hasClubhouse` because it sounds better and more clear than `this._haveClubhouse`. It'd be also more consistent with glib/gtk libs.
I am fine if you don't want to change it though.

cosimoc: Fixed

PR #317, opened 2018-10-17
T24015

cosimoc:
https://phabricator.endlessm.com/T24015

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -68,6 +68,18 @@ const FlipIcon = GObject.registerClass(class FlipIcon extends St.Icon {
     }
 });
 
+function _ensureAfterFirstFrame(win, callback) {
+    if (win._drawnFirstFrame) {
+        callback();
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -499,7 +472,7 @@ var CodingSession = new Lang.Class({
 
     _isCurrentWindow: function(window) {
         let actor = this._actorForCurrentState();
-        return actor.meta_window === window;
+        return actor && actor.meta_window === window;
```
GeorgesStavracas: I do not see how the other changes might make `this._actorForCurrentState()` return `null` here.

cosimoc: `this._windowsNeedSync()` only returns `true` if both "sides" of the session are present (app and hacking toolbox).

Before this commit, `this._isCurrentWindow()` would assume that `this._actorForCurrentState()` always returns a non-`null` actor, since otherwise we would not have reached that code as `this._windowsNeedsSync()` would have returned `false`.

However, now I want to update the button also in the case only the app side is present, so I added this condition.

PR #319, opened 2018-10-17
Support alternative PAYG code formats

wjt:
There are four parts to this:

* Always ask the daemon if it's enabled, rather than reading its config file (https://github.com/endlessm/eos-payg/pull/24 teaches the daemon to time out after 30 seconds if PAYG is disabled, and in any case its resource usage is minimal)
* Pick up the CodeFormat property introduced in https://github.com/endlessm/eos-payg/pull/24
* Rework code validation to use that property, rather than assuming all codes must be 8 digits
* Adjust the CSS a bit so that the longer codes used by the provider plugin at https://github.com/endlessm/eos-payg-nonfree/pull/1 fit in the available space

https://phabricator.endlessm.com/T23965

### DIFF
--- a/js/misc/paygManager.js
+++ b/js/misc/paygManager.js
@@ -85,34 +85,6 @@ const notificationAlertTimesSecs = [
     30,           // 30 seconds
 ];
 
-// This function checks the configuration file of PAYG directly
-// from the expected locations on disk, on an attempt to figure
-// out whether the feature is enabled, so that we don't wake up
-// the D-Bus service and keep it running when it's disabled.
-function _isPaygEnabled() {
-    // See man page eos-payg.conf(5)
-    let searchDirs = [
-        '/etc/eos-payg',
-        '/usr/local/share/eos-payg',
-        '/usr/share/eos-payg',
-    ];
-
-    let configFileName = 'eos-payg.conf'
-    let keyfile = new GLib.KeyFile();
-    try {
-        keyfile.load_from_dirs(configFileName,
-                               searchDirs,
-                               GLib.KeyFileFlags.NONE);
-        return keyfile.get_boolean('PAYG', 'Enabled');
-    } catch (e) {
-        // A non-existent file is a perfectly normal use case.
-        if (!e.matches(GLib.KeyFileError, GLib.KeyFileError.NOT_FOUND))
-            logError(e, "Error reading PAYG configuration file from " + configFileName);
-    }
-
-    return false;
-}
-
 // Takes an UNIX timestamp (in seconds) and returns a string
 // with a precision level appropriate to show to the user.
 //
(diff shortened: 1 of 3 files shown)


js/misc/paygManager.js
```diff
@@ -35,18 +35,18 @@ const Signals = imports.signals;
...
+        <method name="AddCode">
+            <arg type="s" direction="in" name="code"/>
+        </method>
+        <method name="ClearCode" />
+        <signal name="Expired" />
+        <property name="ExpiryTime" type="t" access="read"/>
+        <property name="Enabled" type="b" access="read"/>
+        <property name="RateLimitEndTime" type="t" access="read"/>
+    </interface>
+</node>`;
```
GeorgesStavracas: I enjoy the indentation fix, but using backticks breaks gettext pretty badly. See https://gitlab.gnome.org/GNOME/gnome-shell/issues/537

wjt: Yuck! I guess I'll have to remove this. Adding a separate resource file etc. seems a little heavy handed unless the whole shell is getting this treatment.

wjt:
> did not break anything at least

I'm using this branch rebased onto 3.4 on my laptop which has payg enabled; both locking when credit expires and unlocking by entering a code seems to still work!

wjt:
Testing this change now…

wjt:
After a bit of a fiasco where I based my changes on an old version of this branch, pushed ca7b9c0adcc73b54b57f3ab7f53ebd3a31e43782 after testing it. The weird issue I was seeing with the lock screen seems only to happen if I mess with the PAYG daemon's state in ways that are not possible with the UI (namely, entering new codes while the device is unlocked). This is buggy but it's not a regression. We do have a ticket about adding UI to add new codes before your current credit expires, so I'll make a note there.

js/misc/paygManager.js
```diff
@@ -350,6 +375,16 @@ var PaygManager = new Lang.Class({
         this._proxy.ClearCodeRemote();
     },
 
+    validateCode(code, partial = false) {
```
GeorgesStavracas: Nitpick: remove extra spaces around `=`

js/ui/paygUnlockDialog.js
```diff
@@ -152,7 +146,15 @@ var PaygUnlockCodeEntry = new Lang.Class({
...
-        if (!this._enabled || !GLib.unichar_isdigit(character))
+        if (!this._enabled || !GLib.unichar_isprint(character))
+            return;
+
+        let pos = this.clutter_text.get_cursor_position();
+        let before = pos === -1 ? this._code : this._code.slice(0, pos);
+        let after = pos === -1 ? "" : this._code.slice(pos);
+        let newCode = before + character + after;
+
+        if (!Main.paygManager.validateCode(newCode, true /* partial */))
```
GeorgesStavracas: Nitpick: use `//`, move the comment above the `if ()` line

js/misc/paygManager.js
wjt: Might be nice to include this in the HACKING style guide.

js/ui/paygUnlockDialog.js
wjt: I don't think the comment is clear unless it's inline:

```js
// partial
if (!Main.paygManager.validateCode(newCode, true))
```

so I'll just remove it.

PR #320, opened 2018-10-18
codeView: add shader effect for flipped window

cosimoc:
We add a shader effect on the back of the window, as for the latest
design. The shader is a gradient map with a 5-point color gradient.

https://phabricator.endlessm.com/T24093


src/shell-code-view-effect.c
```diff
@@ -0,0 +1,288 @@
...
+                              paint_opacity,
+                              paint_opacity,
+                              paint_opacity);
+  cogl_push_source (self->pipeline);
+
+  cogl_rectangle (0, 0,
+                  cogl_texture_get_width (texture),
+                  cogl_texture_get_height (texture));
+
+  cogl_pop_source ();
```
GeorgesStavracas: I understand this is basically a copy paste of ClutterDessaturateEffect, but I have a request that will help me in the future. Instead of using the implicit framebuffer API (`push|pop_source()`, `cogl_rectangle()`, etc), can you please use the explicit variants? It'll be something like

```c
  CoglFramebuffer *fb = cogl_get_draw_framebuffer ();

  ...

  cogl_framebuffer_draw_rectangle (fb, self->pipeline, 0, 0, width, height);
```

src/shell-code-view-effect.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef __SHELL_CODE_VIEW_EFFECT_H__
+#define __SHELL_CODE_VIEW_EFFECT_H__
+
+#include <clutter/clutter.h>
+
+G_BEGIN_DECLS
+
+#define SHELL_TYPE_CODE_VIEW_EFFECT          (shell_code_view_effect_get_type ())
+#define SHELL_CODE_VIEW_EFFECT(obj)          (G_TYPE_CHECK_INSTANCE_CAST ((obj), SHELL_TYPE_CODE_VIEW_EFFECT, ShellCodeViewEffect))
+#define SHELL_IS_CODE_VIEW_EFFECT(obj)       (G_TYPE_CHECK_INSTANCE_TYPE ((obj), SHELL_TYPE_CODE_VIEW_EFFECT))
```
GeorgesStavracas: Worth using `G_DECLARE_FINAL_TYPE()`

src/shell-code-view-effect.c
```diff
@@ -0,0 +1,288 @@
...
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author:
+ *   Emmanuele Bassi <ebassi@linux.intel.com>
+ *   Cosimo Cecchi <cosimo@endlessm.com>
+ */
+
+#define SHELL_CODE_VIEW_EFFECT_CLASS(klass) (G_TYPE_CHECK_CLASS_CAST ((klass), SHELL_TYPE_CODE_VIEW_EFFECT, ShellCodeViewEffectClass))
+#define SHELL_IS_CODE_VIEW_EFFECT_CLASS(klass) (G_TYPE_CHECK_CLASS_TYPE ((klass), SHELL_TYPE_CODE_VIEW_EFFECT))
+#define SHELL_CODE_VIEW_EFFECT_GET_CLASS(obj) (G_TYPE_INSTANCE_GET_CLASS ((obj), SHELL_TYPE_CODE_VIEW_EFFECT, ShellCodeViewEffectClass))
```
GeorgesStavracas: These will go out if `G_DECLARE_FINAL_TYPE()` is used; also, the first two ones aren't used anyway.

cosimoc:
Thanks @joaquimrocha and @GeorgesStavracas. I addressed your comments.

PR #321, opened 2018-10-18
clubhouse: position notification banner at the top right

erikos:
https://phabricator.endlessm.com/T24064


erikos:
I reduced the margin, since it gets close on the left side to the DiscoveryFeed button.

PR #323, opened 2018-10-20
T24113

cosimoc:
https://phabricator.endlessm.com/T24113

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -80,6 +80,12 @@ function _ensureAfterFirstFrame(win, callback) {
     });
 };
 
+function _getAppId(win) {
+    let app = Shell.WindowTracker.get_default().get_window_app(win);
+    // remove .desktop suffix
+    return app.get_id().slice(0, -8);
+};
+
 function _getWindowId(win) {
     if (win.gtk_window_object_path)
         return win.gtk_window_object_path;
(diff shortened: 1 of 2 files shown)


js/ui/codeView.js
```diff
@@ -1013,17 +1016,18 @@ var CodeViewManager = new Lang.Class({
...
-            let [targetBusName, targetObjectPath] = variant.deep_unpack();
+            let [targetAppId, targetWindowId] = variant.deep_unpack();
             let session = this._getSessionForTargetApp(
-                targetBusName, targetObjectPath);
+                targetAppId, targetWindowId);
 
             if (session)
                 handled = session.admitToolboxWindowActor(actor);
         } else {
+            let shellApp = Shell.WindowTracker.get_default().get_window_app(actor.meta_window);
```
joaquimrocha: Are we guaranteed to have an app here? When I was working on Clubhouse Shell buttons I noticed that I couldn't get the app for the Clubhouse based on its actor's meta window (I didn't have time to investigate further, and using the GTK ID worked for me).

joaquimrocha: Are we guaranteed to have an app here? When I was working on Clubhouse Shell buttons I noticed that I couldn't get the app for the Clubhouse based on its actor's meta window (I didn't have time to investigate further, and using the GTK ID worked for me).

cosimoc: Yeah, side components are treated specially in that respect for some reason, and would not be returned by that function, but it will work for any of the regular apps.

cosimoc:
@GeorgesStavracas approved this on Slack.

PR #329, opened 2018-10-29
windowManager: use unmanaged signal to unregister animations

cosimoc:
Avoid the destroy animation code path, since there are many cases
when that callback returns early (e.g. when the overview is visible),
which will still lead to an unbounded array of animatable surfaces
left behind.

The 'unmanaged' signal is emitted by Mutter when a MetaWindow is
destroyed, so this commit uses that instead.

https://phabricator.endlessm.com/T24192

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -2224,6 +2224,11 @@ var WindowManager = new Lang.Class({
                 let parent = window.get_transient_for();
                 if (parent)
                     this._checkDimming(parent);
+
+                if (actor._animatableSurface) {
+                    this._animationsServer.unregister_surface(actor._animatableSurface);
+                    actor._animatableSurface = null;
+                }
         }));
 
         // Endless libanimation extension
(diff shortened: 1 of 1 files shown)


js/ui/windowManager.js
```diff
@@ -2224,6 +2224,11 @@ var WindowManager = new Lang.Class({
                 let parent = window.get_transient_for();
                 if (parent)
                     this._checkDimming(parent);
+
+                if (actor._animatableSurface) {
+                    this._animationsServer.unregister_surface(actor._animatableSurface);
+                    actor._animatableSurface = null;
+                }
```
GeorgesStavracas: Can you confirm this isn't necessary at https://github.com/endlessm/gnome-shell/pull/329/commits/a11c8dd72d3fae0748c2aa990ed1645d54c1df9c#diff-3411cdb80aaaddfcaa2c6d036a436b69L2461 as well?

cosimoc:
@GeorgesStavracas approved this after talking about it on Slack.

js/ui/windowManager.js
cosimoc: It is not necessary to also have it there, because the `unmanaged` signal I connect to is emitted for every window where the effect is registered in the first place.

PR #331, opened 2018-10-30
clubhouse: Don't allow the notification source to be fully destroyed

joaquimrocha:
Turns our that the Shell Quest View could only be used once because the
notification source we create for it gets destroyed and removed from
the GtkNotificationDaemon's list. Since have to keep the source assigned
to the Clubhouse's app ID in the GtkNotificationDaemon, we should
prevent the source from being fully destroyed.

Thus, these changes implement the "destroy" vfunc for the
ClubhouseNotificationSource, doing everything its parent does except for
emitting the "destroy" signal.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -110,6 +110,17 @@ var ClubhouseNotificationSource = new Lang.Class({
     _createNotification: function(params) {
         return new ClubhouseNotification(this, params);
     },
+
+    // Override the method so we don't emit the 'destroy' signal, which would end up
+    // actually destroying the source and removing it from the GtkNotificationDaemon,
+    // meaning we could only use the source (and thus the Shell Quest View) once.
+    destroy: function(reason) {
+        let notifications = this.notifications;
+        this.notifications = [];
+
+        for (let i = 0; i < notifications.length; i++)
+            notifications[i].destroy(reason);
+    },
 });
 
 var ClubhouseButtonManager = new Lang.Class({


js/ui/components/clubhouse.js
```diff
@@ -110,6 +110,17 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    // Override the method so we don't emit the 'destroy' signal, which would end up
+    // actually destroying the source and removing it from the GtkNotificationDaemon,
+    // meaning we could only use the source (and thus the Shell Quest View) once.
+    destroy: function(reason) {
+        let notifications = this.notifications;
+        this.notifications = [];
+
+        for (let i = 0; i < notifications.length; i++)
+            notifications[i].destroy(reason);
+    },
```
cosimoc: I'm a bit confused as to how we can get into the state where the `destroy` signal is emitted on the clubhouse source, and I haven't been able to reproduce the issue at will on my laptop today. Do you have a set of steps that reliably reproduce this issue?

However, instead of overriding this method, you could also re-create another identical source if this one is destroyed... That would probably be more robust, as it does not depend on the implementation details of the Soruce class.

joaquimrocha: It's easy to reproduce it: Click on Ada for the Fizzics Intro quest (or any quest, I guess), then click the "invite-out" button, then close the Shell Quest View so the Clubhouse reopens, and try to do these steps again: the Shell Quest View (notification banner) is only shown the first time.

As for the solution you propose, I understand it's not elegant to use details of the base class's implementation, but if I understand correctly, you're suggesting we connect to to the destroy signal (or override the destroy method) to re-add the source again, and this seems even more hacky IMO (i.e. if the prob is that the source gets destroyed, then we should prevent it from being destroyed). It also may prevent any eventual race conditions between adding a new source while the old one is getting removed if a notification arrives in the meanwhile.

cosimoc:
Went for a different solution in #336

PR #332, opened 2018-10-30
codeView: support different shader types

cosimoc:
In particular, we are interested right now in supporting a 'none' type
for apps such as HackUnlock for which we want no shader.

In the future, this can be extended to have different apps choose
different shader types.

https://phabricator.endlessm.com/T24241

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -24,6 +24,13 @@ const STATE_TOOLBOX = 1;
 
 const _FLIP_ICON_FRONT = 'resource:///org/gnome/shell/theme/flip-symbolic.svg';
 const _HACKABLE_DESKTOP_KEY = 'X-Endless-Hackable';
+const _HACK_SHADER_DESKTOP_KEY = 'X-Endless-HackShader';
+
+const _HACK_SHADER_MAP = {
+    'none': null,
+    'desaturate': Shell.CodeViewEffect
+};
+const _HACK_DEFAULT_SHADER = 'desaturate';
 
 const FlipIcon = GObject.registerClass(class FlipIcon extends St.Icon {
     _init(props = {}) {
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -453,12 +459,24 @@ var CodingSession = new Lang.Class({
...
+            effect.enabled = enabled;
         } else {
-            desaturate = new Shell.CodeViewEffect({ enabled: enabled });
-            actor.add_effect_with_name('codeview-desaturate', desaturate);
+            let appInfo = this._shellApp.get_app_info();
+            let shaderEffect = appInfo.get_string(_HACK_SHADER_DESKTOP_KEY);
+            let shaderProto;
+
+            // By default, all apps get Shell.CodeViewEffect
+            if (shaderEffect)
```
joaquimrocha: Nitpick: I'd rather set the `shaderEffect` to the key that has the default effect (`let shaderEffect = appInfo.get_string(_HACK_SHADER_DESKTOP_KEY) || 'desaturate';`). But of course this works too.

js/ui/codeView.js
```diff
@@ -453,12 +459,24 @@ var CodingSession = new Lang.Class({
...
+                shaderProto = _HACK_SHADER_MAP[shaderEffect];
+            else
+                shaderProto = Shell.CodeViewEffect;
+
+            if (shaderProto) {
+                effect = new shaderProto({ enabled: enabled });
```
ptomato: coding style nitpick: In modern JS, you can write `new shaderProto({enabled})`

js/ui/codeView.js
cosimoc: Good suggestion, fixed.

js/ui/codeView.js
cosimoc: Fixed.

PR #333, opened 2018-10-31
gtk-embed: Don't allow multiple async allocations

GeorgesStavracas:
HACK!!! There is an asynchronous roundtrip cycle that happens in this
code. It goes through:

 1. Clutter calls ClutterActor.allocate() on ShellEmbeddedWindow
 2. That calls into here, which calls _shell_embedded_window_allocate
 3. ShellEmbeddedWindow calls gdk_window_move_resize()
 4. That generates an XEvent
 5. X11 sends that event to Mutter
 6. meta_compositor_sync_window_geometry()
 7. meta_window_actor_sync_actor_geometry()
 8. MetaWindowActor.sync_actor_geometry() sets the position and size of
    the systray actor, and that causes a relayout
 9. goto 1

This is a known problem, and ShellEmbeddedWindow works around that by
checking if the position and sizes actually changed. If they didn't,
bail out of the allocation phase and stop generating the XEvents - and
that used to work.

Right now though there is a bug in between Clutter and St that causes
the cached X position to be garbage by the time we need to check it to
avoid relayouts. This causes an neverending asynchronous loop between
Xorg and GNOME Shell when there is any systray icon around.

While the actual bug isn't fixed, though, manually keeps track of the
number of allocations and prevent the layout cycle. 3 is the hardcoded
number because:

 * 1st allocation happen during the initial size negotiation, and it
   must be kept
 * 2nd allocation is the response to the 1st one
 * 3rd is the actual, valid allocation that Clutter does

Anything beyond that should be ignored.

This commit must be reverted once the real bug in the Clutter allocation
machinery is fixed.

https://phabricator.endlessm.com/T24189


GeorgesStavracas:
For more context, this was discussed in today's standup and we agreed on merging this hack first (the ticket is a release blocker) while I keep tracking the actual bug in the size negotiation machinery in Clutter.

src/shell-gtk-embed.c
```diff
@@ -280,6 +282,47 @@ shell_gtk_embed_allocate (ClutterActor          *actor,
...
+   * Right now though there is a bug in between Clutter and St that causes
+   * the cached X position to be garbage by the time we need to check it to
+   * avoid relayouts. This causes an neverending asynchronous loop between
+   * Xorg and GNOME Shell when there is any systray icon around.
+   *
+   * While the actual bug isn't fixed, though, manually keeps track of the
+   * number of allocations and prevent the layout cycle. 3 is the hardcoded
+   * number because:
+   *
+   *  * 1st allocation happen during the initial size negotiation, and it
```
cosimoc: `happens`

src/shell-gtk-embed.c
```diff
@@ -280,6 +282,47 @@ shell_gtk_embed_allocate (ClutterActor          *actor,
...
+   * checking if the position and sizes actually changed. If they didn't,
+   * bail out of the allocation phase and stop generating the XEvents - and
+   * that used to work.
+   *
+   * Right now though there is a bug in between Clutter and St that causes
+   * the cached X position to be garbage by the time we need to check it to
+   * avoid relayouts. This causes an neverending asynchronous loop between
+   * Xorg and GNOME Shell when there is any systray icon around.
+   *
+   * While the actual bug isn't fixed, though, manually keeps track of the
```
cosimoc: `keep track`

src/shell-gtk-embed.c
```diff
@@ -280,6 +282,47 @@ shell_gtk_embed_allocate (ClutterActor          *actor,
...
+   * number of allocations and prevent the layout cycle. 3 is the hardcoded
+   * number because:
+   *
+   *  * 1st allocation happen during the initial size negotiation, and it
+   *    must be kept
+   *  * 2nd allocation is the response to the 1st one
+   *  * 3rd is the actual, valid allocation that Clutter does
+   *
+   * Anything beyond that should be ignored.
```
cosimoc: Is that really what this code does though?
It looks to me that the way you have it, this will happen:
- the first four times `shell_gtk_embed_allocate()` is called, the allocation will go through to the embedded window
- starting from the fifth time this is called, every other time `shell_gtk_embed_allocate()` is called, the allocation will go through to the embedded window

Maybe I'm missing something?

cosimoc:
I'm OK using a hacky workaround until we understand the bug better, but the code does not seem to 100% reflect the intention described in the comment.

src/shell-gtk-embed.c
GeorgesStavracas: That's right. Every other allocation is ignored (to break the Xorg - Mutter - Shell - GDK cycle). I'll rephrase that.

PR #334, opened 2018-10-31
switcherPopup: Bind to the stage, not the monitor

GeorgesStavracas:
https://phabricator.endlessm.com/T24191


GeorgesStavracas:
@cosimoc approved on Slack

PR #335, opened 2018-11-01
notificationDaemon: create the clubhouse notification source earlier

cosimoc:
Instead of injecting it into the sources table when the component
starts up, change _ensureAppSource() to create it when required.

Note that it's important that clubhouse.js is imported after the
GtkNotification classes are defined in notificationDaemon.js,
otherwise subclassing will fail.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -36,7 +36,7 @@ const MessageTray = imports.ui.messageTray;
 const NotificationDaemon = imports.ui.notificationDaemon;
 const SideComponent = imports.ui.sideComponent;
 
-const CLUBHOUSE_ID = 'com.endlessm.Clubhouse';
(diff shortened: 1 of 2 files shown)


joaquimrocha:
Hi @cosimoc , for the notification to be shown a second time you need the fixup commit I have added here. Also please notice the explanation in the fixup's commit body about how the whole issue is not fixed by this approach.

BTW, you still use your approach by adding a `getNotificationSourceForAppId` function, and override it in the prototype from the Clubhouse's component side, meaning you'd have all these without having to patch or early initialize the source.

joaquimrocha:
Uhh, for the suggestion in the second paragraph of my previous comment to work you'd have of course to destroy any existing source because of the time when things are initialized.

cosimoc:
Went for a different approach in #336

PR #336, opened 2018-11-01
clubhouse: Use the notification source outside of the GtkNotification…

joaquimrocha:
…Daemon

We need to have control over our custom notification source for the
Clubhouse, but since it is managed like all the sources in the
GtkNotificationDaemon, it is difficult to reconnect the source when
it is destroyed a new one is created.

This patch changes the approach used to use the Clubhouse's notification
source by overriding the notification daemon's prototype. This way the
notification daemon knows nothing about the source, and we get to
control it as we desire.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -36,6 +36,8 @@ const MessageTray = imports.ui.messageTray;
 const NotificationDaemon = imports.ui.notificationDaemon;
 const SideComponent = imports.ui.sideComponent;
 
+const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
+
 const CLUBHOUSE_ID = 'com.endlessm.Clubhouse';
 const CLUBHOUSE_DBUS_OBJ_PATH = '/com/endlessm/Clubhouse';
 
@@ -247,6 +249,8 @@ var ClubhouseComponent = new Lang.Class({
         this._banner = null;
         this._clubhouseSource = null;
         this._clubhouseProxyHandler = 0;
+
+        this._oldAddNotificationFunc = null;
     },
 
     enable: function() {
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -304,6 +307,20 @@ var ClubhouseComponent = new Lang.Class({
         this.proxy.hideRemote(timestamp);
     },
 
+    _getClubhouseSource: function() {
+        if (this._clubhouseSource != null)
+            return this._clubhouseSource;
+
+        this._clubhouseSource = new ClubhouseNotificationSource(CLUBHOUSE_ID);
+        this._clubhouseSource.connect('notify', Lang.bind(this, this._onNotify));
+        this._clubhouseSource.connect('destroy', () => {
```
joaquimrocha: I still think that (especially if we use the prototype hack) we should prevent the clubhouse source to be destroyed by the behavior of the base class as there's no need for it to be destroyed when it runs out of notifications.

js/ui/components/clubhouse.js
```diff
@@ -294,6 +293,10 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseButtonManager.destroy();
         this._clubhouseButtonManager = null;
         this._clubhouseSource = null;
+
+        // Reset the AddNotificationAsync old prototype
+        if (this._oldAddNotificationFunc)
+            GtkNotificationDaemon.prototype.AddNotificationAsync = this._oldAddNotificationFunc;
```
joaquimrocha: We should discuss this in the context of the notification disappearing when locking the screen (and not coming back on unlock). Also, if a quest message is shown while the screen is locked, then it'd hit the regular notifications source. So I'd rather remove the resetting of the old function prototype.

PR #337, opened 2018-11-02
clubhouse: Pulse the "open" button to indicate the user to click it

manuq:
Add a new signal PulsingIcon in the D-Bus interface. When triggered,
the icon of the Open button is styled with a "highlighted" class.

When the Open button is clicked, the "highlighted" class is removed.

The pulsing animation is done in the CSS by transforming two images in
loop.

The clubhouse-icon-highlighted.png image is a placeholder made by
coloring the icon for the normal state.

https://phabricator.endlessm.com/T24164

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -51,6 +51,7 @@ const ClubhouseIface =
       <arg type="u" direction="in" name="timestamp"/> \
     </method> \
     <property name="Visible" type="b" access="read"/> \
+    <signal name="SuggestOpen" /> \
   </interface> \
 </node>';
 
@@ -118,12 +119,30 @@ var ClubhouseNotificationSource = new Lang.Class({
     },
 });
 
+var ClubhouseOpenButton = new Lang.Class({
+    Name: 'ClubhouseOpenButton',
+    Extends: St.Button,
+
+    _init: function(params) {
+        params = params || {};
+        this._icon = new St.Icon({ style_class: 'clubhouse-open-button-icon' });
+        params.child = this._icon;
+        this.parent(params);
+    },
+
+    setHighlighted: function(highlighted) {
+        if (highlighted)
+            this._icon.add_style_pseudo_class('highlighted');
+        else
+            this._icon.remove_style_pseudo_class('highlighted');
+    },
+});
+
 var ClubhouseButtonManager = new Lang.Class({
     Name: 'ClubhouseButtonManager',
 
     _init: function() {
-        this._openButton =
-            new St.Button({ child: new St.Icon({ style_class: 'clubhouse-open-button-icon' }) });
+        this._openButton = new ClubhouseOpenButton();
         this._openButton.connect('clicked', () => { this.emit('open-clubhouse'); })
 
         Main.layoutManager.addChrome(this._openButton);
@@ -223,6 +242,10 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
     },
 
+    setSuggestOpen: function(suggestOpen) {
+        this._openButton.setHighlighted(suggestOpen);
+    },
+
     destroy: function() {
         Main.overview.disconnect(this._overviewShowingHandler);
         this._overviewShowingHandler = 0;
@@ -262,12 +285,15 @@ var ClubhouseComponent = new Lang.Class({
         this._clubhouseSource = null;
         this._oldAddNotificationFunc = null;
         this._clubhouseProxyHandler = 0;
+        this._clubhouseProxySuggestOpenHandler = 0;
 
         this._clubhouseButtonManager = new ClubhouseButtonManager();
         this._clubhouseButtonManager.connect('open-clubhouse', () => {
+            this._clubhouseButtonManager.setSuggestOpen(false);
             this.show(global.get_current_time());
         });
         this._clubhouseButtonManager.connect('close-clubhouse', () => {
+            this._clubhouseButtonManager.setSuggestOpen(false);
             this.hide(global.get_current_time());
         });
 
@@ -289,6 +315,13 @@ var ClubhouseComponent = new Lang.Class({
                     this._clearBanner();
                 }
             });
+        }
+
+        if (this._clubhouseProxySuggestOpenHandler == 0) {
+            this._clubhouseProxySuggestOpenHandler =
+                this.proxy.connectSignal('SuggestOpen', () => {
+                    this._clubhouseButtonManager.setSuggestOpen(true);
+                });
         }
 
         this._enabled = true;
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -972,6 +972,11 @@ popup-separator-menu-item {
   width: 80px;
   height: 80px;
   background-image: url('resource:///org/gnome/shell/theme/clubhouse-icon-normal.png');
+  border: none;
+  &:highlighted {
+    border: 5px solid rgba(150, 220, 255, 0.7);
+    border-radius: 40px;
+  }
 }
 
 .clubhouse-close-button-icon {


js/ui/components/clubhouse.js
```diff
@@ -51,6 +51,7 @@ const ClubhouseIface =
       <arg type="u" direction="in" name="timestamp"/> \
     </method> \
     <property name="Visible" type="b" access="read"/> \
+    <property name="PulsingIcon" type="b" access="read"/> \
```
manuq: Actually I need a `<signal>` not a property.

js/ui/components/clubhouse.js
```diff
@@ -217,6 +221,16 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
     },
 
+    startPulsingIcon: function() {
+        this._openButton.get_children()[0].add_style_pseudo_class('highlighted');
+    },
+
+    _stopPulsingIcon: function() {
+        // FIXME remove log
+        log('STOP PULSING ICON');
```
joaquimrocha: Remove the log.

js/ui/components/clubhouse.js
```diff
@@ -217,6 +221,16 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
     },
 
+    startPulsingIcon: function() {
+        this._openButton.get_children()[0].add_style_pseudo_class('highlighted');
```
joaquimrocha: I don't love that we are getting the con this way. Maybe creating a subclass of the button called `NotifiableButton` that gets a `pulsate` method is cleaner and easier to maintian.

js/ui/components/clubhouse.js
```diff
@@ -120,7 +121,10 @@ var ClubhouseButtonManager = new Lang.Class({
     _init: function() {
         this._openButton =
             new St.Button({ child: new St.Icon({ style_class: 'clubhouse-open-button-icon' }) });
-        this._openButton.connect('clicked', () => { this.emit('open-clubhouse'); })
+        this._openButton.connect('clicked', () => {
```
joaquimrocha: We may want to stop the pulsating right on `hover`, but this is just my impression. We may keep it like this for now.

js/ui/components/clubhouse.js
```diff
@@ -381,5 +395,15 @@ var ClubhouseComponent = new Lang.Class({
         if (this.proxy.g_name_owner && this._clubhouseSource)
             this._clubhouseSource.activateAction('stop-quest', null);
     },
+
+    _onPropertiesChanged: function(proxy, changedProps, invalidatedProps) {
```
joaquimrocha: See my comments in the Clubhouse's PR, I think we should not use a property for this.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,33 @@ var ClubhouseNotificationSource = new Lang.Class({
     },
 });
 
+var NotifiableButton = new Lang.Class({
+    Name: 'NotifiableButton',
```
joaquimrocha: I think it'd be fine to use this name and the overall approach if we were to have more than one button that's notifiable. Since that's not the case for now, I'd prefer to encapsulate things more and call it really something like `ClubhouseOpenButton`. See the rest of the comments which are related to this.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,33 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    Extends: St.Button,
+
+    _init: function(params) {
+        this.parent(params);
+
+        this.connect('clicked', () => {
+            this._stopPulsing();
+        })
```
joaquimrocha: Since this is the button implementation, we shouldn't connect to this button. Instead, we should stop the highlighting in a `clicked` implementation. However, I think that for now we should simply stop pulsating when the Clubhouse is to be opened (already handled in the `ClubhouseComponent`).

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,33 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    },
+
+    pulsate: function() {
+        this.get_children()[0].add_style_pseudo_class('highlighted');
```
joaquimrocha: We should be able to call this on the icon directly, instead of getting the first child which is not set by the button directly. So instead of allowing any child to be passed to this button, we should stop passing the icon and instead create and add it here (in the constructor), since this isto become the "open button" only and not a generic notifiable button.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,33 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    },
+
+    _stopPulsing: function() {
```
joaquimrocha: Make the function public so we call from outside, like the `pulsate` one. In fact, I'd simply create a function `setHighlighted(true|false)`, which simplifies the code.

js/ui/components/clubhouse.js
```diff
@@ -217,6 +239,10 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
     },
 
+    startPulsingIcon: function() {
```
joaquimrocha: This method has a generic name, yet it refers to the open button. So I'd call it `suggestOpen` or `setOpenButtonHighlight(true|false)`.

js/ui/components/clubhouse.js
manuq: Sounds good. `NotifiableButton` was suggested by you in the previous review pass.

manuq:
@joaquimrocha this is ready for another review. I amended the commit message to reflect that I wasn't able to add a pulsing animation.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,31 @@ var ClubhouseNotificationSource = new Lang.Class({
...
 
+var ClubhouseOpenButton = new Lang.Class({
+    Name: 'ClubhouseOpenButton',
+    Extends: St.Button,
+
+    _init: function(params) {
+        this.parent(params);
+
+        this._icon = new St.Icon({ style_class: 'clubhouse-open-button-icon' });
+        this.child = this._icon;
```
joaquimrocha: I think it's more appropriate to call `this.add_child(this._icon);` here.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,31 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+    },
+
+    setHighlighted: function(highlighted=true) {
```
joaquimrocha: Don't use a default value here. I think it's better to explicitely see `setHighlighted(true)` than an empty method call.

js/ui/components/clubhouse.js
```diff
@@ -114,12 +115,31 @@ var ClubhouseNotificationSource = new Lang.Class({
...
+        if (highlighted) {
+            this._icon.add_style_pseudo_class('highlighted');
+        } else {
```
joaquimrocha: Nitpick: Don't use curly brackets for one-line bodies.

js/ui/components/clubhouse.js
```diff
@@ -217,6 +237,10 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
     },
 
+    setOpenButtonHighlight: function(highlighted=true) {
```
joaquimrocha: Same comment about the highlighted's default value.

joaquimrocha:
Okay, I have tested it and it works. However, we need to not highlight the button is the Clubhouse is visible (otherwise, after closing it, the button will pulsate).

js/ui/components/clubhouse.js
manuq: Indeed, thanks.

js/ui/components/clubhouse.js
manuq: I think this is a matter of taste. But I will do as you say to move this PR forward.

js/ui/components/clubhouse.js
manuq: Yeah, it seems the Shell follows that style. So I'll adjust.

js/ui/components/clubhouse.js
manuq: I remember now why I did it this way. Changing this line to use `add_child` throws an error. Here is the traceback: 

```
== Stack trace for context 0x55d402f1b650 ==

(gnome-shell:6234): Gjs-CRITICAL **: Attempting to call back into JSAPI during the sweeping phase of GC. This is most likely caused by not destroying a Clutter actor or Gtk+ widget with ::destroy signals connected, but can also be caused by using the destroy(), dispose(), or remove() vfuncs. Because it would crash the application, it has been blocked and the JS callback not invoked.

(gnome-shell:6234): Gjs-CRITICAL **: The offending signal was destroy on Gjs_BaseIcon 0x55d406829cb0.
== Stack trace for context 0x55d402f1b650 ==

(gnome-shell:6234): Gjs-CRITICAL **: Attempting to call back into JSAPI during the sweeping phase of GC. This is most likely caused by not destroying a Clutter actor or Gtk+ widget with ::destroy signals connected, but can also be caused by using the destroy(), dispose(), or remove() vfuncs. Because it would crash the application, it has been blocked and the JS callback not invoked.

(gnome-shell:6234): Gjs-CRITICAL **: The offending signal was actor-removed on Gjs_IconGrid 0x55d4067be600.
== Stack trace for context 0x55d402f1b650 ==

(gnome-shell:6234): Gjs-CRITICAL **: Attempting to call back into JSAPI during the sweeping phase of GC. This is most likely caused by not destroying a Clutter actor or Gtk+ widget with ::destroy signals connected, but can also be caused by using the destroy(), dispose(), or remove() vfuncs. Because it would crash the application, it has been blocked and the JS callback not invoked.

(gnome-shell:6234): Gjs-CRITICAL **: The offending signal was destroy on StButton 0x55d406831620.
== Stack trace for context 0x55d402f1b650 ==
```

manuq:
Regarding the situation where the clubhouse is already open: I thought this signal will be emited by the clubhouse itself, so it already knows if its open or not to emit it. But yes, I can add the check in the Shell component too.

manuq:
@joaquimrocha this is ready for another review pass.

joaquimrocha:
> Regarding the situation where the clubhouse is already open: I thought this signal will be emited by the clubhouse itself, so it already knows if its open or not to emit it. But yes, I can add the check in the Shell component too.

Right. I thought about that at first too, but the fact that the Clubhouse is open doesn't mean it's visible (e.g. when it's open but the screen is locked, or the overview/grid is visible). So the Shell component it's a better way to set whether we need to highlight the button or not.

js/ui/components/clubhouse.js
joaquimrocha: I haven't seen this used before in the Shell so I thought it'd be better not to use it (besides the logic about `setHighlighted` hinting it's going to set something), but I may be mistaken.

manuq:
> > Regarding the situation where the clubhouse is already open: I thought this signal will be emited by the clubhouse itself, so it already knows if its open or not to emit it. But yes, I can add the check in the Shell component too.
> 
> Right. I thought about that at first too, but the fact that the Clubhouse is open doesn't mean it's visible (e.g. when it's open but the screen is locked, or the overview/grid is visible). So the Shell component it's a better way to set whether we need to highlight the button or not.

Oh I see now. Good point.

js/ui/components/clubhouse.js
manuq: I did grep the codebase prior to use it and I've found a few cases. But now I see they are more the exception than the rule.

js/ui/components/clubhouse.js
```diff
@@ -118,12 +119,30 @@ var ClubhouseNotificationSource = new Lang.Class({
...
 });
 
+var ClubhouseOpenButton = new Lang.Class({
+    Name: 'ClubhouseOpenButton',
+    Extends: St.Button,
+
+    _init: function(params) {
+        this.parent(params);
+
+        this._icon = new St.Icon({ style_class: 'clubhouse-open-button-icon' });
```
cosimoc: You could create the icon before calling `this.parent()` and then assign it to `params.child` to get it set during construction.

js/ui/components/clubhouse.js
```diff
@@ -291,6 +316,13 @@ var ClubhouseComponent = new Lang.Class({
             });
         }
 
+        if (this._clubhouseProxySuggestOpenHandler == 0) {
+            this._clubhouseProxySuggestOpenHandler = this.proxy.connectSignal('SuggestOpen', () => {
+                if (!Main.overview.visible)
```
cosimoc: I don't think this check here is needed. `setOpenButtonHighlight()` does not affect the visibility of the button, which is what we care about the overview visibility for.

js/ui/components/clubhouse.js
manuq: @cosimoc I tried but it broke my shell with:
```
== Stack trace for context 0x55db51a18650 ==

(gnome-shell:17187): Gjs-CRITICAL **: Attempting to call back into JSAPI during the sweeping phase of GC. This is most likely caused by not destroying a Clutter actor or Gtk+ widget with ::destroy signals connected, but can also be caused by using the destroy(), dispose(), or remove() vfuncs. Because it would crash the application, it has been blocked and the JS callback not invoked.

(gnome-shell:17187): Gjs-CRITICAL **: The offending signal was destroy on StButton 0x55db552f8280.
```

manuq:
@joaquimrocha the fixups look good for me. Feel free to merge it.

js/ui/components/clubhouse.js
cosimoc: @manuq That message does not look like one that would break your shell, and I'm pretty sure it happens regardless of these changes (I see tons of those warnings just running the shell normally)

PR #339, opened 2018-11-05
clubhouse: Don't destroy the different parts when the component is di…

joaquimrocha:
…sabled

The Clubhouse component gets disabled when e.g. the screen is locked and
it was destroying the different parts (buttons, notification, etc.) when
that occurred. This is a problem because when the user unlocks the
screen, the Clubhouse related notifications are supposed to still be
visible, as well as the Clubhouse open button's state (it will soon have
a highlighted state).

This patch ensures the desired UX by just hiding/showing the different
UI parts instead of destroying them.

https://phabricator.endlessm.com/T24283

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -137,6 +137,8 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
         this._clubhouseNotifyHandler = 0;
 
+        this._visible = true;
+
         this._updateVisibility();
 
         // If the Clubhouse is open and the Shell is restarted, then global.get_window_actors()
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -137,6 +137,8 @@ var ClubhouseButtonManager = new Lang.Class({
         this._clubhouseWindowActor = null;
         this._clubhouseNotifyHandler = 0;
 
+        this._visible = true;
```
cosimoc: I find it a bit confusing that this flag is called `this._visible`, because `this._visible = true` does not imply that the button is actually visible (there are the other conditions). I think the meaning of this option would be clearer if it was called `this._enabled`.

js/ui/components/clubhouse.js
```diff
@@ -378,12 +380,23 @@ var ClubhouseComponent = new Lang.Class({
             Main.layoutManager.addChrome(this._banner.actor);
             this._banner.reposition();
         }
+
+        // Sync the visibility here because the screen may be locked when a notification
+        // happens
+        this._syncVisibility();
```
cosimoc: I'm not sure I understand why this is here; if the screen is locked when a notification happens, we will have been disabled before. Upon unlocking, we will be re-enabled and the banner will become visible, if it was triggered. Am I missing something?

js/ui/components/clubhouse.js
cosimoc: On second thought, I am just going to merge this as-is.

PR #341, opened 2018-11-07
sounds: Add FlipToHack sounds

cfoch:
https://phabricator.endlessm.com/T24173

This depends on https://github.com/endlessm/hack-sound-server/pull/17

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -16,6 +16,7 @@ const AppActivation = imports.ui.appActivation;
 const Main = imports.ui.main;
 const Params = imports.misc.params;
 const Tweener = imports.ui.tweener;
+const SoundServer = imports.misc.soundServer;
 
 const WINDOW_ANIMATION_TIME = 0.25;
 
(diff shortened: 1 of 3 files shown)


ptomato:
Rob has requested that all PRs for gnome-shell be connected with a Phabricator ticket.

js/ui/codeView.js
```diff
@@ -201,6 +202,10 @@ var WindowTrackingButton = new Lang.Class({
         params = Params.parse(params, buttonParams, true);
 
         this.parent(params);
+
+        this.set_track_hover(true);
```
cosimoc: `track_hover` is already `true`, you should not need to reset this here.

js/ui/codeView.js
```diff
@@ -201,6 +202,10 @@ var WindowTrackingButton = new Lang.Class({
...
+        this.connect("notify::hover", this._onHoverChanged.bind(this));
```
cosimoc: Please use single quotes. (Also below.)

js/ui/codeView.js
```diff
@@ -265,6 +270,24 @@ var WindowTrackingButton = new Lang.Class({
     set state(value) {
         this.child.flipped = value == STATE_TOOLBOX;
     },
+
+    _onHoverChanged: function() {
```
ptomato: nitpick: Use the new method syntax: `_onHoverChanged() {` (here and below.)

cosimoc: I agree with @cfoch here... Let's merge it with the consistent syntax since this is an old file. We can change it to use the new syntax all at once in a single pass later on.

PR #343, opened 2018-11-07
T24365

joaquimrocha:
https://phabricator.endlessm.com/T24365


js/ui/components/clubhouse.js
```diff
@@ -55,6 +55,82 @@ const ClubhouseIface =
...
+        this._windowActor = null;
+    },
+
+    getWindowX: function() {
+        if (Main.overview.visible || this._windowActor == null)
+            return -1;
+        return this._windowActor.x;
+    },
+});
+Signals.addSignalMethods(ClubhouseWindowTracker.prototype);
```
GeorgesStavracas: Inherit from `GObject.Object`, add signals [like this](https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/js/ui/layout.js#L181), and remove this `Signals.addSignalMethods(...);` line.

js/ui/components/clubhouse.js
```diff
@@ -283,57 +277,12 @@ var ClubhouseButtonManager = new Lang.Class({
     },
 
     _updateVisibility: function() {
-        this._closeButton.visible = this._visible && !Main.overview.visible && this._clubhouseWindowActor;
+        this._closeButton.visible = this._visible && !Main.overview.visible &&
+            this._clubhouseWindowTracker.getWindowX() != -1;
```
GeorgesStavracas: Nitpick: alignment:

```javascript
        this._closeButton.visible = this._visible && !Main.overview.visible &&
                                    this._clubhouseWindowTracker.getWindowX() != -1;
```

js/ui/components/clubhouse.js
```diff
@@ -85,7 +169,15 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
-        this.actor.x = monitor.x + monitor.width - (this.actor.width + margin);
+
+        let clubhouseWindowX = this._clubhouseWindowTracker.getWindowX();
+
+        if (clubhouseWindowX == -1)
+            this.actor.x = monitor.x + monitor.width;
+        else
+            this.actor.x = clubhouseWindowX;
+
+        this.actor.x -=  this.actor.width + margin;
```
cosimoc: Double space after `=`

cosimoc:
Also needs a rebase.

js/ui/components/clubhouse.js
joaquimrocha: I had that but then I decided to use the `addSignalMethods` instead because I think it's more compact. Is there a reason why using this is less recommendable (so I understand what's better for next time)?

js/ui/components/clubhouse.js
```diff
@@ -55,6 +55,82 @@ const ClubhouseIface =
   </interface> \
 </node>';
 
+var ClubhouseWindowTracker = new Lang.Class({
+    Name: 'ClubhouseWindowTracker',
+
+    _init: function() {
+        this.parent();
```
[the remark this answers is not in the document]
joaquimrocha: I will keep it since @GeorgesStavracas has asked for it to be a GObject. I'd rather skip the extra code if using `addSignalMethods` works fine, but if this is more appropriate in the Shell, then I will do it.

joaquimrocha:
@cosimoc / @GeorgesStavracas , I have addressed your comments and rebased the PR. Please look at the session mode changes (moved into the `ClubhouseWindowTracker`) which I changed when rebasing. I haven't tested the FBE but I think it should all be working fine.

manuq:
Tested, is working nicely.

js/ui/components/clubhouse.js
```diff
@@ -55,6 +55,85 @@ const ClubhouseIface =
...
+
+        this._windowActor = null;
+
+        // If the Clubhouse is open and the Shell is restarted, then global.get_window_actors()
+        // will still not have any contents when components are initialized, and there's no "map"
+        // signal emitted from the Window Manager either. So instead of relying on the map signal
+        // (which would allow us to track windows individually instead of always having to check the
+        // current list of windows) we connect to the screen's "restacked" signal.
+        this._restackedHandler = global.screen.connect('restacked',
```
cosimoc: `this._restackedHandler` is never used

js/ui/components/clubhouse.js
```diff
@@ -77,6 +156,14 @@ var ClubhouseNotificationBanner = new Lang.Class({
 
         // Always wrap the body's text
         this._expandedLabel.actor.add_style_class_name('clubhouse-quest-view-label');
+
+        this._clubhouseWindowTracker = new ClubhouseWindowTracker();
```
cosimoc: I'm concerned that we create one of these for each banner (plus one for the button manager), but they will never get destroyed. There are a few problems with that:
- these objects are not cheap, as they react to frequent signals (`restacked`) and iterate over all the windows.
- I know that the banner disconnects the signal when it's destroyed, but each tracker object itself will stay alive and keep all of its connections/signals alive and running. That may be avoidable by setting `this._clubhouseWindowTracker = null` after disconnecting the signals. It should get garbage-collected.
- If you do the above and manage to destroy the tracker, the callbacks from the overview and sessionMode will still be connected, since those objects outlive the tracker and nothing disconnects the signals when the tracker is destroyed. I think that at present, those callbacks would crash the app in that situation, since the code would still try to call `emit()` on an instance that is already destroyed.

I am open to either solving this by creating a single tracker object, owned by the clubhouse component, that ends up being shared by all the other objects that need it (my preferred solution), or by making sure to destroy the trackers with each banner, and making the tracker robust to that case (my less preferred solution because of the performance/resources concern.)

PR #344, opened 2018-11-08
clubhouse: Paginate text by paragraph in the quest view banner

joaquimrocha:
We need to make sure that the text displayed in the quest view banner
is split to fit in the expected UI (because some quest messages have a
long text to be displayed).

Even though we may need a more advanced technique to make this text
split, for now we're just splitting the text in parapraphs.

To advance through the text, we include a "next page" button, and only
add any eventual notification actions when the last page is displayed.

https://phabricator.endlessm.com/T24219

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -61,7 +61,19 @@ var ClubhouseNotificationBanner = new Lang.Class({
     Extends: MessageTray.NotificationBanner,
 
     _init: function(notification) {
+        this._textPages = [];
+        this._textIdx = 0;
+        this._splitTextInPages(notification.bannerBodyText);
+
+        // We will show the text differently, so have the parent set no text for now
+        notification.bannerBodyText = '';
+
         this.parent(notification);
+
+        this._setNextPage();
+
+        if (this._textPages.length > 1)
+            this._setupNextPageButton();
 
         // We don't have an "unexpanded" state for now
         this.expand(false);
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -79,6 +91,19 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
 
+    _addActions: function() {
+        // Only set up the actions if we're showing the last page of text
+        if (!this._inLastPage())
+            return;
+
+        this.parent();
+    },
+
+    splitTextInPages: function(fulltext) {
```
cosimoc: Should be "private" (i.e. have a leading underscore)

joaquimrocha:
Hey @cosimoc ,

> This looks good to me in general, but I was surprised (not negatively, just surprised) to see this as a shell PR rather than a clubhouse PR, because I had imagined that's where the text would be broken down into paragraphs.

Maybe it wasn't fully clear from my comment in the Phab task.
I think that paginating the text is a "view" matter, rather than a "model/control" matter. i.e. the Quest needs to display this much text, and it's up to the view to ensure that's done.

I also believe having it in the Clubhouse, even if abstracting it as much as possible, would be messier because the usual flow is "show this much text, and when there's a reply, go to the next step", and dividing the text in several messages for the Shell while keeping this flow transparent to the quest can be tricky.

Hope this clarifies my reasoning.

PR #348, opened 2018-11-09
T24414

cosimoc:
https://phabricator.endlessm.com/T24414

### DIFF
--- a/src/shell-code-view-effect.c
+++ b/src/shell-code-view-effect.c
@@ -248,27 +248,41 @@ shell_code_view_effect_init (ShellCodeViewEffect *self)
 
   priv->pipeline = cogl_pipeline_copy (klass->base_pipeline);
 
-  /* FIXME: this should not be hardcoded here, and be set ideally from
-   * the CSS theme.
-   */
-  clutter_color_from_string (&priv->gradient_colors[0], "#05213f");
-  clutter_color_from_string (&priv->gradient_colors[1], "#031c39");
-  clutter_color_from_string (&priv->gradient_colors[2], "#00275c");
-  clutter_color_from_string (&priv->gradient_colors[3], "#8d6531");
-  clutter_color_from_string (&priv->gradient_colors[4], "#f4f1a2");
-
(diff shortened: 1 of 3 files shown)


src/shell-code-view-effect.c
```diff
@@ -271,16 +256,55 @@ shell_code_view_effect_init (ShellCodeViewEffect *self)
...
+ * @gradient_len: length of gradient stops
  *
  * Creates a new #ShellCodeViewEffect to be used with
  * clutter_actor_add_effect()
  *
  * Return value: the newly created #ShellCodeViewEffect or %NULL
  */
 ClutterEffect *
-shell_code_view_effect_new (void)
+shell_code_view_effect_new (gchar **gradient_colors,
```
ptomato: This function doesn't seem to be used. In any case, I would not recommend setting the gradient stops in the `new()` function as it can cause surprises when creating the class from JS. I think it's clearer for calling code to just set the gradient stops after constructing the object, as that is what happens in JS anyway. (I wrote a [blog post](https://ptomato.wordpress.com/2018/01/28/geek-tip-g_object_new-and-constructors/) on this topic.)

src/shell-code-view-effect.c
```diff
@@ -271,16 +256,55 @@ shell_code_view_effect_init (ShellCodeViewEffect *self)
...
+{
+  ShellCodeViewEffectPrivate *priv = shell_code_view_effect_get_instance_private (effect);
+  gint i;
+
+  g_return_if_fail (gradient_colors != NULL);
+  g_return_if_fail (gradient_points != NULL);
+  g_return_if_fail (gradient_len == 5);
+
+  memcpy (priv->gradient_points, gradient_points, sizeof (gfloat) * gradient_len);
+  for (i = 0; i < 5; i++)
```
ptomato: Small difference, but everything else in this function is parametrized in `gradient_len` even if it can currently only be 5, so this should probably be `i < gradient_len`.

cosimoc:
@ptomato Thanks, updated.

PR #349, opened 2018-11-12
T24401

joaquimrocha:
https://phabricator.endlessm.com/T24401

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -569,6 +569,22 @@ var ClubhouseComponent = new Lang.Class({
             oldAddNotificationFunc.apply(Main.notificationDaemon._gtkNotificationDaemon,
                                          [params, invocation]);
         }
+
+        let oldRemoveNotificationFunc = GtkNotificationDaemon.prototype.RemoveNotificationAsync;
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -567,9 +566,25 @@ var ClubhouseComponent = new Lang.Class({
...
             }
 
-            this._oldAddNotificationFunc.apply(Main.notificationDaemon._gtkNotificationDaemon,
-                                               [params, invocation]);
+            oldAddNotificationFunc.apply(Main.notificationDaemon._gtkNotificationDaemon,
+                                         [params, invocation]);
         }
+
+        let oldRemoveNotificationFunc = GtkNotificationDaemon.prototype.RemoveNotificationAsync;
+        GtkNotificationDaemon.prototype.RemoveNotificationAsync = (params, invocation) => {
```
ptomato: Drive-by comment: This is a pre-existing problem and probably not an issue here because you have a singleton, but when you are defining methods directly on a prototype it's better to use `function` rather than arrow functions. That's because the arrow function binds the current value of `this` to the function. So if you were to call `otherDaemon.RemoveNotificationAsync()` it would not call the method on `otherDaemon`!

PR #350, opened 2018-11-14
T24438

cosimoc:
https://phabricator.endlessm.com/T24438


cosimoc:
The real fix is the last commit, the others are just upstream backports.

PR #351, opened 2018-11-14
clubhouse: Ensure the Clubhouse window is focused when shown

joaquimrocha:
The Clubhouse window has a DOCK hint type and that apparently prevents
it from geting focused most of the time due to how Mutter works [1].

To work around that, we directly activate the window from the Clubhouse
Shell component, hence making it focused.

This is not a great solution, because if the window gets unfocused, then
it won't be focused back until the window is hidden and shown again.
However, this is not a problem for now because the Clubhouse window is
supposed to hidden automatically when unfocused anyway.

An alternative solution from the Clubhouse's side, i.e. making it a
normal window and making it behave like a DOCK one (always on top, no
decoration, etc.) has its caveats because it's easy to still drag it
around (and apparently there's no way to avoid that from the application
side).

[1] https://github.com/endlessm/mutter/blob/master/src/core/window.c#L2232

https://phabricator.endlessm.com/T24075

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -111,6 +111,10 @@ var ClubhouseWindowTracker = new Lang.Class({
 
         this._windowActor = actor;
 
+        // We need to activate the window directly as otherwise it won't be focused because
+        // of having a DOCK hint type
+        actor.meta_window.activate(global.get_current_time());
+
         this.emit('window-changed');
 
         // Track Clubhouse's window actor to make the closeButton always show on top of


js/ui/components/clubhouse.js
```diff
@@ -111,6 +111,10 @@ var ClubhouseWindowTracker = new Lang.Class({
 
         this._windowActor = actor;
 
+        // We need to activate the window directly as otherwise it won't be focused because
+        // of having a DOCK hint type
+        actor.meta_window.activate(global.get_current_time());
```
cosimoc: Calling `activate()` on the side component window sounds like a good idea, but not every time this callback is received... I am not sure under which circumstances the `restacked` signal is received, and always activating a window in response to what is a signal emitted potentially when another window is activated does not feel like a robust solution to me. (Even though I understand that ill side effects are protected by the `this._windowActor == actor` check above at the moment.)

Since I offered an easy way to solve the dragging issue from the app side in the ticket, I'd like to explore that avenue further before committing this workaround.

joaquimrocha:
We'll go with the alternative of making the Clubhouse window use a "normal" type-hint and extra measures to make it work like a dock one.

PR #352, opened 2018-11-15
ShellCodeViewEffect: ignore alpha areas for shader

cosimoc:
Our gradient map algorithm does not work well on areas with an alpha
channel; in particular, we are interested in excluding the invisible
borders around the window from the shader.

This commit disables the shader where an alpha channel is in use in
the app window.

https://phabricator.endlessm.com/T24193


ptomato:
I don't really know GLSL so I can't give a good review of this. But the approach seems correct.

cosimoc:
Merging this to get it in for Beta 2.

PR #353, opened 2018-11-15
T24480

cosimoc:
https://phabricator.endlessm.com/T24480


cosimoc:
@ptomato Updated

ptomato:
I think you amended the wrong commit, but the resulting code looks good to me.

cosimoc:
Whoops, yes... re-amended the right commit :)

PR #354, opened 2018-11-15
codeView: Add hover loops for flip-to-hack button

ptomato:
Same scheme as in the codeview to gracefully handle a sound being
stopped before the UUID arrives.

https://phabricator.endlessm.com/T24523

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -275,13 +275,50 @@ var WindowTrackingButton = new Lang.Class({
         this.child.flipped = value == STATE_TOOLBOX;
     },
 
+    _startHoverSound(flipped) {
+        if (this._hoverSoundID === 'pending')
+            return;
+        if (this._hoverSoundID === 'cancel') {
+            // Hovered in and out and back in quickly, before the first UUID was
+            // returned. In this case, un-cancel the original sound but don't
+            // request another one.
+            this._hoverSoundID = 'pending';
+            return;
+        }
+        this._hoverSoundID = 'pending';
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -275,13 +275,42 @@ var WindowTrackingButton = new Lang.Class({
         this.child.flipped = value == STATE_TOOLBOX;
     },
 
+    _startHoverSound(flipped) {
+        this._hoverSoundID = 'pending';
+        const id = flipped ? 'flip-inverse' : 'flip';
+        SoundServer.getDefault().playAsync(`shell/tracking-button/${id}/hover`)
+        .then(uuid => {
+            print('received uuid');
```
cosimoc: Remove this

cfoch:
I was implementing this, too. I had the problem that if I passed the mouse very quick over the arrow button, a loop kept playing even when the mouse was out of the button. The same problem occurs here.

cfoch:
It seems that the calls to the hover signal handler is faster than the response of the uuid from the server.

cosimoc:
@cfergeau That's what the `pending`/`cancel` status is supposed to catch in this PR as far as I can see.

ptomato:
It's certainly possible that it doesn't yet work right!

cosimoc:
@ptomato I can reproduce the issue that @cfoch mentioned here easily by hovering on and off the flip button in quick succession.
I think what is happening is that the callback returns the UUID after the second hover already changed the state back to `pending`, so the first sound never gets cancelled.

PR #355, opened 2018-11-16
codeView: Implement flip-to-hack blacklist

ptomato:
This looks in the location where com.endlessm.HackComponents should
install the hack-data.ini file. If the file isn't there, or doesn't
contain the relevant key, then nothing is blacklisted. This is
important, so that things still work in case the HackComponents flatpak
is uninstalled.

If the file does contain the key, then the app ID (whether from Flatpak
or an installed .desktop file) is checked against the blacklist, and if
it is present in the blacklist, then the CodeViewManager should not
handle mapping that window.

https://phabricator.endlessm.com/T24416

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -106,6 +106,56 @@ function _getWindowId(win) {
     return ('window:%d').format(win.get_stable_sequence());
 };
 
+const _ensureHackDataFile = (function () {
+    let keyfile = new GLib.KeyFile();
+    let initialized = false;
+
+    const flatpakInstallationPaths = [
+        GLib.build_filenamev([GLib.get_home_dir(), '.local/share/flatpak']),
+        '/var/lib/flatpak',
+    ];
+    const flatpakPath = 'app/com.endlessm.HackComponents/current/active/files';
+    const fileRelPath = 'share/hack-components';
+    const searchPaths = flatpakInstallationPaths.map(installation =>
+        GLib.build_filenamev([installation, flatpakPath, fileRelPath]));
+
+    return function() {
+        if (initialized)
+            return keyfile;
+
+        try {
+            keyfile.load_from_dirs('hack-data.ini', searchPaths,
+                GLib.KeyFileFlags.NONE);
+        } catch (err) {
+            if (!err.matches(GLib.FileError, GLib.FileError.NOENT))
+                logError(err, 'Error reading hack data file');
+            keyfile = null;
+        }
+
+        initialized = true;
+        return keyfile;
+    }
+}());
+
+function _isBlacklistedApp(desktop_id) {
+    const keyfile = _ensureHackDataFile();
+    if (keyfile === null)
+        return false;
+
+    const app_id = desktop_id.slice(0, -8);  // remove ".desktop"
+    let blacklist;
+    try {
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -106,6 +106,47 @@ function _getWindowId(win) {
     return ('window:%d').format(win.get_stable_sequence());
 };
 
+const _ensureHackDataFile = (function () {
+    let keyfile = new GLib.KeyFile();
+    let initialized = false;
+
+    const flatpakPath = '/var/lib/flatpak';
+    const fileRelPath = 'share/hack-components/hack-data.ini';
+    const filePath = `${flatpakPath}/app/com.endlessm.HackComponents/current/active/files/${fileRelPath}`;
```
cosimoc: Please use `g_build_filenamev()` or similar to build this

js/ui/codeView.js
```diff
@@ -106,6 +106,47 @@ function _getWindowId(win) {
...
+
+    return function() {
+        if (initialized)
+            return keyfile;
+
+        try {
+            keyfile.load_from_file(filePath, GLib.KeyFileFlags.NONE);
+        } catch (err) {
+            logError(err, 'Error reading hack data file');
```
cosimoc: We should catch the case here where the file does not exist, and avoid printing an error in that case

js/ui/codeView.js
```diff
@@ -106,6 +106,47 @@ function _getWindowId(win) {
     return ('window:%d').format(win.get_stable_sequence());
 };
 
+const _ensureHackDataFile = (function () {
+    let keyfile = new GLib.KeyFile();
+    let initialized = false;
+
+    const flatpakPath = '/var/lib/flatpak';
```
cosimoc: I'd like it if we could also support (with higher priority) this flaptak in the home directory flatpak installation.

js/ui/codeView.js
```diff
@@ -106,6 +106,47 @@ function _getWindowId(win) {
...
+    const keyfile = _ensureHackDataFile();
+    if (keyfile === null)
+        return false;
+
+    const app_id = desktop_id.slice(0, -8);  // remove ".desktop"
+    let blacklist;
+    try {
+        [blacklist] = keyfile.get_string_list('flip-to-hack', 'blacklist');
+    } catch (err) {
+        logError(err, 'No blacklist in hack data file');
```
cosimoc: I don't think we should log an error if the group name or key name are missing, since it could just be that we removed the blacklist but have other uses for this file. (Or alternatively, look up this information from a file that explicitly only has the blacklist in it, instead of a more general "hack data" keyfile.)

js/ui/codeView.js
ptomato: That's possible, but we also don't do that for any of the other hack-components...

js/ui/codeView.js
```diff
@@ -106,6 +106,56 @@ function _getWindowId(win) {
     return ('window:%d').format(win.get_stable_sequence());
 };
 
+const _ensureHackDataFile = (function () {
+    let keyfile = new GLib.KeyFile();
+    let initialized = false;
+
+    const flatpakInstallationPaths = [
```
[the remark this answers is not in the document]
ramcq: Ohhh.... hmm... fair enough. One day, we will put apps somewhere else and this might break though. :)

PR #356, opened 2018-11-16
codeView: improve placement of FlipToHack button

cosimoc:
This does not actually change the position of the button, but at the
moment we set a large left padding, which makes it so  the button is
clickable for a large portion immediately left of the window.

Instead, position it flush on the window edge, and add a small left
padding.

https://phabricator.endlessm.com/T24407

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -141,7 +141,7 @@ function _synchronizeMetaWindowActorGeometries(src, dst) {
 }
 
 function _synchronizeViewSourceButtonToRectCorner(button, rect) {
-    button.set_position(rect.x - button.width / 2,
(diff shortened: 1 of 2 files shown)


js/ui/codeView.js
```diff
@@ -141,7 +141,7 @@ function _synchronizeMetaWindowActorGeometries(src, dst) {
 }
 
 function _synchronizeViewSourceButtonToRectCorner(button, rect) {
-    button.set_position(rect.x - button.width / 2,
+    button.set_position(rect.x,
```
ptomato: Oh! This is probably why the animation didn't work, back when I tried to animate the button...

PR #363, opened 2018-11-20
T24445

joaquimrocha:
https://phabricator.endlessm.com/T24445


js/ui/components/clubhouse.js
```diff
@@ -279,9 +297,11 @@ var ClubhouseQuestBanner = new Lang.Class({
     Name: 'ClubhouseQuestBanner',
     Extends: ClubhouseNotificationBanner,
 
-    _init: function(notification) {
+    _init: function(notification, isFirstBanner) {
```
cosimoc: It would be clearer to call this `shouldSlideIn` instead of `isFirstBanner` -- the banner class itself does not need to know whether it was the first banner or not, only if it should slide.

js/ui/components/clubhouse.js
```diff
@@ -238,7 +245,18 @@ var ClubhouseNotificationBanner = new Lang.Class({
         else
             this.actor.x = clubhouseWindowX;
 
-        this.actor.x -= this.actor.width + margin;
+        let endX = this.actor.x;
+        endX -= this.actor.width + margin;
```
cosimoc: You can fold this into the previous line

js/ui/components/clubhouse.js
```diff
@@ -290,6 +290,33 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+            return;
+
+        let endX = monitor.x + monitor.width;
+
+        Tweener.addTween(this.actor,
+                         { x: endX,
+                           time: CLUBHOUSE_BANNER_ANIMATION_TIME,
+                           transition: 'easeOutQuad',
+                           onComplete: () => {
+                               this.actor.destroy();
```
cosimoc: Perhaps call `this.destroy(false)` here? Otherwise set `this.actor = null` as well for consistency.

js/ui/components/clubhouse.js
joaquimrocha: I did this on purpose so we associated the first-banner -> slide-in inside the banner itself. But I can change it.

js/ui/components/clubhouse.js
joaquimrocha: True. Will do it.

js/ui/components/clubhouse.js
joaquimrocha: Hmm, `this.destroy` is the function that calls this. That's inviting unwanted/erroneous recursion down the road.
I should however let the actor be null as well in the `this.destroy` function even when calling the slideOut.

js/ui/components/clubhouse.js
joaquimrocha: I decided to keep it in the end. It makes sense in my mind to make the distinction as it can also be better in the future if we need to change this slide in effect by something else.

joaquimrocha:
I have updated the PR. @GeorgesStavracas , since Cosimo is out for the week (and we need this) can you add your review/merge?

js/ui/components/clubhouse.js
```diff
@@ -238,7 +245,17 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+        let endX = this.actor.x - this.actor.width - margin;
+
+        if (this._shouldSlideIn)
+            Tweener.addTween(this.actor,
+                             { x: endX,
+                               time: CLUBHOUSE_BANNER_ANIMATION_TIME,
+                               transition: 'easeOutQuad'
+                             });
+        else
+            this.actor.x = endX;
```
GeorgesStavracas: Nitpick: use curly braces

js/ui/components/clubhouse.js
```diff
@@ -36,11 +36,15 @@ const Mainloop = imports.mainloop;
...
 const SideComponent = imports.ui.sideComponent;
+const Tweener = imports.ui.tweener;
 
 const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
 
 const CLUBHOUSE_BANNER_TIMEOUT_MSEC = 3000;
 
+const CLUBHOUSE_BANNER_ANIMATION_TIME = 0.2;
+
+
```
GeorgesStavracas: Nitpick: remove one of the newlines

js/ui/components/clubhouse.js
```diff
@@ -36,11 +36,15 @@ const Mainloop = imports.mainloop;
 const MessageTray = imports.ui.messageTray;
 const NotificationDaemon = imports.ui.notificationDaemon;
 const SideComponent = imports.ui.sideComponent;
+const Tweener = imports.ui.tweener;
 
 const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
 
 const CLUBHOUSE_BANNER_TIMEOUT_MSEC = 3000;
 
+const CLUBHOUSE_BANNER_ANIMATION_TIME = 0.2;
```
GeorgesStavracas: Nitpick: move to line above

js/ui/components/clubhouse.js
```diff
@@ -289,6 +289,34 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+                           time: CLUBHOUSE_BANNER_ANIMATION_TIME,
+                           transition: 'easeOutQuad',
+                           onComplete: () => {
+                               this.actor.destroy();
+                               this.actor = null;
+                           }
+                         });
+    },
+
+    destroy: function(shouldSlideOut) {
```
GeorgesStavracas: I disagree with the semantics of this function. After calling `destroy()`, it is expected that the object is destroyed and thus, any further access to it is invalid. By overriding it, it's effectively breaking this expectation.

I'd suggest, then, renaming `_slideOut()` to `slideOutAndDestroy()`, and removing the `shouldSlideOut` argument from this function.

js/ui/components/clubhouse.js
joaquimrocha: I thought one-lines didn't need curly braces. But I can do that...

js/ui/components/clubhouse.js
joaquimrocha: I disagree:
1) This is not a widget, calling it `destroy` has no special meaning as in the behavior you mentioned (which is for widgets only); it does call destroy on its actor but after that it won't access it (even sets it as `null`);
2) Using `slideOutAndDestroy` doesn't do the same, as there's a reason why I use the `shouldSlideOut` arg (sometimes it's supposed to slide-out, others it isn't). And branching this from the caller makes the code less readable.

I hope you understand my point.

joaquimrocha:
@GeorgesStavracas , I have updated the PR. Please see my reply to your comment about destroy too. I think it should be good to go now.

js/ui/components/clubhouse.js
joaquimrocha: That's a better explanation. I will update it. Thanks.

joaquimrocha:
Updated the PR again after @GeorgesStavracas 's clarification.

PR #364, opened 2018-11-20
paygManager: Understand ExpiryTime in terms of CLOCK_BOOTTIME

mwleeds:
The "ExpiryTime" property of the EpgProvider interface is changing to be
in terms of CLOCK_BOOTTIME rather than wall clock time, this commit
changes the paygManager class to account for this.

https://phabricator.endlessm.com/T24300

### DIFF
--- a/src/shell-util.c
+++ b/src/shell-util.c
@@ -814,4 +814,3 @@ shell_util_get_boottime (void)
 
   return (((gint64) ts.tv_sec) * 1000000) + (ts.tv_nsec / 1000);
 }
-
--- a/src/shell-util.c
+++ b/src/shell-util.c
@@ -790,3 +790,28 @@ shell_util_get_eos_image_version (void)
 
   return image_version;
 }
+
+/**
+ * shell_util_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses `CLOCK_BOOTTIME`, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ */
+gint64
+shell_util_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
+  if (G_UNLIKELY (result != 0))
+    g_error ("clock_gettime (CLOCK_BOOTTIME) failed: %s",
+             g_strerror (errno));
+
+  return (((gint64) ts.tv_sec) * 1000000) + (ts.tv_nsec / 1000);
+}
+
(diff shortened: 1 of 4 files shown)


mwleeds:
This shouldn't be merged until after https://github.com/endlessm/eos-payg/pull/28

src/shell-util.c
```diff
@@ -776,3 +776,28 @@ shell_util_get_eos_image_version (void)
...
+
+/**
+ * shell_util_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses CLOCK_BOOTTIME, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ **/
```
pwithnall:
```suggestion
 */
```

src/shell-util.c
```diff
@@ -776,3 +776,28 @@ shell_util_get_eos_image_version (void)
 
   return image_version;
 }
+
+/**
+ * shell_util_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses CLOCK_BOOTTIME, hence the name, but is not guaranteed to be
```
pwithnall:
```suggestion
 * suspended. Uses `CLOCK_BOOTTIME`, hence the name, but is not guaranteed to be
```

src/shell-util.c
```diff
@@ -776,3 +776,28 @@ shell_util_get_eos_image_version (void)
...
+ **/
+gint64
+shell_util_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
+  if G_UNLIKELY (result != 0)
```
pwithnall:
```suggestion
  if (G_UNLIKELY (result != 0))
```

mwleeds:
You're right, it would be more correct to get the clock from the provider and call `get_time()` on it than to encode the assumption that CLOCK_BOOTTIME is used into gnome-shell. But I think that would mean changing the D-Bus interface, so probably not worth it. We probably won't need to make a change like this again.

mwleeds:
> Looks good to me, although there’s a remaining GLib.get_real_time() call in js/ui/paygUnlockDialog.js. I think that needs to be replaced with EpgClock usage if the call here is also changed. Although probably every call to g_get_real_time() outside the EpgClock implementation should be changed.

I addressed this with a commit

mwleeds:
Merging this now that https://github.com/endlessm/eos-payg/pull/28 has been merged

mwleeds:
Oops, forgot to squash the fixup. Sorry

PR #368, opened 2018-11-22
T24459

joaquimrocha:
https://phabricator.endlessm.com/T24459

### DIFF
--- a/js/ui/animation.js
+++ b/js/ui/animation.js
@@ -16,6 +16,7 @@ var Animation = new Lang.Class({
         this.actor = new St.Bin();
         this.actor.connect('destroy', Lang.bind(this, this._onDestroy));
         this._speed = speed;
+        this._frameTimeouts = [this._speed];
 
         this._isLoaded = false;
         this._isPlaying = false;
@@ -30,11 +31,15 @@ var Animation = new Lang.Class({
 
     play: function() {
         if (this._isLoaded && this._timeoutId == 0) {
+            // Set the frame to be the previous one, so when we update it
+            // when play is called, it shows the current frame instead of
+            // the next one.
             if (this._frame == 0)
-                this._showFrame(0);
+                this._frame = this._animations.get_n_children() - 1;
+            else
+                this._frame -= 1;
 
-            this._timeoutId = GLib.timeout_add(GLib.PRIORITY_LOW, this._speed, Lang.bind(this, this._update));
-            GLib.Source.set_name_by_id(this._timeoutId, '[gnome-shell] this._update');
+            this._update();
         }
 
         this._isPlaying = true;
@@ -63,7 +68,16 @@ var Animation = new Lang.Class({
 
     _update: function() {
         this._showFrame(this._frame + 1);
-        return GLib.SOURCE_CONTINUE;
+
+        // Show the next frame after the timeout of the current one
+        let currentFrameTimeoutIndex = this._frame % this._frameTimeouts.length;
+        this._timeoutId = GLib.timeout_add(GLib.PRIORITY_LOW,
+                                           this._frameTimeouts[currentFrameTimeoutIndex],
+                                           Lang.bind(this, this._update));
+
+        GLib.Source.set_name_by_id(this._timeoutId, '[gnome-shell] this._update');
+
+        return GLib.SOURCE_REMOVE;
     },
 
     _animationsLoaded: function() {
@@ -75,6 +89,21 @@ var Animation = new Lang.Class({
 
     _onDestroy: function() {
         this.stop();
+    },
+
+    setFrameTimeouts: function(timeoutList) {
+        if (timeoutList.length == 0)
+            throw new Error('Cannot set an empty list as the frame-timeouts for animation!');
+
+        let wasPlaying = this._isPlaying;
+        this.stop();
+
+        this._frameTimeouts = timeoutList;
+
+        // If the animation was playing, we continue to play it here (where it will
+        // use the new timeouts)
+        if (wasPlaying)
+            this.play();
     }
 });
 
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -311,8 +311,6 @@ var ClubhouseNotificationBanner = new Lang.Class({
         this.actor.add_style_class_name('clubhouse-notification');
         this._iconBin.add_style_class_name('clubhouse-notification-icon-bin');
         this._closeButton.add_style_class_name('clubhouse-notification-close-button');
-        let image = this._iconBin.get_first_child();
-        image.add_style_class_name('clubhouse-notification-image');
 
         // Always wrap the body's text
         this._expandedLabel.actor.add_style_class_name('clubhouse-notification-label');
@@ -330,6 +328,11 @@ var ClubhouseNotificationBanner = new Lang.Class({
             getClubhouseWindowTracker().disconnect(this._clubhouseTrackerHandler);
             this._clubhouseTrackerHandler = 0;
         }
+    },
+
+    setIcon: function(actor) {
+        actor.add_style_class_name('clubhouse-notification-image');
+        this.parent(actor);
     },
 
     _rearrangeElements: function() {
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -155,28 +155,13 @@ var ClubhouseWindowTracker = new Lang.Class({
     },
 });
 
-var getClubhouseAnimator = (function () {
-    let clubhouseAnimator;
-    return function() {
-        if (!clubhouseAnimator)
-            clubhouseAnimator = new ClubhouseAnimator();
-        return clubhouseAnimator;
-    };
-}());
-
 var ClubhouseAnimator = new Lang.Class({
     Name: 'ClubhouseAnimator',
-    Extends: GObject.Object,
 
-    _init: function() {
-        this.parent();
-        this._proxy = null;
+    _init: function(proxy) {
+        this._proxy = proxy;
         this._animations = {};
         this._clubhousePaths = this._getClubhousePaths();
-    },
-
-    setClubhouseProxy: function(proxy) {
-        this._proxy = proxy;
     },
 
     _getClubhousePaths: function() {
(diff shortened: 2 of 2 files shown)


manuq:
@joaquimrocha this branch has a conflict with master.

js/ui/components/clubhouse.js
```diff
@@ -53,6 +54,10 @@ const ClubhouseIface =
     <method name="hide"> \
       <arg type="u" direction="in" name="timestamp"/> \
     </method> \
+    <method name="getAnimationMetadata"> \
```
manuq: This must have a counterpart in the clubhouse, but I don't see a PR opened for the clubhouse repo.

joaquimrocha: This is not yet CR Ready.

js/ui/components/clubhouse.js
```diff
@@ -148,6 +153,111 @@ var ClubhouseWindowTracker = new Lang.Class({
...
+                log('Error parsing animation metadata json: ' + e);
+                return;
+            }
+
+            this._animations[uri] = json_data;
+            this.emit('animation-ready');
+        });
+    },
+
+    getAnimation: function(uri) {
```
cosimoc: Expanding from my comment above, I'd like this object to only have one API entry point; it's confusing that `getAnimationByUri()` is an async method that ends up emitting a signal, but `getAnimation()` returns the `Animation` object that is ultimately what's needed.

js/ui/components/clubhouse.js
```diff
@@ -148,6 +153,111 @@ var ClubhouseWindowTracker = new Lang.Class({
...
+
+        return splitDir.join('/');
+    },
+
+    _getClubhouseUri: function(uri) {
+        // Discard the /app/ prefix
+        let uriSuffix = uri.replace(/^\/app\//g, '');
+
+        for (let path of this._clubhousePaths) {
+            let completeUri = path + '/files/' + uriSuffix;
```
cosimoc: This is a path, not an uri... Also consider using e.g. `GLib.build_filenamev()` to build the path.

js/ui/components/clubhouse.js
```diff
@@ -534,6 +676,9 @@ var ClubhouseComponent = new Lang.Class({
                     this._clearQuestBanner();
                 }
             });
+
+            let clubhouseAnimator = getClubhouseAnimator();
+            clubhouseAnimator.setClubhouseProxy(this.proxy);
```
cosimoc: It's clear that the proxy is really a construct parameter for the animator... I understand that it's a singleton, but instead of pretending that it does not matter who calls `setClubhouseProxy()` I think I would find it cleaner if we had an e.g. `createClubhouseAnimator()` method that explicitly takes the proxy as a construct parameter, which would only be called by the component.

js/ui/animation.js
```diff
@@ -75,6 +84,23 @@ var Animation = new Lang.Class({
 
     _onDestroy: function() {
         this.stop();
+    },
+
+    setFrameTimeouts: function(timeoutList) {
+        if (timeoutList.length == 0) {
+            log('Cannot set an empty list as the frame-timeouts for animation!');
```
cosimoc: This should throw an error.

js/ui/animation.js
```diff
@@ -63,7 +63,16 @@ var Animation = new Lang.Class({
 
     _update: function() {
         this._showFrame(this._frame + 1);
-        return GLib.SOURCE_CONTINUE;
+
+        // Show the next frame after the timeout of the current one
+        let currentFrameTimeout = this._frame % this._frameTimeouts.length;
```
cosimoc: This variable is an index, not the timeout itself

js/ui/animation.js
```diff
@@ -75,6 +84,23 @@ var Animation = new Lang.Class({
 
     _onDestroy: function() {
         this.stop();
+    },
+
+    setFrameTimeouts: function(timeoutList) {
+        if (timeoutList.length == 0) {
```
cosimoc: Perhaps the other thing to check for is that the length of the list is the same as the number of frames in the animation here?

js/ui/animation.js
```diff
@@ -30,11 +31,10 @@ var Animation = new Lang.Class({
...
     play: function() {
         if (this._isLoaded && this._timeoutId == 0) {
-            if (this._frame == 0)
-                this._showFrame(0);
-
-            this._timeoutId = GLib.timeout_add(GLib.PRIORITY_LOW, this._speed, Lang.bind(this, this._update));
-            GLib.Source.set_name_by_id(this._timeoutId, '[gnome-shell] this._update');
+            // Set the frame to be the last one, so when we update it
+            // the first time, the first frame will be displayed.
+            this._frame = this._animations.get_n_children() - 1;
```
cosimoc: Is this always what you want? It's a behavior change from how it worked previously, where if you called `stop()` and then again `play()` it would start again from the frame it stopped at. Now you're always resetting it instead.

js/ui/components/clubhouse.js
```diff
@@ -251,10 +251,15 @@ var ClubhouseAnimator = new Lang.Class({
...
         let metadata = this._animations[uri];
 
-        return new Animation(Gio.File.new_for_path(realUri),
-                             metadata.width,
-                             metadata.height,
-                             200);
+        let animation = new Animation(Gio.File.new_for_path(realUri),
+                                      metadata.width,
+                                      metadata.height,
+                                      100);
```
cosimoc: Does it matter if we put `100` or `200` here at this point?

js/ui/components/clubhouse.js
```diff
@@ -148,6 +153,111 @@ var ClubhouseWindowTracker = new Lang.Class({
...
+    _getClubhousePaths: function() {
+        let paths = [];
+        let installations = Flatpak.get_system_installations(null);
+        installations.unshift(Flatpak.Installation.new_user(null));
+
+        for (let installation of installations) {
+            let app = null;
+            try {
+                app = installation.get_current_installed_app(CLUBHOUSE_ID, null);
+            } catch (err if err.matches(Flatpak.Error, Flatpak.Error.NOT_INSTALLED)) {
```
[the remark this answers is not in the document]
joaquimrocha: Okay.

js/ui/components/clubhouse.js
```diff
@@ -148,6 +153,111 @@ var ClubhouseWindowTracker = new Lang.Class({
...
+        for (let path of this._clubhousePaths) {
+            let completeUri = path + '/files/' + uriSuffix;
+            if (GLib.file_test(completeUri, GLib.FileTest.EXISTS))
+                return completeUri;
+        }
+
+        return null;
+    },
+
+    getAnimationByUri: function(uri) {
```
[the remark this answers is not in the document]
joaquimrocha: I have just realized I should disconnect from the `animation-ready` signal when the banner is dismissed to avoid the cb eventually getting called if in the future we use animations in other places (the animator is a singleton).

I don't see how to add a cb to this function in this case, as it'd require more changes to the animator, but if I am seeing it more complicated than it is, please let me know.

js/ui/components/clubhouse.js
joaquimrocha: Please see my doubt in the reply to your previous comment. I will rename the `getAnimationByUri` to `loadAnimationByUri` in any case and this should avoid some of the confusion.

js/ui/animation.js
joaquimrocha: I know. Anyway, I will add the `Index` suffix if that makes it more clear.

js/ui/animation.js
joaquimrocha: That's not necessary because we're wrapping if the number of frames is higher. This may be useful if we have a repeated pattern of timeouts for the frames.

js/ui/components/clubhouse.js
joaquimrocha: Nope (assuming we have a delays array in the metadata). But I didn't want to change the constructor. I can add a default value to that arg though, if you prefer.

joaquimrocha:
@cosimoc Thanks for the review. I have updated the PR (fixup commits).

js/ui/components/clubhouse.js
```diff
@@ -27,10 +27,12 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
 const Pango = imports.gi.Pango;
+const Json = imports.gi.Json;
```
cosimoc: Alphabetical order

js/ui/components/clubhouse.js
```diff
@@ -149,6 +155,136 @@ var ClubhouseWindowTracker = new Lang.Class({
...
+
+    loadAnimationByPath: function(path) {
+        let res = this._proxy.getAnimationMetadataRemote(path, (results) => {
+            let [metadataVariant] = results;
+            let jsonData = null;
+
+            try {
+                jsonData = Json.gvariant_serialize(metadataVariant);
+            } catch (e) {
+                log('Error parsing animation metadata json: ' + e);
```
cosimoc: Still needs to use `logError()` here

js/ui/components/clubhouse.js
cosimoc: No worries, I just didn't want to change it from one arbitrary value to another :-) I just left it at `200` here.

js/ui/animation.js
cosimoc: OK, let's leave it as it is.

js/ui/components/clubhouse.js
cosimoc: I changed this to a more natural flow.

js/ui/components/clubhouse.js
cosimoc: I ended up implementing my suggestion and the code becomes indeed much simpler as I imagined.

joaquimrocha:
@cosimoc I see. I imagined you were asking something a bit different than passing the animator around. I'd have preferred if we made the Animator receive the proxy as its constructor arg and make it a singleton as it was but using the global `var` way instead of the funciton. This way we'd not require it to be passed every time a class needs it.
But this is merged and working, so we can do it later if needed. Thanks for the help getting it merged.

cosimoc:
@joaquimrocha Yeah initially I wanted to make it that way, but since this is only used by the banner, I found it cleaner to just pass it directly for now. As you say we can change it later if we need to extend it.

PR #369, opened 2018-11-22
clubhouse: Theme the notification banner

manuq:
To mimic the banner inside the clubhouse application.

Rearrange the three main elements of the banner: icon, content box,
and buttons box. In the superclass these elements are layed out inside
horizontal and vertical boxes of type St.BoxLayout. This commit
removes them from the boxes and adds them to a new wrapper widget. The
wrapper has a layout manager Clutter.BinLayout, so the elements are
arranged in layers.

Also update the CSS to make the banner as similar as possible to the
banner inside the clubhouse.

https://phabricator.endlessm.com/T24503

### DIFF
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -930,38 +930,65 @@ popup-separator-menu-item {
 }
 
 // Clubhouse Notifications View
-.clubhouse-notification {
-  font-size: 14pt;
-  width: 20em;
-  margin: 5px;
-  border-radius: 12px;
-  color: #000000;
-  background-color: #ffffff;
-  border: 1px solid #afafaf;
-  &:hover, &:focus { background-color: #dfdfdf; color: #000000; }
 
-  .notification-icon { padding: 5px; icon-size: 4.09em; }
-  .notification-content { padding: 5px; spacing: 5px; }
-  .secondary-icon { icon-size: 2.09em; }
-  .notification-actions {
+$clubhouse-notification-width: 444px;
+$clubhouse-image-width: 175px;
+$clubhouse-image-height: 177px;
+
+.clubhouse-notification {
+  background-color: transparent;
+  width: $clubhouse-notification-width;
+  &:hover, &:focus {
     background-color: transparent;
-    padding-top: 2px;
-    spacing: 1px;
+  }
+  .clubhouse-notification-close-button {
+    color: black;
+    padding: 7px 5px 0 0;
+  }
+  .notification-actions {
+    padding: 0;
+    margin: 0 0 0 30px;
   }
   .notification-button {
-    padding: 4px 4px 5px;
-    background-color: #ffffff;
-    border: 1px solid #afafaf;
-    border-radius: 15px 15px;
+    color: white;
+    background-color: #E97C28;
+    border-radius: 17.5px;
+    border: none;
+    padding: 10px;
+    margin: 5px 12px 5px 0;
+    font: 15px "Metropolis Medium, Sans-Serif";
+    box-shadow: 0 2px 4px 4px rgba(0, 0, 0, .4);
+    text-shadow: 0 0.5px 2px rgba(0, 0, 0, .4);
+    .next {
+        border-radius: 30px;
+        padding: 15px 20px;
+        margin: 0 12px 0 0;
+    }
   }
-}
-
-.clubhouse-notification-label {
-  color: #000000;
-}
-
-.clubhouse-notification-icon-bin > StIcon {
-  icon-size: 5em;
+  .clubhouse-content-box {
+    background-color: rgba(255, 255, 255, .96);
+    border-radius: 12px;
+    border: 1px solid #8A6A45;
+    box-shadow: 0 2px 8px 8px rgba(0, 0, 0, .4);
+    padding: 0;
+    height: ($clubhouse-image-height - 42px);
+    margin-bottom: 23px;
+  }
+  .clubhouse-notification-label {
+    color: black;
+    margin: 7px ($clubhouse-image-width + 15px) 17.5px 30px;
+    font: 15px "Metropolis Medium, Sans-Serif";
+  }
+  .clubhouse-notification-icon-bin {
+    padding: 0;
+    margin: 0;
+  }
+  .clubhouse-notification-image {
+    icon-size: $clubhouse-image-width;
+    width: $clubhouse-image-width;
+    height: $clubhouse-image-height;
+    margin-bottom: 5px;
+  }
 }
 
 .clubhouse-open-button-icon {
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -26,6 +26,7 @@ const Flatpak = imports.gi.Flatpak
 const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
+const Pango = imports.gi.Pango;
 const Shell = imports.gi.Shell;
 const Signals = imports.signals;
 const St = imports.gi.St;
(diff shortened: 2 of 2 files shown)


data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -920,38 +920,69 @@ popup-separator-menu-item {
...
+    padding: 0;
+    margin: 0 0 0 30px;
+  }
+  .notification-button, .notification-button-next {
+    color: white;
+    background-color: #E97C28;
+    // @todo: This gradient background is what we want, but it breaks the box-shadow:
+    // background-gradient-start: #E97C28;
+    // background-gradient-end: #CE5611;
+    // background-gradient-direction: horizontal;
```
joaquimrocha: Let's not add this comment. If we need to have it somewhere, please add it to the Phab task.

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -920,38 +920,69 @@ popup-separator-menu-item {
...
+  .clubhouse-notification-label {
+    color: black;
+    margin: 7px ($clubhouse-image-width + 15px) 17.5px 30px;
+    font: 15px "Metropolis Medium, Sans-Serif";
+  }
+  .clubhouse-notification-icon-bin {
+    padding: 0;
+    margin: 0;
+  }
+  .clubhouse-notification-icon-bin > StIcon {
```
joaquimrocha: The notification may not have a subclass of `StIcon` after the animations PR.

js/ui/components/clubhouse.js
```diff
@@ -219,6 +279,17 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
 
+    // Override this method because we don't want the button
+    // horizontally expanded:
+    addAction: function(label, callback) {
+        let button = new St.Button({ style_class: 'notification-button',
+                                     label: label,
+                                     x_expand: false,
+                                     can_focus: true });
+
+        return this.addButton(button, callback);
```
joaquimrocha: You could just override the `addButton` instead and change the needed properties there (then call the parent's implementation).

js/ui/components/clubhouse.js
```diff
@@ -260,9 +331,9 @@ var ClubhouseNotificationBanner = new Lang.Class({
     },
 
     _setupNextPageButton: function() {
-        let button = new St.Button({ style_class: 'notification-button',
+        let button = new St.Button({ style_class: 'notification-button-next',
                                      label: '»',
-                                     x_expand: true,
+                                     x_expand: false,
```
joaquimrocha: If you override the `addButton` method as suggested above, then you can remove this line completely.

js/ui/components/clubhouse.js
```diff
@@ -195,6 +199,62 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+        this._expandedLabel.actor.clutter_text.line_wrap_mode = Pango.WrapMode.WORD_CHAR;
+
+        hbox.remove_child(contentBox);
+        wrapWidget.add_child(contentBox);
+
+        hbox.remove_child(this._iconBin);
+        wrapWidget.add_child(this._iconBin);
+
+        vbox.remove_child(this._actionBin);
+        wrapWidget.add_child(this._actionBin);
```
joaquimrocha: I'd also like to discuss this with you more directly on Slack.

data/theme/gnome-shell-sass/_endless.scss
manuq: Removed.

data/theme/gnome-shell-sass/_endless.scss
manuq: Yes, if we merge that one first. I added a CSS class `.clubhouse-notification-image` to the St.Icon for now.

js/ui/components/clubhouse.js
manuq: Good point!

js/ui/components/clubhouse.js
manuq: +1

js/ui/components/clubhouse.js
manuq: Sounds good. Otherwise let me know here if there is any other way in a message here.

joaquimrocha:
I have pre-approved the changes for now, but I'd still like to see if there could be a way to set up the widgets without so many changes (but I couldn't discuss this with @manuq yet).
I also think that the animations' PR should be merged before, as it may require changes to this code in order to look right, and it'd be better to do it in this PR than having a new one after the animations.

joaquimrocha:
Adding @cosimoc and @GeorgesStavracas to the review process.
**Please don't merge** this before we have the animations in (as it's cleaner to do so in that order): #368 .

js/ui/components/clubhouse.js
```diff
@@ -192,6 +196,8 @@ var ClubhouseNotificationBanner = new Lang.Class({
             getClubhouseWindowTracker().disconnect(this._clubhouseTrackerHandler);
         });
 
+        this._rearrange_elements();
```
cosimoc: Please use camel case for method names.

js/ui/components/clubhouse.js
```diff
@@ -200,6 +206,62 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+        let wrapWidget = new St.Widget({ layout_manager: new Clutter.BinLayout(),
+                                         x_align: Clutter.ActorAlign.FILL,
+                                         y_align: Clutter.ActorAlign.END });
+        wrapBin.set_child(wrapWidget);
+
+        this._iconBin.set_fill(false, false);
+        this._iconBin.set_x_expand(true);
+        this._iconBin.set_y_expand(true);
+        this._iconBin.set_x_align(Clutter.ActorAlign.END);
+        this._iconBin.set_y_align(Clutter.ActorAlign.END);
```
cosimoc: Is it possible to use `Object.assign()` to simplify this code and the one below?

js/ui/components/clubhouse.js
```diff
@@ -200,6 +206,62 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+                                   x_expand: true,
+                                   y_expand: true });
+        hbox.add_child(wrapBin);
+
+        // A Clutter.BinLayout is used to rearrange the notification
+        // elements in layers:
+        let wrapWidget = new St.Widget({ layout_manager: new Clutter.BinLayout(),
+                                         x_align: Clutter.ActorAlign.FILL,
+                                         y_align: Clutter.ActorAlign.END });
+        wrapBin.set_child(wrapWidget);
```
cosimoc: This is quite a bit of changes. I am curious to know if you could tweak the original content box instead of having a different hierarchy, but it's still OK if not. Perhaps you can add a comment as to why this reparenting it necessary here.

js/ui/components/clubhouse.js
```diff
@@ -276,9 +345,8 @@ var ClubhouseNotificationBanner = new Lang.Class({
     },
 
     _setupNextPageButton: function() {
-        let button = new St.Button({ style_class: 'notification-button',
+        let button = new St.Button({ style_class: 'notification-button-next',
```
cosimoc: A more typical way to implement this would be to add a second style class, instead of overriding the initial one.

js/ui/components/clubhouse.js
manuq: Good catch. Fixed.

js/ui/components/clubhouse.js
manuq: I didn't knew about `Object.assign()`. Yes using it the code is more readable and consistent with the parameters passed to the constructors.

js/ui/components/clubhouse.js
manuq: @joaquimrocha had the same concern. In fact initially I started tweaking the current layout, which is basically a horizontal box inside a vertical box, and overriding the actor's positions. But I couldn't get it to work and even worse looked too hacky. This way is more elegant IMO because it's the layout as I would do it from scratch, using a `Clutter.BinLayout` to layout the actors in layers, arranging them with their expand/align properties and with CSS.

js/ui/components/clubhouse.js
manuq: +1

js/ui/components/clubhouse.js
```diff
@@ -198,6 +204,74 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+        vbox.remove_child(this._actionBin);
+        wrapWidget.add_child(this._actionBin);
+    },
+
+    // Override the callback that changes the button opacity on hover:
+    _sync: function() {
+        this._closeButton.opacity = 255;
+    },
+
+    _resetTimeout: function() {
```
cosimoc: I think this function got added here by mistake. I will remove it before merging it.

manuq: Yeah @cosimoc that was my fault whlie rebasing. thanks for noticing.

PR #370, opened 2018-11-23
clubhouse: Do not close the item banner after a timeout

joaquimrocha:
The banner is supposed to be always visible until the user eventually
dismisses it.

https://phabricator.endlessm.com/T24558


joaquimrocha:
Should be trivial.

PR #379, opened 2018-11-27
clubhouse: Don't add actions directly in the item banner

joaquimrocha:
The item banner was calling the _addActions method directly in its
constructor, but due to some changes in the base class, this is no
longer needed and as a result, it'll end up adding the double of
actions/buttons to the banner.

https://phabricator.endlessm.com/T24625


erikos:
Nice catch! Sounds plausible.

PR #382, opened 2018-11-29
T24563

tchx84:
https://phabricator.endlessm.com/T24563

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -37,6 +37,15 @@ const _HACK_SHADER_MAP = {
 };
 const _HACK_DEFAULT_SHADER = 'desaturate';
 
+const HackableIface = `
+<node>
+  <interface name='com.endlessm.Hackable'>
+    <property name="Hackable" type="b" access="read"/>
+  </interface>
+</node>
+`;
+const HackableProxy = Gio.DBusProxy.makeProxyWrapper(HackableIface);
+
 function _ensureAfterFirstFrame(win, callback) {
     if (win._drawnFirstFrame) {
         callback();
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -437,12 +449,32 @@ var CodingSession = new Lang.Class({
...
+        this._setupToyAppProxy();
+
         _ensureAfterFirstFrame(actor, () => {
             Main.layoutManager.addChrome(this._button);
             this._synchronizeButton(actor.meta_window);
         });
     },
 
+    _setupToyAppProxy: function() {
+        const ToyAppProxy = Gio.DBusProxy.makeProxyWrapper(ToyAppIface);
```
cosimoc: This can be called after you declare `ToyAppIface` so we don't need to create an iface wrapper every time we need to create a proxy.

js/ui/codeView.js
```diff
@@ -437,12 +449,32 @@ var CodingSession = new Lang.Class({
...
+        this._toyAppProxy =
```
ptomato: Are you sure it's harmless to create this proxy for an object that doesn't have that interface?

cosimoc: It should be harmless (we should test it), but we should also verify that the property gives us sensible values in that case, since we use its value directly in `_onToyAppPropsChanged()`

ptomato: Also, maybe we should name the interface something else so that it doesn't seem limited to toy apps?

tchx84: Yeah, I tested that case.

If an app doesn't implement the interface, the properties changed signal will never be emitted and the callback never be called, so I don't think we need to worry about the value.

But I can make even more robust lile
```
this._hackable = this._toyAppProxy.Hackable === true;
```

tchx84: > Also, maybe we should name the interface something else so that it doesn't seem limited to toy apps?

Yeah, I thought about this too, lets do it.

PR #383, opened 2018-11-30
payg: Add success screen

mazen-asef:
Add a success screen with the time added, using a new signal
so we can control the amount of time it is shown.

https://phabricator.endlessm.com/T24128

### DIFF
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -221,7 +221,7 @@ var PaygUnlockDialog = new Lang.Class({
         this.actor.add_child(mainBox)
 
         let titleLabel = new St.Label({ style_class: 'unlock-dialog-payg-title',
-                                        text: _("Your Endless pay-as-you-go usage credit has expired."),
+                                        text: _("Your Endless PayGo usage credit has expired."),
                                         x_align: Clutter.ActorAlign.CENTER });
         mainBox.add_child(titleLabel);
 
@@ -314,6 +314,11 @@ var PaygUnlockDialog = new Lang.Class({
         this._clearTooManyAttemptsId = 0;
         this.connect('destroy', this._onDestroy.bind(this));
 
+        log('Displaying success screen');
+        Main.paygManager.connect('expiry-time-changed', (expirytime) => {
+            this._onSuccess(mainBox);
+        });
+
         this._idleMonitor = Meta.IdleMonitor.get_core();
         this._idleWatchId = this._idleMonitor.add_idle_watch(IDLE_TIMEOUT_SECS * MSEC_PER_SEC, Lang.bind(this, this._onCancelled));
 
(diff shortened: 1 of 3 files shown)


js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
         }
     },
 
+    _onSuccess: function(mainBox) {
+            mainBox.remove_all_children();
+
+            //Measuring the ceiling of the timeRemainingSecs because by the time this runs, some time has already elapsed
+            let timeString = Main.paygManager._convertTimeToString(Math.ceil(Main.paygManager._timeRemainingSecs()));
+            let successLabel = new St.Label({ style_class: 'unlock-dialog-payg-title',
+                                        text: _(timeString + " have been added to your PayGo credit."),
```
wjt: You can't use string concatenation in a translated string, you'd need to use something like:

```js
_("%s have been added to your PayGo credit.", timeString)
```

but this won't work because "1 minute have been added to your PayGo credit" is not correct English (it's "1 minute has" vs "2 minutes have"). Other languages have different plurals. Gettext does support this but I think you're actually going to need to move this string into `timeToString` to do that.

js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
         }
     },
 
+    _onSuccess: function(mainBox) {
```
wjt: I don't think this is the right name for this function. It's called when the expiry time changes. We hope that that will only happen because more time is added, but it's possible that it could be emitted without any change, or with a negative change.

js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
...
+
+            this._successButton.add_style_pseudo_class('default');
+            successButtonBox.add_child(this._successButton);
+
+            mainBox.add_child(successButtonBox);
+
+            //Creating a variable to control the trigger to the signal whether or not it`s
+            //coming from a button, since they are asynchronous.
+            let successButtonClicked = false;
+            log ('Trigger deactivation of ScreenShield');
```
wjt: Shouldn't be here.

js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
...
+                                           y_align: St.Align.CENTER });
+
+            this._successButton.add_style_pseudo_class('default');
+            successButtonBox.add_child(this._successButton);
+
+            mainBox.add_child(successButtonBox);
+
+            //Creating a variable to control the trigger to the signal whether or not it`s
+            //coming from a button, since they are asynchronous.
+            let successButtonClicked = false;
```
wjt: Instead, store the timeout ID returned by `Mainloop.timeout_add()`, and explicitly remove it if the button is clicked.

js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
...
+            log ('Trigger deactivation of ScreenShield');
+            this._successButton.connect('clicked', () => {
+                successButtonClicked = true;
+                Main.paygManager.emit('success-on-expiry-changed');
```
wjt: I don't like the name of this signal. I'd prefer `credit-added` or something like that. And it seems very dodgy that the unlock dialog is emitting signals on the manager.

js/ui/paygUnlockDialog.js
```diff
@@ -328,6 +333,53 @@ var PaygUnlockDialog = new Lang.Class({
         }
     },
 
+    _onSuccess: function(mainBox) {
+            mainBox.remove_all_children();
+
+            //Measuring the ceiling of the timeRemainingSecs because by the time this runs, some time has already elapsed
+            let timeString = Main.paygManager._convertTimeToString(Math.ceil(Main.paygManager._timeRemainingSecs()));
```
wjt: This is kind of an interesting problem because if the system is under heavy load, it could conceivably take more than 1 second for us to get to this point, at which point `Math.ceil` will not save us and the user will be upset because they paid for 1 day and it shows them 23 hours 59 minutes or similar.

PR #385, opened 2018-12-03
T24468 format changes

manuq:
The frame-by-frame animations for the clubhouse require more advanced features than the one introduced in 4fa269cd . The animation JSON format has been improved to allow:

- Repeating frames with different timeouts. This requires passing a custom frame sequence.
- Default delay. All frames that don't define a delay will use it.
- Random delays: A delay can now be a string "START-END" and each time this frame is played the delay will be randomly picked from the specified range.

https://phabricator.endlessm.com/T24468

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -155,6 +155,50 @@ var ClubhouseWindowTracker = new Lang.Class({
     },
 });
 
+var ClubhouseAnimation = new Lang.Class({
+    Name: 'ClubhouseAnimation',
+    Extends: Animation,
+
+    _init: function(file, width, height, defaultDelay, frames) {
+        let speed = defaultDelay || 200;
+        this.parent(file, width, height, speed);
+
+        if (frames)
+            this.setFramesInfo(this._parseFrames(frames));
+    },
+
+    _getCurrentDelay: function() {
+        let delay = this.parent();
+        if (typeof delay === 'string') {
+            let [delayA, delayB] = delay.split('-');
+            return GLib.random_int_range(parseInt(delayA), parseInt(delayB));
+        }
+        return delay;
+    },
+
+    _parseFrame: function(frame) {
+        if (typeof frame === 'string') {
+            let [frameIndex, frameDelay] = frame.split(' ');
+            frameIndex = parseInt(frameIndex);
+
+            if (frameDelay.indexOf('-') == -1)
+                frameDelay = parseInt(frameDelay);
+
+            return [frameIndex, frameDelay];
+        }
+        return [frame, this._speed];
+    },
+
+    _parseFrames: function(frames) {
+        let framesInfo = [];
+        for (let frameInfo of frames) {
+            let [frameIndex, frameDelay] = this._parseFrame(frameInfo);
+            framesInfo.push({'frameIndex': frameIndex, 'frameDelay': frameDelay});
+        }
+        return framesInfo;
+    }
+});
+
 var ClubhouseAnimator = new Lang.Class({
     Name: 'ClubhouseAnimator',
 
@@ -265,13 +309,11 @@ var ClubhouseAnimator = new Lang.Class({
             }
 
             let realPath = this._getClubhousePath(path);
-            let animation = new Animation(Gio.File.new_for_path(realPath),
-                                          metadata.width,
-                                          metadata.height,
-                                          200);
-            if (metadata.delays)
-                animation.setFrameTimeouts(metadata.delays);
-
+            let animation = new ClubhouseAnimation(Gio.File.new_for_path(realPath),
+                                                   metadata.width,
+                                                   metadata.height,
+                                                   metadata['default-delay'],
+                                                   metadata.frames);
             callback(animation);
         });
     },


js/ui/animation.js
```diff
@@ -66,13 +66,21 @@ var Animation = new Lang.Class({
             newFrameActor.show();
     },
 
+    _convertDelay: function(delay) {
+        if (typeof delay === 'string') {
+            let [delay_a, delay_b] = delay.split('-');
+            return GLib.random_int_range(parseInt(delay_a), parseInt(delay_b));
+        }
+        return delay;
```
joaquimrocha: As I mentioned during our hangout, I don't love the idea of having different types of values mixed together, even if the language allows it. I'd rather have a list if we really need random delays. But just FTR now, I also think that if we need randomly delayed frames, then we should have two animations and specify when to load one of the other (and those could be random).

js/ui/animation.js
```diff
@@ -54,14 +54,23 @@ var Animation = new Lang.Class({
         this._isPlaying = false;
     },
 
+    _getCurrentFrame: function() {
+        return this._frames[this._frameIndex];
```
joaquimrocha: You always end up using `this._getCurrentFrame().frame`, so I'd rather have this function return that than always having to declare a variable and get the member from it.

js/ui/animation.js
```diff
@@ -91,6 +100,14 @@ var Animation = new Lang.Class({
     _animationsLoaded: function() {
         this._isLoaded = this._animations.get_n_children() > 0;
 
+        if (this._isLoaded && this._frames.length === 0) {
+            // If a custom sequence of frames wasn't provided,
+            // fallback to play the frames in sequence.
+            for (let i = 0; i < this._animations.get_n_children(); i++) {
+                this._frames.push({'frame': i, 'delay': this._speed});
+            }
```
joaquimrocha: Can be rewritten as `this._frames = new Array(this._animations.get_n_children()).fill({'frame': i, 'delay': this._speed})`

js/ui/animation.js
```diff
@@ -99,17 +116,31 @@ var Animation = new Lang.Class({
...
+            let [frame, delay] = frameInfo.split(' ');
+            frame = parseInt(frame);
 
+            if (delay.indexOf('-') > -1)
+                delay = parseInt(delay);
+
+            return [frame, delay];
+        }
+        return [frameInfo, this._speed];
+    },
```
joaquimrocha: I understand we are passing the metadata format that's already being used in the Clubhouse, however I think this should be parsed by some helper function outside of the `Animation` and set directly it (otherwise we should just pass it the whole json and parse it here too). Also, the more we keep things independent of this arbitrary format, the higher the chances of eventually getting the `Animation` changes upstream.

js/ui/animation.js
```diff
@@ -99,17 +116,31 @@ var Animation = new Lang.Class({
...
+
+    setFrames: function(framesList) {
```
joaquimrocha: Maybe call it something more detailed like `setFrameInfo` or `setFrameMetadata` (like there was before: `setFrameTimeouts`).

js/ui/animation.js
```diff
@@ -99,17 +116,31 @@ var Animation = new Lang.Class({
...
         let wasPlaying = this._isPlaying;
         this.stop();
 
-        this._frameTimeouts = timeoutList;
+        this._frames = [];
+        for (let frameInfo of framesList) {
+            let [frame, delay] = this._parseFrame(frameInfo);
```
joaquimrocha: See comment about having this outside of the `Animation` class.

js/ui/animation.js
manuq: OK! I can do that. So should I extend Animation in the clubhouse? I didn't do it because your previous change to add the different offsets were applied to this class directly, but yeah sounds like this is too much changes for the Animation class.

js/ui/animation.js
manuq: +1

js/ui/animation.js
joaquimrocha: I just meant to express my concerns for the other reviewers to see them, since we talked about it 1:1. I am not picking on your solution, just on the concept that we have random delays for frame animations, because the need for the random case seems to be because we're mixing two animations (blinking + normal). Having the random intervals in JSON or in coding is a different matter, because we could specify the sequence of animations and have the random start times there.
Maybe I should have written this in the Phab task itself. Just so it's clear, this comment is not about invalidating the whole PR, it's about my concern for this complex use-case to satisfy one single animation concept (so far).

js/ui/animation.js
manuq: +1

js/ui/animation.js
joaquimrocha: Ah! Then I missed that. Sorry.

js/ui/animation.js
joaquimrocha: No. I mean, you keep the logic you have here, but **IMO** the `Animation` class shouldn't know about our arbitrary JSON format. It's one thing to have a `setFrameInfo` (similar to what I did with the timeouts), it's another to have it parse the JSON. So I'd parse the JSON in the Clubhouse component (Shell or Clubhouse if we need it only in one place), and pass it to the Animation already with:
`[[frame, delay], [frame, delay], ...]` or whatever is better.
The idea is that the Animation class doesn't know about the string format, just that we need to display the frames at a certain delay.

js/ui/animation.js
manuq: No, I couldn't use `fill` here.

js/ui/animation.js
```diff
@@ -54,25 +54,37 @@ var Animation = new Lang.Class({
         this._isPlaying = false;
     },
 
+    _getCurrentFrame: function() {
+        return this._framesInfo[this._frameIndex];
+    },
+
+    _getCurrentFrameActor: function() {
+        let currentFrame = this._getCurrentFrame();
+        return this._animations.get_child_at_index(currentFrame.frameIndex);
```
joaquimrocha: Sorry I only caught this now, but if you call the other field `delay`, maybe this one should be simply called `index`? (Or the other one called `frameDelay`, just looking for consistency)

js/ui/animation.js
```diff
@@ -83,6 +95,13 @@ var Animation = new Lang.Class({
     _animationsLoaded: function() {
         this._isLoaded = this._animations.get_n_children() > 0;
 
+        if (this._isLoaded && this._framesInfo.length === 0) {
+            // If a custom sequence of frames wasn't provided,
+            // fallback to play the frames in sequence.
+            for (let i = 0; i < this._animations.get_n_children(); i++)
+                this._framesInfo.push({'frameIndex': i, 'delay': this._speed});
```
joaquimrocha: You didn't change this after all :slightly_smiling_face:

js/ui/components/clubhouse.js
```diff
@@ -155,6 +155,50 @@ var ClubhouseWindowTracker = new Lang.Class({
     },
 });
 
+var ClubhouseAnimation = new Lang.Class({
+    Name: 'ClubhouseAnimation',
+    Extends: Animation,
```
joaquimrocha: I had mentioned you don't need to extend this and instead just give it the already formatted frames. I see that to support the random frames (which AFAIK are not used in the Shell for now), overriding the `_getCurrentDelay` makes it easier.
So in the interest of not requiring yet more changes, I am okay with this. But I think we should change it soon so that we don't parse the format again in the Shell, i.e.:
- We keep this class, but feed it an array of objects like `{'index': 10, 'delay-interval': [2000, 5000]}`
(So, in future changes to the format we only change the interpretation in one place.)

js/ui/animation.js
ptomato: I think you meant `map` instead of `fill`.

PR #386, opened 2018-12-03
clubhouse: Pulsate the open icon when highlighted

joaquimrocha:
These changes add the new pulsing icon animation to the open button when
it's highlighted.

https://phabricator.endlessm.com/T24506


js/ui/components/clubhouse.js
```diff
@@ -628,16 +631,31 @@ var ClubhouseOpenButton = new Lang.Class({
...
-        if (highlighted)
-            this._icon.add_style_pseudo_class('highlighted');
-        else
-            this._icon.remove_style_pseudo_class('highlighted');
+        if (highlighted) {
+            this.child = this._pulseIcon;
+            this._pulseAnimation.play();
+        } else {
+            this.child = this._normalIcon;
+            this._pulseAnimation.stop();
```
joaquimrocha: I have tried to use a bin layout to layer both icons and just affect their visible property in this function, but it was not working (due to the way the base class for the button works AFAIU).

js/ui/components/clubhouse.js
```diff
@@ -650,8 +668,11 @@ var ClubhouseButtonManager = new Lang.Class({
 
         Main.layoutManager.addChrome(this._openButton);
 
-        this._closeButton =
-            new St.Button({ child: new St.Icon({ style_class: 'clubhouse-close-button-icon' }) });
+        let icon = new St.Icon({ style_class: 'clubhouse-close-button-icon',
+                                 width: CLUBHOUSE_BUTTON_SIZE,
+                                 height: CLUBHOUSE_BUTTON_SIZE
```
joaquimrocha: This is so that the close button is placed exactly at the same position as the open icon.

js/ui/components/clubhouse.js
```diff
@@ -628,16 +631,31 @@ var ClubhouseOpenButton = new Lang.Class({
...
+            Gio.File.new_for_uri('resource:///org/gnome/shell/theme/clubhouse-icon-pulse.png');
+        this._pulseAnimation = new Animation(gfile,
+                                             CLUBHOUSE_BUTTON_SIZE,
+                                             CLUBHOUSE_BUTTON_SIZE,
+                                             CLUBHOUSE_BUTTON_PULSE_SPEED);
+        this._pulseIcon = this._pulseAnimation.actor;
+
+        this._normalIcon = new St.Icon({ style_class: 'clubhouse-open-button-icon',
+                                         width: CLUBHOUSE_BUTTON_SIZE,
+                                         height: CLUBHOUSE_BUTTON_SIZE
```
cosimoc: Can you not set `icon-size` in the theme instead of hardcoding this here?

js/ui/components/clubhouse.js
cosimoc: Ditto

js/ui/components/clubhouse.js
joaquimrocha: Good catch! I have updated the PR. Thanks.

PR #391, opened 2018-12-05
clubhouse: Ensure the window is focused directly

joaquimrocha:
When the Clubhouse window is hidden, and then reopened, it may not get
focused, so we always activate it when it appears which ensures that.

https://phabricator.endlessm.com/T24771

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -123,11 +123,6 @@ var ClubhouseWindowTracker = new Lang.Class({
 
         this._windowActor = actor;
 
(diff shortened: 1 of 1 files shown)


cosimoc:
@joaquimrocha This PR looks reasonable, but I can't help but feel that there's something odd going on at a deeper level. Basically I'd imagine that the `show()` call would be enough to get the focus right, since it already passes the timestamp through.

Before accepting this PR, I want to make a couple of other experiments.

cosimoc:
I think https://github.com/endlessm/clubhouse/pull/164 should solve it -- let's close this for now.

PR #392, opened 2018-12-06
clubhouse: Hook up sounds for entry button

cfoch:
https://phabricator.endlessm.com/T24716

### DIFF
--- a/js/js-resources.gresource.xml
+++ b/js/js-resources.gresource.xml
@@ -156,6 +156,7 @@
     <file>ui/monitor.js</file>
     <file>ui/paygUnlockDialog.js</file>
     <file>ui/sideComponent.js</file>
+    <file>ui/soundable.js</file>
     <file>ui/status/automaticUpdates.js</file>
     <file>ui/status/orientation.js</file>
     <file>ui/userMenu.js</file>
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -39,6 +39,7 @@ const Mainloop = imports.mainloop;
 const MessageTray = imports.ui.messageTray;
 const NotificationDaemon = imports.ui.notificationDaemon;
 const SideComponent = imports.ui.sideComponent;
+const Soundable = imports.ui.soundable;
 const SoundServer = imports.misc.soundServer;
 const Tweener = imports.ui.tweener;
 
@@ -670,7 +671,7 @@ var ClubhouseNotificationSource = new Lang.Class({
 
 var ClubhouseOpenButton = new Lang.Class({
     Name: 'ClubhouseOpenButton',
-    Extends: St.Button,
+    Extends: Soundable.Button,
 
     _init: function(params) {
         params = params || {};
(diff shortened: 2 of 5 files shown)


cosimoc:
@joaquimrocha implemented the same concept of the sound-able button in https://github.com/endlessm/gnome-shell/pull/394. We should indeed settle on only one version of the two.

joaquimrocha:
My approach is slightly different: it's an independent class for playing one-off sounds (i.e. no overlapping) which uses the logic that was already in the CodeView. I think this way it can be used for other events too, not only on buttons.

cfoch:
Should I still use the Soundable button on top of the SoundItem class?

cosimoc:
@cfoch Yes sounds good. Please rework your PR to use the SoundItem API that Joaquim introduced.

js/ui/soundable.js
```diff
@@ -0,0 +1,73 @@
...
+            GObject.ParamFlags.READWRITE,
+            ''
+        )
+    },
+
+    _init: function(params) {
+        this.parent(params);
+
+        if (this.stop_hover_sound_on_click === undefined)
+            this.stop_hover_sound_on_click = false;
```
cosimoc: If we make the `stop-hover-on-sound-click` property a construct one (by setting `GObject.ParamFlag.CONSTRUCT`), you should not need this, as the setter will be called during initialization.

js/ui/soundable.js
```diff
@@ -0,0 +1,73 @@
...
+            '',
+            GObject.ParamFlags.READWRITE,
+            ''
+        ),
+        'stop-hover-sound-on-click': GObject.ParamSpec.boolean(
+            'stop-hover-sound-on-click',
+            '',
+            '',
+            GObject.ParamFlags.READWRITE,
+            ''
```
cosimoc: Should be `false` since this is a boolean property

js/ui/soundable.js
```diff
@@ -0,0 +1,73 @@
...
+        this._hover_sound_event_id = value;
+        this._hover_sound_item =
+            new SoundServer.SoundItem(this._hover_sound_event_id);
+        this.connect('notify::hover', this._onHoverChanged.bind(this));
+    },
+
+    set click_sound_event_id(value) {
+        this._click_sound_event_id = value;
+        this.connect('clicked', () => {
+            if (this.stop_hover_sound_on_click) {
```
cosimoc: Remove curly braces for single-line blocks

js/ui/soundable.js
```diff
@@ -0,0 +1,73 @@
...
+                this._stopHoverSound();
+            }
+            SoundServer.getDefault().play(this._click_sound_event_id);
+        });
+    },
+
+    _onHoverChanged: function() {
+        if (this.hover) {
```
cosimoc: Remove curly braces here too

js/ui/components/clubhouse.js
```diff
@@ -41,6 +41,7 @@ const NotificationDaemon = imports.ui.notificationDaemon;
 const SideComponent = imports.ui.sideComponent;
 const SoundServer = imports.misc.soundServer;
 const Tweener = imports.ui.tweener;
+const Soundable = imports.ui.soundable;
```
cosimoc: Please keep alphabetical order

js/js-resources.gresource.xml
```diff
@@ -72,6 +72,7 @@
     <file>ui/messageList.js</file>
     <file>ui/modalDialog.js</file>
     <file>ui/mpris.js</file>
+    <file>ui/soundable.js</file>
```
cosimoc: Put this in the endless-specific section (also alphabetical order)

js/ui/soundable.js
```diff
@@ -0,0 +1,68 @@
...
+            '',
+            '',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT,
+            false
+        )
+    },
+
+    _init: function(params) {
+        this.parent(params);
+    },
```
cosimoc: This method does nothing more than the default implementation now, and can be removed.

PR #394, opened 2018-12-06
T24800

joaquimrocha:
https://phabricator.endlessm.com/T24800

### DIFF
--- a/js/misc/soundServer.js
+++ b/js/misc/soundServer.js
@@ -23,6 +23,63 @@ const SoundServerIface = `
 </node>
 `;
 
+var SoundItemStatusEnum = {
+    NONE: 0,
+    PENDING: 1,
+    CANCELLING: 2,
+};
+
+class SoundItem {
+
+    constructor(name) {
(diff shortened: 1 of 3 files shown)


js/misc/soundServer.js
```diff
@@ -23,6 +23,59 @@ const SoundServerIface = `
...
+        this.name = name;
+    }
+
+    get Status() {
+        return SoundItemStatusEnum;
+    }
+
+    play() {
+        // If we are about to play the sound or already have it playing, we do nothing
+        if (this._id === this.Status.PENDING || this._id !== this.Status.NONE)
```
cosimoc: I think that the second condition here is not correct, as it will also catch `this._id === this.Status.CANCELLING`, which should not return early since it should fall in the if below.

joaquimrocha: Good catch. Optimizations some times don't optimize as much as they break everything :slightly_smiling_face:

joaquimrocha:
Nice catch @cosimoc . I have updated the failing logic.

PR #397, opened 2018-12-07
clubhouse: Hook up dialog sounds

cfoch:
https://phabricator.endlessm.com/T24716


js/ui/components/clubhouse.js
```diff
@@ -367,6 +367,10 @@ var ClubhouseNotificationBanner = new Lang.Class({
             getClubhouseWindowTracker().connect('window-changed', this.reposition.bind(this));
 
         this.actor.connect('destroy', this._untrackClubhouse.bind(this));
+        this.connect('activated', this._activatedCb.bind(this));
```
cosimoc: This does not look like it should be here.

PR #398, opened 2018-12-11
payg: Add success screen

mazen-asef:
Add a success screen for the tray applet and the unlock dialog, showing the amount of time added. 

Please note i'm using https://github.com/endlessm/gnome-shell/pull/389 as base, and this should not be merged before PR389 does.

Former PR:
https://github.com/endlessm/gnome-shell/pull/383


js/ui/paygUnlockDialog.js
```diff
@@ -401,9 +402,17 @@ var PaygUnlockDialog = new Lang.Class({
...
         this._successButton.connect('button-press-event', () => {
+            this._onCreditAdded();
+        });
+
+        this._successButton.connect('key-press-event', () => {
+            this._onCreditAdded();
+        });
+    },
+
+    _onCreditAdded: function() {
```
treitter: Is this the right name for this function? Usually, a name like `onFoo()` is used for a handler for the `foo` event, not for a function that's notifying a `foo` event.

It might make more sense to name this `_onButtonActivated` or similar. You might also want to check if the button emits a signal covering both cases and just connect to that. I think there's actually a signal named `activate` for that purpose but double-check how it works.

treitter: on new code, try to instead use the simpler form `funcName() {`

mwleeds:
Just a note, I believe this conflicts with #364 so whichever of the two is merged second will have to resolve the conflicts. That should be simple I think.

mazen-asef:
Am not really satisfied with the wording on the signals for the screenshield (where it sends a `credit-added` signal). Suggestions?

js/ui/payg.js
```diff
@@ -456,6 +457,52 @@ function timeToString(seconds) {
...
+// Similar to timeToStrings, but does not process partial time,
+// since it's meant to be used for expiration time changes, with
+// a wider range of periods
+function successMessage(seconds) {
+
+    let messageStr = null;
+
+    if (seconds < 60) {
+        messageStr = Gettext.ngettext("%s second has", "%s seconds have", seconds).format(Math.floor(seconds));
+        return _("%s been added to your %s credit.").format(messageStr, NOTIFICATION_TITLE_TEXT);
```
wjt: You can't break the translatable string in two like this. It needs to be something like:

```js
return Gettext.ngettext(
    "%s second has been added to your Pay As You Go credit.",
    "%s seconds have been added to your Pay As You Go credit.",
    Math.floor(seconds));
```

And the same for all the others. There may be languages where you can't just translate that bit in isolation.

js/ui/screenShield.js
```diff
@@ -582,7 +582,7 @@ var ScreenShield = new Lang.Class({
         this._syncInhibitor();
 
         Main.paygManager.connect('code-expired', () => { this.lock(true) });
-        Main.paygManager.connect('expiry-time-changed', () => {
+        Main.paygManager.connect('success-message-shown', () => {
```
wjt: This will mean that any change to the expiry time that doesn't go through the lock screen will be ignored. Try provisioning a system with Endless PAYG, adding some credit, signing in, then running `eos-payg-ctl clear-code`. Does the screen lock?

js/ui/paygUnlockDialog.js
```diff
@@ -358,6 +362,56 @@ var PaygUnlockDialog = new Lang.Class({
...
+        });
+
+        this._successButton.connect('key-press-event', () => {
+            this._onButtonConnect();
+        });
+    },
+
+    _onButtonConnect() {
+            Mainloop.source_remove(this._timeoutId);
+            Main.paygManager.emit('success-message-shown');
```
wjt: I still don't like this design where this bit of UI tells some other object to emit a signal. If no unlock dialog is being displayed when expiry-time-changed is emitted, what's supposed to happen?

js/ui/paygUnlockDialog.js
```diff
@@ -358,6 +362,56 @@ var PaygUnlockDialog = new Lang.Class({
...
+        this._successButton.connect('button-press-event', () => {
+            this._onButtonConnect();
+        });
+
+        this._successButton.connect('key-press-event', () => {
+            this._onButtonConnect();
+        });
+    },
+
+    _onButtonConnect() {
```
wjt: This function name is misleading. No button is doing any connecting.

mazen-asef: this method runs when the success button either detects a keyboard tap or a click (on a few lines above it) and it is connecting to the signal emitted by either mouse or keyboard to that object. Or did I get that wrong?

js/ui/paygUnlockDialog.js
mazen-asef: Hmm I did this because the screenshield depended already on the emission of the signal emitted by the paygManager, and since both objects have a reference to the manager, i though it would be ok to use it to make it communicate across the code. 

The idea that the unlock dialog is not being displayed, it won't be able to capture that signal because it was killed by the screenshield, right?

js/ui/screenShield.js
mazen-asef: Actually it does - on the js/ui/payg.js, the tray applet can capture the signal for unlocking the screen. 

I'm not really confortable with the signal names, because here is 

code-expired -> No more time available
expiry-time-changed -> Code successfully added.

I tried to keep the same wording that the tray applet and the unlockDialog uses, but it does not makes much sense for the screenshield, and it seems weird to emit two signals with the same puporse to different objects

mazen-asef: Hmm, I got different behaviour just now - when i tested it out, the screen locked immediatly. After putting 4 days and trying to clear that, it didn't. However, letting time run out did had the desired outcome.

js/ui/payg.js
```diff
@@ -441,6 +441,65 @@ function timeToString(seconds) {
...
+    }
+
+    let hours = Math.floor(minutes / 60);
+    if (hours < 24) {
+        messageStr = Gettext.ngettext("%s hour has", "%s hours have", hours).format(hours);
+        return _("%s been added to your %s credit.").format(messageStr, NOTIFICATION_TITLE_TEXT);
+    }
+
+    let days = Math.floor(hours / 24);
+    if (days < 30) {
```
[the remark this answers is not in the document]
mazen-asef: This has kinda passed by me, but for the success message, the user will rarely buy a product with ""broken"" time, right? Based on that, I put the minutes to show `minutes < 60`, so when buying an hour of credit, it shows 1 Hour, instead of 60 minutes ( basing on the assumption it's more a more common product wording ).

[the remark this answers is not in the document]
treitter: The reason I prefer `threshold * 2` (eg, 120 minutes) is to compensate for the inaccuracy of using `floor()`.

Eg, if someone has 119 minutes left, they'd probably much prefer to see that than "1 hour" which is basically half the remaining time. Same for 47 hours vs. 1 day. It'd be more-accurate to simply round but we can't ever round up because then we'd tell the user they could use the computer longer than they actually can. And that would matter if, eg, they're a university student whose open-computer CS exam ends in 47 hours but they only have 40 hours of credit left (and we told them they have >= 48 hours since we rounded to "2 days"). Better to tell them they have 40 hours in this case or 2 days (in case they have 2.7 days left) and then lock at a slightly later time. (And we'll still give them more accurate times as they approach the actual lock time since we switch to a finer-grained time scale).

js/ui/paygUnlockDialog.js
treitter: I agree with Will. It'd be one thing if you were calling a method on `paygManager` and it happened to emit a signal (especially after some other work) but it looks like `paygManager` is emitting a signal that should really belong to `PaygUnlockScreen`.

js/ui/payg.js
```diff
@@ -457,6 +466,53 @@ function timeToString(seconds) {
     return ("%s %s").format(daysStr, hoursStr);
 }
 
+// Similar to timeToStrings, but does not process partial time,
+// since it's meant to be used for expiration time changes, with
+// a wider range of periods
+function successMessage() {
```
treitter: here (and in a number of other places), please instead use the format `fooBar(args) { ... }` instead of `fooBar: function(args) {` or `function fooBar(args) {`

mazen-asef: I was retesting this and it seems that outside of class declarations the <function> argument is necessary so the file can identify the method, so I have left as it is.

wjt:
Please coordinate with @GeorgesStavracas before merging this branch.

PR #400, opened 2018-12-12
clubhouse: Remove CSS rule for notification image

manuq:
The image size in the CSS was hardcoded and can be removed. This
change has no effect for images coming from the clubhouse with a size
equal or less of the hardcoded size. But for larger images, this
change will allow them to expand to the right.

https://phabricator.endlessm.com/T24917

### DIFF
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -995,8 +995,8 @@ $clubhouse-image-height: 187px;
   }
   .clubhouse-notification-image {
     icon-size: $clubhouse-image-width;
-    width: $clubhouse-image-width;
-    height: $clubhouse-image-height;
+    min-width: $clubhouse-image-width;
+    min-height: $clubhouse-image-height;
   }
 }
 


data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -993,11 +993,6 @@ $clubhouse-image-height: 187px;
     padding: 0;
     margin: 0;
   }
-  .clubhouse-notification-image {
-    icon-size: $clubhouse-image-width;
-    width: $clubhouse-image-width;
-    height: $clubhouse-image-height;
```
cosimoc: I'm on the fence about this; at the moment I believe this style class will be applied both to the clubhouse animations and to the item banners.

For the former, only `width` and `height` are really taken into consideration, since `icon-size` is not used by the animation actor; the item banners use a regular icon in there though, so I'd expect `icon-size` to apply. Did you verify that this does not affect the style of the item banners?

It may also be a good idea to still use the old values as `min-width` and `min-height` as well. But overall I'd like to make sure the `icon-size` is kept for the item banners, or I'm afraid they'll be rendered too small.

manuq: Ouch you are right! I forgot about the items.

manuq: PR updated, I moved the width, height to use minimum values. It does the work:
captura de pantalla de 2018-12-14 13-14-50

PR #401, opened 2018-12-13
payg: Consistently rename PayGo to Pay As You Go

ramcq:
PayGo doesn't seem to be the most common term and I think is more narrowly
applied to specific brands / trademarks than the (admittedly more verbose)
generic Pay As You Go. Switch to this consistently as it's what we were
already using.

Internally we already use payg / PAYG for component naming, so stick to this as
the identifier name.

https://phabricator.endlessm.com/T24125


js/ui/payg.js
```diff
@@ -41,8 +41,8 @@ var SPINNER_ICON_SIZE_PIXELS = 16;
 var SPINNER_ANIMATION_DELAY_SECS = 1.0;
 var SPINNER_ANIMATION_TIME_SECS = 0.3;
 
-const NOTIFICATION_TITLE_TEXT = _("PayGo");
-const NOTIFICATION_EARLY_CODE_ENTRY_TEXT = _("Enter an unlock code to extend PayGo time before expiration.");
+const NOTIFICATION_TITLE_TEXT = _("Pay As You Go");
+const NOTIFICATION_EARLY_CODE_ENTRY_TEXT = _("Enter an unlock code to extend the time before your Pay As You Go credit expires.");
```
wjt:
```suggestion
const NOTIFICATION_EARLY_CODE_ENTRY_TEXT = _("Enter an unlock code to extend the time before your credit expires.");
```

(Do we need to repeat "Pay As You Go" in the title and the body?)

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -845,7 +845,7 @@ popup-separator-menu-item {
     }
 }
 
-// PayGo unlock screen (based on .login-dialog)
+// Pay-As-You-Go unlock screen (based on .login-dialog)
```
treitter: Nitpick: I think the Camel-And-Hyphens style is the ugliest way to write it, so I'd prefer "Pay As You Go" (here and the case below) instead

PR #402, opened 2018-12-13
clubhouse: Align clubhouse button with FtH button

dnohales:
https://phabricator.endlessm.com/T24783

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -751,8 +751,10 @@ var ClubhouseButtonManager = new Lang.Class({
         if (!monitor)
             return;
 
+        let workarea = Main.layoutManager.getWorkAreaForMonitor(monitor.index);
+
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -752,7 +752,8 @@ var ClubhouseButtonManager = new Lang.Class({
             return;
 
         this._openButton.x = monitor.x + monitor.width - this._openButton.width / 2;
-        this._openButton.y = Math.floor(monitor.height / 2.0 - this._openButton.width / 2.0);
+        this._openButton.y =
+            Math.floor((monitor.height - Main.panel.height) / 2.0 - this._openButton.height / 2.0);
```
cosimoc: Instead of looking at the panel height directly, can you use the layout manager API and use the primary monitor workarea? It should already have that computed including the panel (and whatever else may be affecting the workarea)

PR #403, opened 2018-12-19
clubhouse: Fix banners' position when initially showing them

joaquimrocha:
When a banner is sliding in and the Clubhouse is hidden, the banner may
end up in the wrong position in the screen. This happens because, even
though we reassign the banner's X coordinate, the tweener may still be
actively changing it too.

To avoid that, this patch removes any tweens in the banner's actor if a
new position is to be assigned to it, and only changes the
"shouldSlideIn" variable at the end of the tween. This means that until
a first slide in is achieved, the banner cancels and starts a new slide
in animation if its target X coordinate changes.
This gets us the right effect of having the banner sliding in, even when
a new X coordinate is set in the middle of a slide.

https://phabricator.endlessm.com/T24691

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -492,13 +492,19 @@ var ClubhouseNotificationBanner = new Lang.Class({
         let endX = this.actor.x - this.actor.width - margin;
 
         if (this._shouldSlideIn) {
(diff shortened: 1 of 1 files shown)


manuq:
Tested. I cannot reproduce the bug with this change. Looks good!

PR #404, opened 2018-12-19
clubhouse: Don't stop quests when closing their banner

joaquimrocha:
The new approach for closing the quest banner is that we don't stop the
quest, instead we simply inform the Clubhouse that the banner has been
closed.

https://phabricator.endlessm.com/T24641


joaquimrocha:
The Clubhouse PR needs to be added before: https://github.com/endlessm/clubhouse/pull/189

PR #405, opened 2018-12-27
T24406

cosimoc:
https://phabricator.endlessm.com/T24406

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -1116,6 +1116,11 @@ var CodeViewManager = new Lang.Class({
                 windowActor._drawnFirstFrame = true;
             });
         });
+
+        this._stopped = false;
+        global.window_manager.connect('stop', () => {
+            this._stopped = true;
+        });
     },
 
     _addSession: function(actor) {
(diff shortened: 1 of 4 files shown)


cosimoc:
@GeorgesStavracas While your proposed solution should work fine when the restart helper is used (e.g. typing `r` in the Alt+F2 dialog), it won't work when the shell is restarted from the terminal, with e.g. `gnome-shell --replace`.

In that case replacing the WM is done entirely using the X selection; the new mutter instance will send `SelectionClear` [1] when `--replace` is specified on the command line, and the old instance will close the display when it receives it [2].

So I'd like something along the lines of my proposed solution to be re-considered. An alternative API that would still work, and would probably be less invasive, is to expose a simple method such as `meta_display_is_closing()` (or even `meta_is_quit()` since we already have `meta_is_restart()`).

Please advise as to what solution is preferrable.

[1] https://github.com/endlessm/mutter/blob/master/src/core/screen.c#L558
[2] https://github.com/endlessm/mutter/blob/master/src/x11/events.c#L1134

GeorgesStavracas:
retest this please

PR #410, opened 2018-12-28
T24016

GeorgesStavracas:
Hiding the applications while  ParentalControlsManager is not
initialized may lead to losing all icons. Since the icon grid
is updated after ParentalControlsManager is initialized anyway,
and initialization shouldn't take more than a split second,
showing the icons by default is reasonable to prevent data loss.

https://phabricator.endlessm.com/T24016

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -14,6 +14,7 @@ const AppDisplay = imports.ui.appDisplay;
 const Background = imports.ui.background;
 const BackgroundMenu = imports.ui.backgroundMenu;
 const LoginManager = imports.misc.loginManager;
+const ParentalControlsManager = imports.misc.parentalControlsManager;
 
 const DND = imports.ui.dnd;
 const Main = imports.ui.main;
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
> Can we not delay showing the icon grid until the parental controls are initialised?
 Hm, indeed. Should be quick too.

js/ui/overview.js
```diff
@@ -296,6 +297,18 @@ var Overview = new Lang.Class({
         if (this.isDummy)
             return;
 
+        // To avoid creating the IconGrid (through OverviewControls),
+        // delay until ParentalControlsManager is initialized
```
pwithnall: Could do with a reference to https://phabricator.endlessm.com/T24016.

js/ui/layout.js
```diff
@@ -537,10 +540,30 @@ var LayoutManager = new Lang.Class({
...
+
+            this._maybePrepareStartupAnimation();
         }));
     },
 
+    _loadParentalControls: function() {
+        let parentalControlsManager = ParentalControlsManager.getDefault();
+
+        if (!parentalControlsManager.initialized)
+            parentalControlsManager.connect('initialized', this._maybePrepareStartupAnimation.bind(this));
```
pwithnall: Maybe this should be disconnected once `_maybePrepareStartupAnimation()` succeeds?

pwithnall:
We [decided](https://phabricator.endlessm.com/T24016#665616) to go with [the other approach](https://github.com/endlessm/gnome-shell/pull/411).

PR #412, opened 2019-01-02
T24619

tchx84:
https://phabricator.endlessm.com/T24619

### DIFF
--- a/js/ui/shellDBus.js
+++ b/js/ui/shellDBus.js
@@ -62,6 +62,7 @@ const GnomeShellIface = '<node> \
 <property name="Mode" type="s" access="read" /> \
 <property name="OverviewActive" type="b" access="readwrite" /> \
 <property name="ShellVersion" type="s" access="read" /> \
+<property name="FocusedApp" type="s" access="read" /> \
 </interface> \
 </node>';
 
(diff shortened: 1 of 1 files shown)


js/ui/shellDBus.js
```diff
@@ -311,7 +312,15 @@ var GnomeShell = new Lang.Class({
             Main.overview.hide();
     },
 
-    ShellVersion: Config.PACKAGE_VERSION
+    ShellVersion: Config.PACKAGE_VERSION,
+
+    get FocusedApp() {
```
cosimoc: This works to fetch the focused app as a one-shot, but we should also send a `PropertiesChanged` signal when the focused app changes, so clients are able to track it (see the `OverviewActive` property for how to implement this here.)

Fortunately, the `ShellWindowTracker` singleton already tracks the focused app [1], so we can reuse that. While at that, we can probably `ShellApp.get_id()` as what we return here, or whatever ID the `FocusApp()` method already uses.

https://github.com/endlessm/gnome-shell/blob/master/src/shell-window-tracker.c#L108

tchx84: @cosimoc thanks, addressed your comment. One note though, Shellapp.get_Id() returns app ids with a `.desktop` suffix... should we crop it? It will save a lot of manual `.desktop` addition on the clients side. I left it as-in for now.

PR #413, opened 2019-01-03
T25012

cosimoc:
https://phabricator.endlessm.com/T25012

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -320,6 +320,11 @@ var WindowTrackingButton = new Lang.Class({
     }
 });
 
+const SessionDestroyEvent = {
+    SESSION_DESTROY_APP_DESTROYED: 0,
+    SESSION_DESTROY_TOOLBOX_DESTROYED: 1,
+};
+
 var CodingSession = new Lang.Class({
     Name: 'CodingSession',
     Extends: GObject.Object,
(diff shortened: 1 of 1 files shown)


js/ui/codeView.js
```diff
@@ -672,17 +659,25 @@ var CodingSession = new Lang.Class({
...
+        // Save the actor, so we can complete the destroy transition later
+        this._appRemovedActor = this.app;
         this.app = null;
+        this.appRemovedByFlipBack = false;
 
+        // We assume we have a toolbox here, or this.appRemovedByFlipBack would be false
         this._setState(STATE_TOOLBOX);
 
-        this._completeRemoveWindow();
+        let actor = this._actorForCurrentState();
```
joaquimrocha: I see that all uses of this function verify whether `actor` is `null`, is that needed elsewhere/here?

js/ui/codeView.js
```diff
@@ -1141,39 +1136,30 @@ var CodeViewManager = new Lang.Class({
...
+        if (this._stopped)
             return false;
 
-        if (session.appRemovedByFlipBack) {
-            session.removeAppWindow();
-            return true;
+        // First, determine if this is an app window getting destroyed
+        let session = this._getSession(actor, SessionLookupFlags.SESSION_LOOKUP_APP);
+        if (session) {
+            // If the app window destroyed because the toolbox flipped it back,
```
joaquimrocha: `If the app window destroyed` maybe it's missing `*was* destroyed`?

PR #415, opened 2019-01-03
clubhouse: Activate the app instead of using a D-Bus method

joaquimrocha:
The Clubhouse side component was being shown by just calling a D-Bus
method on the app. However, this meant that the usual activation logic
for launching apps was not used, and thus there was no startup
notification for the user; i.e. no spinner cursor was being shown to the
user, indicating that the app is loading, which hindered the UX.

This patch fixes that issue by launching the app with the meantioned
activation logic. It however only does that if the app is not running
already, as a workaround for a common issue in the Shell: if the app is
already running, activating it again will show the spinner cursor for a
long time.

https://phabricator.endlessm.com/T24728

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -796,7 +796,7 @@ var ClubhouseComponent = new Lang.Class({
     Extends: SideComponent.SideComponent,
 
     _init: function() {
-        this._useClubhouse = this._imageUsesClubhouse() && this._hasClubhouse();
+        this._useClubhouse = this._imageUsesClubhouse() && !!this._getClubhouseApp();
         if (!this._useClubhouse)
             return;
 
@@ -874,7 +874,15 @@ var ClubhouseComponent = new Lang.Class({
     },
 
     callShow: function(timestamp) {
-        this.proxy.showRemote(timestamp);
+        if (this.proxy.g_name_owner) {
+            this.proxy.showRemote(timestamp);
+            return;
+        }
+
+        // We only activate the app here if it's not yet running, otherwise the cursor will turn
+        // into a spinner for a while, even after the window is shown.
+        // @todo: Call activate alone when we fix the problem mentioned above.
+        this._getClubhouseApp().activate();
     },
 
     callHide: function(timestamp) {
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -874,7 +877,16 @@ var ClubhouseComponent = new Lang.Class({
...
+        if (this.proxy.g_name_owner) {
+            this.proxy.showRemote(timestamp);
+            return;
+        }
+
+        // We only activate the app here if it's not yet running, otherwise the cursor will turn
+        // into a spinner for a while, even after the window is shown.
+        // @todo: Use the activate logic alone when we fix the problem mentioned above.
+        let activationContext = new AppActivation.AppActivationContext(this._clubhouseApp);
+        activationContext.activate();
```
cosimoc: You should pass the timestamp here, it will be used by `activate()`.
However we don't need any of the complexity of `AppActivationContext` which exists to deal with the splash screens, and using it could potentially introduce issues down the line...

I'd prefer it if you called directly `activate()` or `activate_full()` on the `ShellApp`.

js/ui/components/clubhouse.js
```diff
@@ -796,7 +797,9 @@ var ClubhouseComponent = new Lang.Class({
     Extends: SideComponent.SideComponent,
 
     _init: function() {
-        this._useClubhouse = this._imageUsesClubhouse() && this._hasClubhouse();
+        this._clubhouseApp = this._getClubhouseApp();
```
cosimoc: Apps are already cached in `ShellAppSystem` and doing the lookup for the clubhouse app should be cheap (a hash table lookup) so I'd like to avoid storing it as an extra state here.

cosimoc:
I've tested this and it works, but we need another PR to add `StartupNotify=true` to the clubhouse desktop file. I'll do that in a moment, but in the meantime we can squash and merge this.

joaquimrocha:
> I've tested this and it works, but we need another PR to add `StartupNotify=true` to the clubhouse desktop file. I'll do that in a moment, but in the meantime we can squash and merge this.

Damn. I completely forgot to open the PR about the `StartupNotify=true` (I did have it locally). Thanks!

PR #419, opened 2019-01-04
status/payg: Fix applet visibility

mazen-asef:
Fix the issue where the applet does not appear, since it's
depending on a library that is not loaded during build.

Remove all hard dependencies of the library and always instantiate
the indicator.

https://phabricator.endlessm.com/T24125


js/misc/meson.build
```diff
@@ -5,7 +5,6 @@ jsconf.set('GETTEXT_PACKAGE', meson.project_name())
 jsconf.set('LIBMUTTER_API_VERSION', mutter_api_version)
 jsconf.set10('HAVE_BLUETOOTH', bt_dep.found())
 jsconf.set10('HAVE_NETWORKMANAGER', have_networkmanager)
-jsconf.set10('HAVE_PAYG', payg_dep.found())
```
wjt: Please also remove the optional dependency itself. `payg_dep` is defined in the top-level `meson.build`.

meson.build
```diff
@@ -148,8 +148,6 @@ else
   have_networkmanager = false
 endif
 
-payg_dep = dependency('eos-payg-1', version: payg_req, required: false)
```
wjt: Sorry, I didn't notice this before.. I think you could also remove the `payg_req` variable defined somewhere above.

PR #420, opened 2019-01-05
windowManager: always pass window destroy event to CodeViewManager

cosimoc:
At the moment, we only forward the window destroy event to
CodeViewManager when a determination is made that the window needs an
animation, and is a normal window.

However, whether the window needs an animation also depends on
external factors, such as whether the overview is visible. We don't
want to miss the even in those cases, or we can e.g. end up with a
mismatched app/toolbox pair when an app is closed while the overview
is visible.

This commit fixes that by simply passing the event over to
CodeViewManager earlier. If the event referred to a window that is not
tracked by CodeViewManager, no harm is done anyway.

https://phabricator.endlessm.com/T25040

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -2401,6 +2401,9 @@ var WindowManager = new Lang.Class({
         if (window.is_attached_dialog())
             this._checkDimming(window.get_transient_for(), window);
 
+        if (this._codeViewManager.handleDestroyWindow(actor))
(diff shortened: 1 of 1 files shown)


joaquimrocha:
jenkins build this please

joaquimrocha:
test this please

joaquimrocha:
LGTM. I will merge after it successfully builds.

PR #424, opened 2019-01-07
clubhouse: Use blank icon in quest view until the animation is loaded

joaquimrocha:
When the quest banner is shown, it sometimes happens that its character
animation is still not loaded, and for those milliseconds a default icon
is shown. This makes things look a bit broken, thus, before we address
the issue in a more complex way, this patch simply uses an empty icon
(the white space of the dialog is shown) until the animation is
displayed.
This is less "shocking" than seeing the default icon and makes for a
safe and simple improvement ATM.

https://phabricator.endlessm.com/T24941


cosimoc:
Closing this old PR, since we had decided in the Phabricator ticket to use a different fix.

PR #427, opened 2019-01-18
T24855 Fix polkit dialogue signal handling

pwithnall:
This is a tentative fix for the residual polkit dialogue seen on some OpenQA test runs, such as https://openqa.endlessm.com/tests/12952#step/desktop_travel/23. That polkit dialogue should have been cancelled by polkitd, as can be seen from the journal for that test run:
```
Jan 18 11:20:30 endless polkitd[367]: Operator of unix-session:2 FAILED to authenticate to gain authorization for action org.freedesktop.Flatpak.runtime-install for system-bus-name::1.249 [/usr/bin/gnome-software --gapplication-service] (owned by unix-user:test)
Jan 18 11:20:30 endless gnome-software[1612]: failed to call gs_plugin_app_install on flatpak: failed to get permissions: Unable to load metadata from remote eos-apps: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Flatpak system operation Deploy not allowed for user
```

Helps https://phabricator.endlessm.com/T24855

### DIFF
--- a/js/ui/components/polkitAgent.js
+++ b/js/ui/components/polkitAgent.js
@@ -198,10 +198,10 @@ var AuthenticationDialog = new Lang.Class({
         this.destroySession();
         this._session = new PolkitAgent.Session({ identity: this._identityToAuth,
                                                   cookie: this._cookie });
-        this._session.connect('completed', Lang.bind(this, this._onSessionCompleted));
-        this._session.connect('request', Lang.bind(this, this._onSessionRequest));
-        this._session.connect('show-error', Lang.bind(this, this._onSessionShowError));
-        this._session.connect('show-info', Lang.bind(this, this._onSessionShowInfo));
+        this._sessionCompletedId = this._session.connect('completed', Lang.bind(this, this._onSessionCompleted));
+        this._sessionRequestId = this._session.connect('request', Lang.bind(this, this._onSessionRequest));
+        this._sessionShowErrorId = this._session.connect('show-error', Lang.bind(this, this._onSessionShowError));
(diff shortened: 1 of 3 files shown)


wjt:
I don't fully understand the sequence of events here. I think it goes something like:

* test suite switches to VT4
* back in the X session, gnome-software tries to install some updates
* because the org.freedesktop.Flatpak.rules polkit rule only returns YES if `subject.active == true` (boo, comparison to boolean…), this falls through to requiring interactive authentication
* due to a known issue that will be fixed in a future g-s, interactive authentication goes ahead even though the action was not user-initiated
* something (and this is the point where I'm not following) causes that authentication to actively fail – what is that?
* the shell does not correctly dismiss the auth dialog when told to by polkit

pwithnall:
>     * something (and this is the point where I'm not following) causes that authentication to actively fail – what is that?

I’m not sure. The polkit messages don’t give any hints, and I don’t think it’s worth trying to reproduce the actual `desktop_travel` environment to get debug output of `polkitd` at the moment. Might be worth doing that if this branch doesn’t fix the OpenQA test. This branch definitely fixes a bug, though.

My working theory was that `polkitd` eventually timed out waiting for the agent to respond, and cancelled the request — but looking at the `polkitd` code, it seems to go to lengths to make sure that agent requests don’t have a timeout.

My new working theory is that it’s some kind of timeout in gnome-software, or a cancellation of the previous background request as a result of the command line `flatpak` invocations. I don’t have much evidence to support that though. It does seem that there’s a default timeout of 60s on plugin jobs, set in `gs-plugin-job.c`.

PR #429, opened 2019-01-28
clubhouse: Play pulse sound only if highlighted button visible

cfoch:
https://phabricator.endlessm.com/T25236

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -789,7 +789,8 @@ var ClubhouseButtonManager = new Lang.Class({
     },
 
     setSuggestOpen: function(suggestOpen) {
-        this._openButton.setHighlighted(suggestOpen);
+        if (this._openButton.visible)
+            this._openButton.setHighlighted(suggestOpen);
     },
 
     setVisible: function(visible) {


js/ui/components/clubhouse.js
```diff
@@ -716,21 +723,48 @@ var ClubhouseOpenButton = new Lang.Class({
...
+
+    _play_stop_highlighted_sound() {
+        const states = [
+            SoundServer.SoundItemStatusEnum.NONE,
+            SoundServer.SoundItemStatusEnum.PENDING,
+            SoundServer.SoundItemStatusEnum.CANCELLING
+        ];
+
+        if (this.visible && this.highlighted) {
+            // If the sound item has already an UUID assigned, then do not
```
manuq: "UUID assigned" is an implementation detail of the sound server. There are no UUIDs here. I would keep the comment with the method names used here, or remove it.

js/ui/components/clubhouse.js
```diff
@@ -716,21 +723,48 @@ var ClubhouseOpenButton = new Lang.Class({
...
+        this._highlighted = highlighted;
+        this.notify('highlighted');
+    },
+
+    get highlighted() {
+        return this._highlighted;
+    },
+
+    _play_stop_highlighted_sound() {
+        const states = [
```
manuq: These states should be defined outside of the callback.

js/ui/components/clubhouse.js
```diff
@@ -716,21 +723,48 @@ var ClubhouseOpenButton = new Lang.Class({
...
             new SoundServer.SoundItem('clubhouse/entry/pulse-loop');
+
+        this.connect('notify::visible',
+                     this._play_stop_highlighted_sound.bind(this));
+        this.connect('notify::highlighted',
+                     this._play_stop_highlighted_sound.bind(this));
     },
 
-    setHighlighted: function(highlighted) {
+    set highlighted(highlighted) {
```
manuq: So you are converting the highlighted setter into a property. But should this be set at all if the button is not visible? A simpler approach would be to skip setting the highlighted if the button is hidden, which will also save playing an animation.

cfoch: I will opt for this. I was thinking that after completing the InitialContact puzzle the pulsating button would be shown unhighlighted but this is not the case.

cfoch:
I am opting for an easier solution, but just in case I left the old changes in T25236-visibility-dependent

PR #430, opened 2019-01-30
T25241 clubhouse button ext monitor

danigm:
This PR fixes the button positioning with multiples monitors. I've tried adding an external monitor to the right, top, left and bottom and it works okay in all the cases.

To do this I've changed the button, instead of drawing the complete circle 110x110 and positioned at the edge of the screen, I've changed that to draw it 55x110 so there's no part of it in the secondary monitor.

The second commit fixes the button positioning when we've the external monitor above or below the primary.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -776,7 +776,7 @@ var ClubhouseButtonManager = new Lang.Class({
         let workarea = Main.layoutManager.getWorkAreaForMonitor(monitor.index);
 
         this._openButton.x = monitor.x + monitor.width - this._openButton.width / 2;
-        this._openButton.y = Math.floor(workarea.height / 2.0 - this._openButton.height / 2.0);
+        this._openButton.y = workarea.y + Math.floor(workarea.height / 2.0 - this._openButton.height / 2.0);
 
         this._updateCloseButtonPosition();
     },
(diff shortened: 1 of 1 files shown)


joaquimrocha:
@danigm IMO, this approach is clever but too hacky and may result in issues in the future if we try to animate the position of the button with a tween (to do what Robin wanted with a "bobbing animation").
If we don't have animate the button's position on our own, then I think it'd make sense to just use half of the button in the asset (i.e. clipping the animation file).
As a more robust alternative, maybe it's possible to add the button to another actor (that sits in the right position, i.e. at the right edge of the screen) but positioning it on that actor's right edge, so hopefully that actor's visible clip will not allow the right half of the button to be seen. Feel free to ping me on Slack if this isn't clear.

danigm:
> @danigm IMO, this approach is clever but too hacky and may result in issues in the future if we try to animate the position of the button with a tween (to do what Robin wanted with a "bobbing animation").
> If we don't have animate the button's position on our own, then I think it'd make sense to just use half of the button in the asset (i.e. clipping the animation file).
> As a more robust alternative, maybe it's possible to add the button to another actor (that sits in the right position, i.e. at the right edge of the screen) but positioning it on that actor's right edge, so hopefully that actor's visible clip will not allow the right half of the button to be seen. Feel free to ping me on Slack if this isn't clear.

@joaquimrocha, thanks for the review. I tried a quick solution, but your solution is better and easier, so I did as you suggested. I've added a wrap Actor and apply the clip there, so there's no need to change css styles and if we change the animation in the future, it'll be easier to change.

I thought about to apply the clip directly to the button without the wrap Actor, but that change will need to change the css styles and animation, so I think that's better with the wrap.

PR #433, opened 2019-01-31
sideComponent: Clip side components during slide

danigm:
When we've multiple monitors and the workarea is expanded, we can view
the animation for the side components in the external monitor during the
slide.

To avoid this, this patch adds a callback for each Tweener animation to
clip the actor during the update. Doing this, the actor isn't visible in
the other monitors.

This problem was detected as part of the https://github.com/endlessm/gnome-shell/pull/430, so it's related with this task: https://phabricator.endlessm.com/T25241

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1776,14 +1776,20 @@ var WindowManager = new Lang.Class({
                                onCompleteParams: [shellwm, actor],
                                onOverwrite: onOverwrite,
                                onOverwriteScope: this,
-                               onOverwriteParams: [shellwm, actor]
+                               onOverwriteParams: [shellwm, actor],
+                               onUpdate: this._clipActor,
+                               onUpdateParams: [Clutter.ContentGravity.TOP, endY, actor],
                              });
         } else {
             let endX;
-            if (actor.x <= monitor.x)
+            let origin;
+            if (actor.x <= monitor.x) {
                 endX = monitor.x - actor.width;
-            else
+                origin = Clutter.ContentGravity.LEFT;
+            } else {
                 endX = monitor.x + monitor.width;
+                origin = Clutter.ContentGravity.RIGHT;
+            }
 
             Tweener.addTween(actor,
                              { x: endX,
@@ -1794,7 +1800,9 @@ var WindowManager = new Lang.Class({
                                onCompleteParams: [shellwm, actor],
                                onOverwrite: onOverwrite,
                                onOverwriteScope: this,
-                               onOverwriteParams: [shellwm, actor]
+                               onOverwriteParams: [shellwm, actor],
+                               onUpdate: this._clipActor,
+                               onUpdateParams: [origin, endX, actor],
                              });
         }
     },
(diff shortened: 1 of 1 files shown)


js/ui/windowManager.js
```diff
@@ -1776,14 +1776,20 @@ var WindowManager = new Lang.Class({
...
+                               onUpdateParams: ["top", endY, actor],
                              });
         } else {
             let endX;
-            if (actor.x <= monitor.x)
+            let origin;
+            if (actor.x <= monitor.x) {
                 endX = monitor.x - actor.width;
-            else
+                origin = "left";
```
joaquimrocha: Instead of using strings for determining directions, I think it'd be better to use constant pre-declared values. You can even use `ClutterContentGravity` for this.

js/ui/windowManager.js
```diff
@@ -2205,6 +2223,33 @@ var WindowManager = new Lang.Class({
             this._hideOtherWindows(actor, animateFade);
     },
 
+    _clipActor : function(origin, position, actor) {
+        switch (origin) {
+            case "right": {
+                let clipx = position - actor.x;
```
joaquimrocha: Nitpick: Shouldn't this be called `clipWidth`?

js/ui/windowManager.js
```diff
@@ -2205,6 +2223,33 @@ var WindowManager = new Lang.Class({
...
+                actor.set_clip(0, 0, clipx, actor.height);
+                break;
+            }
+            case "left": {
+                let clipx = actor.width - (actor.x - position);
```
joaquimrocha: Nitpick: maybe it should continue using the CamelCase fashion here (`clipX`) and the rest of the switch body too.

js/ui/windowManager.js
```diff
@@ -2205,6 +2223,33 @@ var WindowManager = new Lang.Class({
...
+                actor.set_clip(clipx, 0, actor.width, actor.height);
+                break;
+            }
+            case "top": {
+                let clipy = actor.height - (actor.y - position);
+                actor.set_clip(0, clipy, actor.width, actor.height);
+                break;
+            }
+            case "bottom": {
+                let clipy = position - actor.y;
```
joaquimrocha: Nitpick: Shouldn't this be called `clipHeight`?

js/ui/windowManager.js
```diff
@@ -2205,6 +2223,33 @@ var WindowManager = new Lang.Class({
             this._hideOtherWindows(actor, animateFade);
     },
 
+    _clipActor : function(origin, position, actor) {
+        switch (origin) {
```
joaquimrocha: I'd condense this switch statement by pre-setting `clipX, clipY, clipWidth, clipHeight` to the default values, and just update the values of the affected variables according to the switch case. Meaning we'd only call `actor.set_clip(...)` once.

js/ui/windowManager.js
danigm: I've used the ClutterContentGravity as you suggested

js/ui/windowManager.js
danigm: I've updated this switch to modify the `clipX, clipY, clipWidth, clipHeight` and do only one `actor.set_clip`.

js/ui/windowManager.js
```diff
@@ -1776,14 +1776,20 @@ var WindowManager = new Lang.Class({
                                onCompleteParams: [shellwm, actor],
                                onOverwrite: onOverwrite,
                                onOverwriteScope: this,
-                               onOverwriteParams: [shellwm, actor]
+                               onOverwriteParams: [shellwm, actor],
+                               onUpdate: this._clipActor,
+                               onUpdateParams: ["top", endY, actor],
```
joaquimrocha: The `"top"` here needs to be updated.

js/ui/windowManager.js
```diff
@@ -2172,19 +2180,27 @@ var WindowManager = new Lang.Class({
...
                                onOverwriteScope: this,
-                               onOverwriteParams: [shellwm, actor]
+                               onOverwriteParams: [shellwm, actor],
+                               onUpdate: this._clipActor,
+                               onUpdateParams: [Clutter.ContentGravity.TOP, monitor.y - actor.height, actor],
                              });
         }
         else {
             let origX = actor.x;
+            let position;
```
GeorgesStavracas: It would be more semantic to name this variable `destX` or `endX` I guess.

js/ui/windowManager.js
```diff
@@ -1776,14 +1776,20 @@ var WindowManager = new Lang.Class({
                                onCompleteParams: [shellwm, actor],
                                onOverwrite: onOverwrite,
                                onOverwriteScope: this,
-                               onOverwriteParams: [shellwm, actor]
+                               onOverwriteParams: [shellwm, actor],
+                               onUpdate: this._clipActor,
+                               onUpdateParams: [Clutter.ContentGravity.TOP, endY, actor],
```
GeorgesStavracas: ClutterContentGravity has a very specific use case, and using it like this is very weird to me, but meh.

joaquimrocha: @GeorgesStavracas If there's another enum that we can already use, then let us know. Otherwise, at least IMO, it's fine to use this and it's probably not worth creating yet another 4 positions enum.

GeorgesStavracas: There is no better alternative in Clutter (and this isn't an actionable comment)

PR #435, opened 2019-02-14
codeView: Preload toolbox before the flip

danigm:
The toolbox is being lazy loaded when the user clicks the FtH button. If
the toolbox is a bit complex, this loading can take some time and the
feedback for the user is odd because the interface freezes after the
click.

To try to *improve* the flipping time, we can preload the hack toolbox
for the app before the user clicks the FtH. This patch calls to the init
method of the HackToolbox, when the button is shown only if it's not
init yet.

This depends on this change for the hack toolbox:
https://github.com/endlessm/hack-toolbox-app/pull/130

if the hack toolbox app doesn't have this action yet, the preload won't
work and the FtH will continue working as usual.

https://phabricator.endlessm.com/T25093

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -930,9 +930,17 @@ var CodingSession = new Lang.Class({
             !Main.overview.visible &&
             !inFullscreen &&
             !locked &&
-            this._hackable)
+            this._hackable) {
+
+            if (!this.toolbox) {
+                this._toolboxAppActionGroup.activate_action(
+                    'init',
+                    new GLib.Variant('(ss)', [_getAppId(this.app.meta_window),
+                                              _getWindowId(this.app.meta_window)]));
+            }
+
             this._button.show();
-        else
+        } else
             this._button.hide();
     },
 


ptomato:
I only looked at @danigm's last commit on the list, it seems the branch is undergoing some renovations :smile:

Looks generally good to me, but what would be the consequence of the shell trying to activate the `init` action if the currently installed toolbox doesn't yet have it? Can we just catch an exception and ignore it in that case?

ptomato:
Oh never mind, I see you addressed that possibility in the comment on the Phabricator ticket, so I assume you tested it and found that it wasn't necessary to catch an exception?

danigm:
> Oh never mind, I see you addressed that possibility in the comment on the Phabricator ticket, so I assume you tested it and found that it wasn't necessary to catch an exception?

Yes, I tested and it works as usual if the init action doesn't exists, so I think there's no problem with that, it should be compatible with any version.

ptomato:
Sounds good, can you rebase this on the new gnome-shell master?

danigm:
Rebased with current master

PR #437, opened 2019-02-20
config: Remove obsolete notifications

mazen-asef:
Pay-As-You-Go Notifications keep appearing after being
useful when using eos-payg-ctl in the middle of a
time-session or when the session expires normally. So we
clean all notifications on the message tray when a
expiry-time-changed signal is sent.

https://phabricator.endlessm.com/T25246

### DIFF
--- a/js/ui/payg.js
+++ b/js/ui/payg.js
@@ -471,8 +471,8 @@ var PaygNotifier = new Lang.Class({
             return;
         }
 
-        if (this._notification)
-            this._notification.destroy();
+        // Clear previous notification
+        clearNotification();
 
         let source = new MessageTray.SystemNotificationSource();
         Main.messageTray.add(source);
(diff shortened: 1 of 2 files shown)


js/ui/payg.js
```diff
@@ -510,5 +510,11 @@ var PaygNotifier = new Lang.Class({
         this._notification.connect('destroy', () => {
             this._notification = null;
         });
+
+    },
+
+    clearNotification() {
```
wjt: This code is copied from near the top of the previous function: could you make the previous function call this new function instead?

mazen-asef: Ah, yes! will do :)

wjt:
Please squash & merged. @GeorgesStavracas FYI since you're midway through the shell rebase.

GeorgesStavracas:
It's fine to land this

PR #438, opened 2019-02-26
clubhouse: Clip close button near the right side

danigm:
When there's an external monitor on the right side we need to clip the
clubhouse close button to avoid this button to be visible outside the
main monitor.

This patch clips the button on every movement so it'll be full visible
when it's full inside the monitor and clipped to the monitor with when
it's at the edge.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -769,6 +769,18 @@ var ClubhouseButtonManager = new Lang.Class({
         let clubhouseWindowX = getClubhouseWindowTracker().getWindowX();
         if (clubhouseWindowX != -1)
             this._closeButton.x = clubhouseWindowX - this._closeButton.width / 2;
+
+        // To avoid this button appearing in a possible right monitor during
+        // the animation we need to clip it to adjust to the monitor width
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+        let buttonEdge = this._closeButton.x + this._closeButton.width;
+        let monitorEdge = monitor.x + monitor.width;
+        let offset =  buttonEdge - monitorEdge;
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -769,6 +769,18 @@ var ClubhouseButtonManager = new Lang.Class({
...
+
+        // To avoid this button appearing in a possible right monitor during
+        // the animation we need to clip it to adjust to the monitor width
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+        let button_edge = this._closeButton.x + this._closeButton.width;
+        let monitor_edge = monitor.x + monitor.width;
+        let offset =  button_edge - monitor_edge;
+        offset = offset > 0 ? offset : 0;
```
cosimoc: You can simplify this with `Math.max()` as suggested in the other PR

js/ui/components/clubhouse.js
```diff
@@ -769,6 +769,18 @@ var ClubhouseButtonManager = new Lang.Class({
...
         if (clubhouseWindowX != -1)
             this._closeButton.x = clubhouseWindowX - this._closeButton.width / 2;
+
+        // To avoid this button appearing in a possible right monitor during
+        // the animation we need to clip it to adjust to the monitor width
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor)
+            return;
+        let button_edge = this._closeButton.x + this._closeButton.width;
+        let monitor_edge = monitor.x + monitor.width;
```
cosimoc: Use camel case for variables

danigm:
Solved in a commit of this PR: https://github.com/endlessm/gnome-shell/pull/451/

PR #439, opened 2019-02-26
clubhouse: Clip notification banner slide in/out

danigm:
When there's an external monitor on the right side we need to clip the
clubhouse notification banner to avoid this bubble to be visible outside
the main monitor.

This patch clips the notification bubble on every movement so it'll be
full visible when it's full inside the monitor and clipped to the
monitor width when it's at the edge.

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -507,6 +507,7 @@ var ClubhouseNotificationBanner = new Lang.Class({
                              { x: endX,
                                time: CLUBHOUSE_BANNER_ANIMATION_TIME,
                                transition: 'easeOutQuad',
+                               onUpdate: this._clipActor.bind(this),
                                onComplete: () => {
                                    // Ensure it only slides in once
                                    this._shouldSlideIn = false;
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -564,13 +565,28 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
                            onComplete: () => {
                                this.actor.destroy();
                                this.actor = null;
                            }
                          });
     },
 
+    _clipActor: function() {
+        let monitor = Main.layoutManager.primaryMonitor;
+        if (!monitor) {
```
cosimoc: Remove curly braces for single-line blocks

js/ui/components/clubhouse.js
```diff
@@ -564,13 +565,28 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+            return;
+        }
+
+        let monitor_edge = monitor.x + monitor.width;
```
cosimoc: Use camel case for variables

js/ui/components/clubhouse.js
```diff
@@ -564,13 +565,28 @@ var ClubhouseNotificationBanner = new Lang.Class({
...
+        let actor_edge = this.actor.x + this.actor.width;
+        let offset = actor_edge - monitor_edge;
+        offset = offset > 0 ? offset : 0;
```
cosimoc: You can simplify this with `let offset = Math.max(0, actor_edge - monitor_edge);`

js/ui/components/clubhouse.js
```diff
@@ -507,6 +507,7 @@ var ClubhouseNotificationBanner = new Lang.Class({
                              { x: endX,
                                time: CLUBHOUSE_BANNER_ANIMATION_TIME,
                                transition: 'easeOutQuad',
+                               onUpdate: this._clipActor.bind(this),
```
cosimoc: I wonder if instead of using `onUpdate()` you could tween the `clip` property of the actor itself. No worries if it's inconvenient.

danigm: I don't know, I can't test this right now because I've not the external monitor here.

danigm:
Solved in a commit of this PR: https://github.com/endlessm/gnome-shell/pull/451/

PR #440, opened 2019-02-26
fix: Remove obsolete notifications

mazen-asef:
Local method is not being found without the <this>
identifier, so we use it.

https://phabricator.endlessm.com/T25246


mazen-asef:
When first merged it was working, but on a separate branch where I rebased it, it was not being able to find the function inside the class scope, for some reason.

mazen-asef:
@GeorgesStavracas Fixed the commit, thanks for the help :)

PR #445, opened 2019-03-22
codeView: remove CodingSession on actor destroy

danigm:
When a hack app is destroyed when it's minimized the corresponding
CodingSession is not removed from the list because the function
handleDestroyWindow is never called and this cause problems in the
future because there's a CodingSession with an app that points to a non
existing window.

To fix this we need to listen the actor destroy event and remove the
CodingSession if this happens.

https://phabricator.endlessm.com/T25880

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -381,6 +381,10 @@ var CodingSession = new Lang.Class({
                                             GObject.ParamFlags.READWRITE,
                                             Meta.WindowActor),
     },
+    Signals: {
+        'minimized': {},
+        'unminimized': {},
+    },
 
     _init: function(params) {
         this._app = null;
@@ -843,7 +847,7 @@ var CodingSession = new Lang.Class({
         // If we have an app window, disconnect any signals and destroy it,
         // unless we are destroying the session because the app window was
         // destroyed in the first place
-        if (this.app) {
+        if (this.app && !this.app.is_destroyed()) {
             let appWindow = this.app.meta_window;
             this.app = null;
             this._shellApp = null;
@@ -855,7 +859,7 @@ var CodingSession = new Lang.Class({
         // If we have a toolbox window, disconnect any signals and destroy it,
         // unless we are destroying the session because the toolbox window was
         // destroyed in the first place
-        if (this.toolbox) {
+        if (this.toolbox && !this.toolbox.is_destroyed()) {
             let toolboxWindow = this.toolbox.meta_window;
             this.toolbox = null;
 
@@ -976,6 +980,8 @@ var CodingSession = new Lang.Class({
         // Only want to minimize if we weren't already minimized.
         if (toMini && !toMini.meta_window.minimized)
             toMini.meta_window.minimize();
+
+        this.emit('minimized');
     },
 
     _applyWindowUnminimizationState: function(shellwm, actor) {
@@ -993,6 +999,8 @@ var CodingSession = new Lang.Class({
         // minimized.
         if (toUnMini && toUnMini.meta_window.minimized)
             toUnMini.meta_window.unminimize();
+
+        this.emit('unminimized');
     },
 
     _overviewStateChanged: function() {
@@ -1261,7 +1269,37 @@ var CodeViewManager = new Lang.Class({
     },
 
     _addSession: function(actor) {
-        this._sessions.push(new CodingSession({ app: actor }));
+        let session = new CodingSession({ app: actor });
+
+        // When the app is minimized the WM doesn't emit the destroy signal if
+        // the app is closed, for this reason we need to listen to the 'destroy'
+        // signal of the actor if it's minimized, to be able to remove from the
+        // CodingSession list and avoid possible gnome-shell crash
+        let destroyAppHandlerId = 0;
+        let destroyToolboxHandlerId = 0;
+        session.connect('minimized', (session) => {
+            if (session.app) {
+                destroyAppHandlerId = session.app.connect('destroy',
+                    this.handleDestroyWindow.bind(this));
+            }
+
+            if (session.toolbox) {
+                destroyToolboxHandlerId = session.toolbox.connect('destroy',
+                    this.handleDestroyWindow.bind(this));
+            }
+        });
+        session.connect('unminimized', (session) => {
+            if (destroyAppHandlerId) {
+                session.app.disconnect(destroyAppHandlerId);
+                destroyAppHandlerId = 0;
+            }
+            if (destroyToolboxHandlerId) {
+                session.toolbox.disconnect(destroyToolboxHandlerId);
+                destroyToolboxHandlerId = 0;
+            }
+        });
+
+        this._sessions.push(session);
     },
 
     _removeSession: function(session, eventType) {


cosimoc:
We discussed with @danigm on Slack that I'd like the unminimize/minimize state to be driven by `windowManager.js` in the same way it happens for the `map` and `destroy` signals at the moment.

js/ui/codeView.js
```diff
@@ -1187,6 +1187,7 @@ var CodeViewManager = new Lang.Class({
     },
 
     _addSession: function(actor) {
+        actor.connect('destroy', this.handleDestroyWindow.bind(this));
```
cosimoc: I have two main questions about this patch:
- why would we not be interested in the toolbox actor destroy events as well?
- more importantly, what happens when a window is unminimized? Consider the use case where a hackable app minimized, then unminimized again (or to add more complexity, even flipped before the first minimization, then flipped back). Would we not end up without a corresponding CodingSession for that hackable app/toolbox, with this patch?

danigm: About the toolbox actor, I think it's not needed because the only actor that can be destroyed when minimized is the app actor.

This change don't remove the coding session on minimize, it only binds the destroy callback to the actor destroy event for the case when the shell doesn't notify the codeviewmanager about a window being destroyed (that's the case when it's minimized). So the only problem with this is the callback being called twice, but in that case, the callback won't do anything because the condingsession won't be there.

cosimoc: > About the toolbox actor, I think it's not needed because the only actor that can be destroyed when minimized is the app actor.

I'm not sure what you mean by this, do you see that the application actor emits a destroy signal when minimized, but the toolbox actor does not do that?

> This change don't remove the coding session on minimize, it only binds the destroy callback to the actor destroy event for the case when the shell doesn't notify the codeviewmanager about a window being destroyed (that's the case when it's minimized). So the only problem with this is the callback being called twice, but in that case, the callback won't do anything because the condingsession won't be there.

Now I am a bit confused here. I thought that the issue is that upon minimization, the window actor is destroyed, but we were keeping the CodingSession alive. So your fix here connects to that destroy signal to remove the CodingSession when that happens. What is the purpose of connecting the callback if not to remove the session? (And you also say it yourself, the second time the callback is called it won't do anything because the session was already removed.)

danigm: > > About the toolbox actor, I think it's not needed because the only actor that can be destroyed when minimized is the app actor.
> 
> I'm not sure what you mean by this, do you see that the application actor emits a destroy signal when minimized, but the toolbox actor does not do that?

What I'm saying is that when you close when minimized, the app actor will emit the destroy signal, so it's not needed to listen to the destroy signal of the toolbox to remove the coding session, the actor destroy signal will do the work.

> > This change don't remove the coding session on minimize, it only binds the destroy callback to the actor destroy event for the case when the shell doesn't notify the codeviewmanager about a window being destroyed (that's the case when it's minimized). So the only problem with this is the callback being called twice, but in that case, the callback won't do anything because the condingsession won't be there.
> 
> Now I am a bit confused here. I thought that the issue is that upon minimization, the window actor is destroyed, but we were keeping the CodingSession alive. So your fix here connects to that destroy signal to remove the CodingSession when that happens. What is the purpose of connecting the callback if not to remove the session? (And you also say it yourself, the second time the callback is called it won't do anything because the session was already removed.)

The real problem is when we close the app when it's minimized. Minimize and unminimize doesn't affect at all and works correctly.

But if you minimize and then right click and close the app, when it's minimized, the actual code doesn't call to the handleDestroyWindow, the shell wm doesn't emit the "destroy" signal, so we have a codingSession in the list pointing to a destroyed app actor and when we try to do "session.app" the shell crashes.

That's why listening to the actor destroy signal fixes this, because that signal is emitted always, no matter how the actor is being destroyed.

cosimoc: Again apologies that this took so long to reply to!

> What I'm saying is that when you close when minimized, the app actor will emit the destroy signal, so it's not needed to listen to the destroy signal of the toolbox to remove the coding session, the actor destroy signal will do the work.

I would expect that to depend on whether the app or the toolbox is "in front". Due to `_applyWindowMinimizationState()` when one window is minimized, both are.  But e.g. in the overview window picker and on the taskbar menu, we always show the window in front, to reinforce the illusion that those are two sides of the same window.

So if you e.g. minimize the app while the toolbox side is showing, then close it by clicking on the X next to it in the window picker, I expect the destroy signal to reach the toolbox actor, not the app one.

> The real problem is when we close the app when it's minimized. Minimize and unminimize doesn't affect at all and works correctly.
> 
> But if you minimize and then right click and close the app, when it's minimized, the actual code doesn't call to the handleDestroyWindow, the shell wm doesn't emit the "destroy" signal, so we have a codingSession in the list pointing to a destroyed app actor and when we try to do "session.app" the shell crashes.
> 
> That's why listening to the actor destroy signal fixes this, because that signal is emitted always, no matter how the actor is being destroyed.

Thanks, things make a lot more sense after your explanation and I now understand why you prefer this patch over the other PR! :-)
I didn't realize that we would not receive a signal from the WindowManager class while windows are minimized.

Would it be possible for you to change this patch so that the Clutter destroy signal only stays connected during the time we know we won't get one from the WM? (e.g. only between the time the window is minimized and unminimized)
I think that would address the last concern I have about this.

danigm: I've updated the code to only listen to the destroy event when the window is minimized. I needed to add a signal to be able to know if the `CodingSession` actor is minimized in the `CodeViewManager` because we need to remove the `CodingSession` from the `CodeViewManager` object.

js/ui/codeView.js
```diff
@@ -1261,7 +1269,32 @@ var CodeViewManager = new Lang.Class({
...
-        this._sessions.push(new CodingSession({ app: actor }));
+        const session = new CodingSession({ app: actor });
+
+        // When the app is minimized the WM doesn't emit the destroy signal if
+        // the app is closed, for this reason we need to listen to the 'destroy'
+        // signal of the actor if it's minimized, to be able to remove from the
+        // CodingSession list and avoid possible gnome-shell crash
+        let destroyAppHandlerId = 0;
+        let destroyToolboxHandlerId = 0;
+        session.connect('minimized', (session) => {
```
cosimoc: Inside this callback you’ll need to check whether session.app and session.toolbox exist before connecting the signals... especially the toolbox may not exist yet at this point.

js/ui/codeView.js
```diff
@@ -976,6 +980,8 @@ var CodingSession = new Lang.Class({
         // Only want to minimize if we weren't already minimized.
         if (toMini && !toMini.meta_window.minimized)
             toMini.meta_window.minimize();
+
+        this.emit('minimized', actor);
```
cosimoc: This and the other signal pass the actor but the signals are defined without parameters and you don’t use the actor in the callback, so should be removed

js/ui/codeView.js
```diff
@@ -843,7 +847,7 @@ var CodingSession = new Lang.Class({
         // If we have an app window, disconnect any signals and destroy it,
         // unless we are destroying the session because the app window was
         // destroyed in the first place
-        if (this.app) {
+        if (this.app && !this.app.is_destroyed()) {
```
cosimoc: Not sure if this change is required anymore now that we avoid having a second handleDestroyWindow callback running?

js/ui/codeView.js
danigm: Removed, that's there because I changed the signal and didn't update that code.

js/ui/codeView.js
danigm: Sure, I've added the check here.

js/ui/codeView.js
danigm: This is needed and I've to add the same to the toolbox  because if we don't do this, when we close when the app is minimized, this function is called with the app/toolbox destroyed and the `meta_window` will be null and the following code will fail.

js/ui/codeView.js
```diff
@@ -1261,7 +1269,39 @@ var CodeViewManager = new Lang.Class({
...
+            }
+
+            if (session.toolbox) {
+                destroyToolboxHandlerId = session.toolbox.connect('destroy',
+                    this.handleDestroyWindow.bind(this));
+            }
+        });
+        session.connect('unminimized', (session) => {
+            if (destroyAppHandlerId) {
+                if (session.app)
```
cosimoc: I wonder if this check is actually required, since `destroyAppHandlerId` only is set if `session.app` is set in the other callback. (Same below.)

cosimoc:
Looks great to me. @GeorgesStavracas Want to give this a quick ACK before we merge it?

js/ui/codeView.js
```diff
@@ -1261,7 +1269,35 @@ var CodeViewManager = new Lang.Class({
     },
 
     _addSession: function(actor) {
-        this._sessions.push(new CodingSession({ app: actor }));
+        const session = new CodingSession({ app: actor });
```
GeorgesStavracas: Why `const`? Shouldn't it be `let`?

js/ui/codeView.js
```diff
@@ -1261,7 +1269,35 @@ var CodeViewManager = new Lang.Class({
...
+
+        // When the app is minimized the WM doesn't emit the destroy signal if
+        // the app is closed, for this reason we need to listen to the 'destroy'
+        // signal of the actor if it's minimized, to be able to remove from the
+        // CodingSession list and avoid possible gnome-shell crash
+        let destroyAppHandlerId = 0;
+        let destroyToolboxHandlerId = 0;
+        session.connect('minimized', (session) => {
+            if (session.app)
```
GeorgesStavracas: Because the if body has more than 1 line, add curly braces.

js/ui/codeView.js
```diff
@@ -1261,7 +1269,35 @@ var CodeViewManager = new Lang.Class({
...
+                destroyAppHandlerId = session.app.connect('destroy',
+                    this.handleDestroyWindow.bind(this));
+
+            if (session.toolbox)
```
GeorgesStavracas: Here too

cosimoc:
(I had asked Daniel to remove curly braces for those blocks since they’re a single condition... sorry about the back and forth!)

GeorgesStavracas:
> (I had asked Daniel to remove curly braces for those blocks since they’re a single condition... sorry about the back and forth!)

No worries. Had I reviewed this branch 2 weeks ago, I would have asked too. But apparently upstream follow the "1 line body" rule.

PR #447, opened 2019-03-27
codeView: Monitor changes on hack-data file

danigm:
https://phabricator.endlessm.com/T26045

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -81,6 +81,7 @@ function _getToolboxTarget(win) {
 const _ensureHackDataFile = (function () {
     let keyfile = new GLib.KeyFile();
     let initialized = false;
+    let monitors = [];
 
     const flatpakInstallationPaths = [
         GLib.build_filenamev([GLib.get_home_dir(), '.local/share/flatpak']),
@@ -94,6 +95,25 @@ const _ensureHackDataFile = (function () {
     return function() {
         if (initialized)
             return keyfile;
+
+        // Only create file monitors the first time
+        if (monitors.length === 0) {
+            for (let path of searchPaths) {
+                path = GLib.build_filenamev([path, 'hack-data.ini']);
+                const file = Gio.file_new_for_path(path);
+                const monitor = file.monitor_file(Gio.FileMonitorFlags.NONE, null);
+                monitor.connect('changed', (_monitor, _file, _otherfile, ev, _data) => {
+                    if (ev === Gio.FileMonitorEvent.CHANGES_DONE_HINT) {
+                        // forces the keyfile reload the next time the function is called
+                        keyfile = new GLib.KeyFile();
+                        initialized = false;
+                    }
+                });
+                // we keep a reference to the monitor object to avoid destroy and
+                // signal disconnection
+                monitors.push(monitor);
+            }
+        }
 
         try {
             keyfile.load_from_dirs('hack-data.ini', searchPaths,


js/ui/codeView.js
```diff
@@ -96,8 +98,16 @@ const _ensureHackDataFile = (function () {
             return keyfile;
 
         try {
-            keyfile.load_from_dirs('hack-data.ini', searchPaths,
+            const [, path] = keyfile.load_from_dirs('hack-data.ini', searchPaths,
                 GLib.KeyFileFlags.NONE);
+            monitorFile = Gio.file_new_for_path(path);
+            monitor = monitorFile.monitor_file(Gio.FileMonitorFlags.NONE, null);
+            monitor.connect('changed', (_monitor, _file, _otherfile, ev, _data) => {
```
pwithnall: Shouldn’t this signal handler be disconnected at some point?

js/ui/codeView.js
```diff
@@ -96,8 +98,16 @@ const _ensureHackDataFile = (function () {
             return keyfile;
 
         try {
-            keyfile.load_from_dirs('hack-data.ini', searchPaths,
+            const [, path] = keyfile.load_from_dirs('hack-data.ini', searchPaths,
```
pwithnall: Technically you need to monitor all the paths in `searchPaths`, otherwise the config could originally be loaded from a directory low down in the priority order, then overridden by one in a different (higher priority) directory.

js/ui/codeView.js
```diff
@@ -96,8 +98,16 @@ const _ensureHackDataFile = (function () {
...
                 GLib.KeyFileFlags.NONE);
+            monitorFile = Gio.file_new_for_path(path);
+            monitor = monitorFile.monitor_file(Gio.FileMonitorFlags.NONE, null);
```
pwithnall: If the config file is ever going to be modified from an OSTree commit, you might also want `Gio.FileMonitorFlags.WATCH_HARD_LINKS`, I suspect?

js/ui/codeView.js
danigm: This will exists until the end of the gnome shell, has the same scope as the keyfile

js/ui/codeView.js
danigm: We're not reloading the keyfile, this path is calculated the first time, so if we monitor all the paths we should recalculate this path on every change.

That will be a bit harder to implement. In that case maybe it's easier to monitor all files and when something changes set the keyfile to null and disconnect all monitors so the next time it's requested it'll be calculated.

js/ui/codeView.js
danigm: done

wjt:
@danigm since @GeorgesStavracas has force-pushed to master, please rebase this branch.

js/ui/codeView.js
```diff
@@ -96,8 +98,16 @@ const _ensureHackDataFile = (function () {
             return keyfile;
 
         try {
-            keyfile.load_from_dirs('hack-data.ini', searchPaths,
+            const [, path] = keyfile.load_from_dirs('hack-data.ini', searchPaths,
                 GLib.KeyFileFlags.NONE);
+            monitorFile = Gio.file_new_for_path(path);
+            monitor = monitorFile.monitor_file(Gio.FileMonitorFlags.WATCH_HARD_LINKS, null);
```
wjt: Have you tested that this actually works when you update the Flatpak? The full path you are monitoring is:

```
/var/lib/flatpak/app/com.endlessm.HackComponents/current/active/files/share/hack-components/hack-data.ini
                                                 ^^^^^^^ ^^^^^^
```

This file will never be overwritten in place, nor replaced in the same directory. The components indicated with `^^^` are symbolic links. When a new version of the app is installed, a new deployment is created and the symbolic link(s) updated.

Does GFileMonitor monitor all parent directories of a file?

danigm: Yes, I've tested downgrading the flatpak app, launching the lightspeed and upgrading the HackComponents again and the change is detected, after the update I see the FtH button in the lightspeed app. The monitor is detecting this particular change in the file.

danigm:
Okay, I've removed the flag because it crashes the shell. The current patch is working with the default update flow.

js/ui/codeView.js
danigm: I've changed the code to monitor all files.

wjt:
I stand corrected. I tested `gio monitor -f /var/lib/flatpak/app/com.endlessm.HackComponents/current/active/files/share/hack-components/hack-data.ini` + `flatpak update com.endlessm.HackComponents`, and it does notice the file being deleted, created, then "changes done".

I think what I was missing from my test is that, after swapping the symlinks, Flatpak deletes the old deployment (because the app is not running). What seems to be happening is that [when a watched file is deleted](https://gitlab.gnome.org/GNOME/glib/blob/master/gio/inotify/inotify-path.c#L568), GLib adds it to a list of "missing" files that gets [polled for existence every 4 seconds](https://gitlab.gnome.org/GNOME/glib/blob/master/gio/inotify/inotify-missing.c). So this works only because HackComponents is not actually an app, so Flatpak never keeps the old deployment around. Tricky! If I delete the old symlink target in my isolated test case, the JS code notices the delete immediately, the create a few seconds later, and then changes_done as hoped for.

js/ui/codeView.js
wjt: See below. Both parts have been resolved.

js/ui/codeView.js
```diff
@@ -95,6 +96,24 @@ const _ensureHackDataFile = (function () {
...
+        for (let path of searchPaths) {
+            path = GLib.build_filenamev([path, 'hack-data.ini']);
+            const file = Gio.file_new_for_path(path);
+            const monitor = file.monitor_file(Gio.FileMonitorFlags.NONE, null);
+            monitor.connect('changed', (_monitor, _file, _otherfile, ev, _data) => {
+                if (ev === Gio.FileMonitorEvent.CHANGES_DONE_HINT) {
+                    // forces the keyfile reload the next time the function is called
+                    keyfile = new GLib.KeyFile();
+                    initialized = false;
+                    // disconnects all the previous monitors
```
wjt: a. does it?
b. why is this desirable? Surely at the next call to the surrounding function, exactly the same set of monitors will be set up?

danigm: Yes, you're right, we don't need to destroy/create again. I've updated the PR to create the monitors only the first time.

js/ui/codeView.js
wjt: We discussed this a bit on the desktop daily. I agree it is good practice to disconnect signal handlers, even if they will last for the lifetime of an application. However, in this case we'll probably be okay: the callback just assigns to some variables in its closure, and we assume that gjs will keep that closure alive until process termination. I guess we'll find out the hard way if we're wrong.

js/ui/codeView.js
```diff
@@ -95,6 +96,25 @@ const _ensureHackDataFile = (function () {
         if (initialized)
             return keyfile;
 
+        // Only create file monitors the first time
+        if (monitors.length === 0) {
```
wjt: Rather than checking this, you can just move this block up to just before `return function() {`.

danigm: We can't, if we do that, the monitors variable doesn't have any reference and all monitors are destroyed. I tried and it takes me a while to understand why it was not working.

If we want to do that outside we need to move the monitors variable to the global scope

wjt: This seems really fragile: is it actually guaranteed that variables in the outer function scope which are not referenced from the inner function stay alive?

wjt: But, OK.

PR #448, opened 2019-03-29
codeView: remove CodingSession on actor minimize

danigm:
This is an alternative solution to this PR: https://github.com/endlessm/gnome-shell/pull/445

When a hack app is destroyed when it's minimized the corresponding
CodingSession is not removed from the list because the function
handleDestroyWindow is never called and this cause problems in the
future because there's a CodingSession with an app that points to a non
existing window.

This patch links with the minimize/unimimize events and remove/add the
CodingSession to the list.

To avoid to leave some toolbox windows over there, before the minimizing
we need to flip back and then the session can be removed without problems.

https://phabricator.endlessm.com/T25880

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -770,7 +770,7 @@ var CodingSession = new Lang.Class({
         // If we have an app window, disconnect any signals and destroy it,
         // unless we are destroying the session because the app window was
         // destroyed in the first place
-        if (this.app) {
+        if (this.app && !this.app.is_destroyed()) {
             let appWindow = this.app.meta_window;
             this.app = null;
             this._shellApp = null;
@@ -856,6 +856,22 @@ var CodingSession = new Lang.Class({
                                  this.app,
                                  null, this.toolbox,
                                  Gtk.DirectionType.LEFT);
+        }
+    },
+
+    flipBackMinimize: function() {
+        if (this._toolboxActionGroup.has_action('flip-back')) {
+            this._toolboxActionGroup.activate_action('flip-back', null);
+            this._appRemovedActor = this.app;
+            this._minimize = true;
+            this.app = null;
+        } else {
+            this._setState(STATE_APP, false);
+            this.app.meta_window.activate(global.get_current_time());
+            this.app.opacity = 255;
+            this.app.rotation_angle_y = 0;
+            this._setEffectsEnabled(this.app, false);
+            this.app.meta_window.minimize();
         }
     },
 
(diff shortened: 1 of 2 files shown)


js/ui/codeView.js
```diff
@@ -1201,6 +1217,24 @@ var CodeViewManager = new Lang.Class({
...
+            return;
+
+        // First, determine if this is an app window getting destroyed
+        let session = this._getSession(actor, SessionLookupFlags.SESSION_LOOKUP_APP);
+        let eventType = SessionDestroyEvent.SESSION_DESTROY_APP_DESTROYED;
+
+        if (session)
+            this._removeSession(session, eventType);
+        else {
+            // if we're minimizing as toolbox, we switch back
```
cosimoc: This is not really the desired behavior though...

The primary goal of the FtH implementation is to maintain an illusion that the front and back of an app/toolbox pair in a session really function as a single entity.
So, when the app is minimized, if there's a toolbox in the session, it should minimize together with it; when the app is unminimized, the toolbox should also unminimize with it, and viceversa. Again, the effect we want to create is that those are really the same physical window.

That's what the `applyWindowMinimizationState()` functions are trying to achieve here: https://github.com/endlessm/gnome-shell/blob/master/js/ui/codeView.js#L892-L923

js/ui/codeView.js
```diff
@@ -770,7 +770,7 @@ var CodingSession = new Lang.Class({
         // If we have an app window, disconnect any signals and destroy it,
         // unless we are destroying the session because the app window was
         // destroyed in the first place
-        if (this.app) {
+        if (this.app && !this.app.is_destroyed()) {
```
cosimoc: I don't know that this is required here, if we go with this approach.

js/ui/codeView.js
danigm: I did this workaround because in other case, we can't remove the CodingSession on minimize status, if we don't flip back, the app will be fliped when unminimize and the CodingSession has been removed.

So I think that if we don't flip back we can't remove the CodingSession on minimize because that will break the link and destroy the toolbox.

I'll try to handle this in the handleMapWindow, and avoid the toolbox destroy on minimize, to avoid the flip back.

cosimoc: Sounds good. Yes, if the app was flipped when it was minimized (i.e. it was actually the toolbox to be minimized) upon unminimization should still be flipped, and viceversa.
Basically minimization should not have any impact on the coding sessions, if at all possible.

danigm:
Closing this PR, we're going with a better solution here: https://github.com/endlessm/gnome-shell/pull/445

PR #449, opened 2019-04-02
eos-payg: Add error handling for time removal

mazen-asef:
Adding a code that does not add time for the users makes
the paygManager to try and trigger the screenShield, and the
signal is received, but does not make it pop since it's not
unlocked, finishing the unlock process before adding a valid
code. So we add a error handling using the time added from the
daemon so we know how much time has been returned, and if it's
zero, do nothing.

https://phabricator.endlessm.com/T25248

### DIFF
--- a/js/misc/paygManager.js
+++ b/js/misc/paygManager.js
@@ -40,6 +40,7 @@ const EOS_PAYG_IFACE = '<node> \
 <interface name="com.endlessm.Payg1"> \
 <method name="AddCode"> \
   <arg type="s" direction="in" name="code"/> \
+  <arg type="x" direction="out" name="time_added"/> \
 </method> \
 <method name="ClearCode" /> \
 <signal name="Expired" /> \
@@ -92,6 +93,7 @@ var PaygManager = new Lang.Class({
 
         this._enabled = false;
         this._expiryTime = 0;
+        this._lastTimeAdded = 0;
         this._rateLimitEndTime = 0;
         this._codeFormat = '';
         this._codeFormatRegex = null;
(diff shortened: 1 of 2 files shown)


mazen-asef:
Also, this PR cannot be merged without:
https://github.com/endlessm/eos-payg/pull/31
https://github.com/endlessm/eos-payg-nonfree/pull/8

js/misc/paygManager.js
```diff
@@ -92,6 +93,7 @@ var PaygManager = new Lang.Class({
 
         this._enabled = false;
         this._expiryTime = 0;
+        this._lastTimeAdded = 0;
```
treitter: You should add a getter to encapsulate this private variable. See `get rateLimitEndTime()`.

js/ui/payg.js
```diff
@@ -210,6 +217,8 @@ var UnlockUi = new Lang.Class({
 
             if (error) {
                 this.processError(error);
+            } else if (Main.paygManager._lastTimeAdded >= 0) {
```
treitter: Cut out the `_` to access this via its getter. See my comment above.

wjt:
Please update the PR description and commit message to reference https://phabricator.endlessm.com/T24128 rather than https://phabricator.endlessm.com/T25248.

wjt:
(and I think you'll want to add that patch to the set of patches for https://phabricator.endlessm.com/T24128 ?)

mazen-asef:
@wjt I will close this PR and move the changes to https://github.com/endlessm/gnome-shell/pull/398 since a good chunk of those changes are here as well

PR #450, opened 2019-04-02
codeView: Hide toolbox and FtH on wobble movement

danigm:
https://phabricator.endlessm.com/T25788

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -405,6 +405,11 @@ var CodingSession = new Lang.Class({
         this._hackablePropsChangedId = 0;
         this._hackable = true;
 
+        this._grabbed = false;
+        this._backParent = null;
+        this._backConstraint = null;
+        this._grabTimeoutId = 0;
+
         this.parent(params);
 
         // FIXME: this should be extended to make it possible to launch
@@ -451,6 +456,69 @@ var CodingSession = new Lang.Class({
 
     get toolbox() {
         return this._toolbox;
+    },
+
+    setGrabbed: function(grabbed) {
+        if (this._grabTimeoutId > 0) {
+            Mainloop.source_remove(this._grabTimeoutId);
+            this._grabTimeoutId = 0;
+        }
+
+        if (grabbed) {
+            this._attachBackendWindow();
+            this._grabbed = true;
+            this._syncButtonVisibility();
+        } else {
+            // giving some time to finish possible window movement FXs
+            this._grabTimeoutId = Mainloop.timeout_add(500,
+                () => {
+                    Mainloop.source_remove(this._grabTimeoutId);
+                    this._grabTimeoutId = 0;
+
+                    this._detachBackendWindow();
+                    this._grabbed = false;
+                    this._syncButtonVisibility();
+                    return GLib.SOURCE_REMOVE;
+                });
+        }
+    },
+
+    _attachBackendWindow: function() {
+        const frontActor = this._actorForCurrentState();
+        const backActor = this._getOtherActor(frontActor);
+
+        if (!backActor)
+            return;
+
+        this._backParent = backActor.get_parent();
+        this._backParent.remove_child(backActor);
+        frontActor.insert_child_at_index(backActor, 0);
+
+        // Constraint to avoid inside actor movement
+        const path = new Clutter.Path();
+        path.add_move_to(0, 0);
+        this._backConstraint = new Clutter.PathConstraint({path: path, offset: 0});
+        backActor.add_constraint(this._backConstraint);
+    },
+
+    _detachBackendWindow: function() {
+        const frontActor = this._actorForCurrentState();
+        const backActor = this._getOtherActor(frontActor);
+
+        if (!this._backParent || !backActor)
+            return;
+
+        if (this._backConstraint) {
+            backActor.remove_constraint(this._backConstraint);
+            this._backConstraint = null;
+        }
+
+        const [x, y] = frontActor.get_position();
+        backActor.set_position(x, y);
+
+        frontActor.remove_child(backActor);
+        this._backParent.insert_child_below(backActor, frontActor);
+        this._backParent = null;
     },
 
     _ensureButton: function() {
@@ -767,6 +835,11 @@ var CodingSession = new Lang.Class({
             this._hackablePropsChangedId = 0;
         }
 
+        if (this._grabTimeoutId !== 0) {
+            Mainloop.source_remove(this._grabTimeoutId);
+            this._grabTimeoutId = 0;
+        }
+
         // If we have an app window, disconnect any signals and destroy it,
         // unless we are destroying the session because the app window was
         // destroyed in the first place
@@ -950,6 +1023,7 @@ var CodingSession = new Lang.Class({
             !Main.overview.visible &&
             !inFullscreen &&
             !locked &&
+            !this._grabbed &&
             this._hackable) {
 
             if (!this.toolbox) {
@@ -1226,6 +1300,16 @@ var CodeViewManager = new Lang.Class({
             this._removeSession(session, eventType);
 
         return false;
+    },
+
+    handleWindowGrab: function(actor, grabbed) {
+        let session = this._getSession(actor,
+                                       SessionLookupFlags.SESSION_LOOKUP_APP |
+                                       SessionLookupFlags.SESSION_LOOKUP_TOOLBOX);
+        if (!session)
+            return;
+
+        session.setGrabbed(grabbed);
     },
 
     handleMapWindow: function(actor) {
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -2940,6 +2940,7 @@ var WindowManager = new Lang.Class({
 
         if (attachedEffect)
             attachedEffect.activate('move', { grabbed: true });
+        this._codeViewManager.handleWindowGrab(actor, true);
     },
 
     _windowUngrabbed: function(display, op, window) {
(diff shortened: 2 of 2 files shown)


js/ui/codeView.js
```diff
@@ -453,6 +456,44 @@ var CodingSession = new Lang.Class({
         return this._toolbox;
     },
 
+    setGrabbed: function(grabbed) {
+        if (this._button)
+            this._button.opacity = grabbed ? 0 : 255;
+
+        if (this.toolbox && this._state == STATE_APP)
+            this.toolbox.opacity = grabbed ? 0 : 255;
```
joaquimrocha: **Nitpicky-pick:** I'd create a var with the opacity value just not to repeat those two lines, i.e.
`let opacity = grabbed ? 0 : 255;`

As indicated, this is a nitpick, so feel free to change it only if you think it's worth it.

js/ui/codeView.js
```diff
@@ -405,6 +405,9 @@ var CodingSession = new Lang.Class({
         this._hackablePropsChangedId = 0;
         this._hackable = true;
 
+        this._appParent = null;
+        this._appPositionId = 0;
```
joaquimrocha: **Nitpick:** When I saw this, I was wondering what `position ID` was; it'd be a bit more clear if it were called `this._appPositionHandler` or `this._appPositionHandlerId`. But it's a mere suggestion (really up to you).

js/ui/codeView.js
```diff
@@ -453,6 +456,44 @@ var CodingSession = new Lang.Class({
...
+
+    _attachAppToToolbox: function() {
+        this._appParent = this.app.get_parent();
+        this._appParent.remove_child(this.app);
+        this.toolbox.insert_child_at_index(this.app, 0);
+        this.app.set_position(0, 0);
+        this._appPositionId = this.app.connect('notify::position', (actor) => {
+            const [x, y] = actor.get_position();
+            if (x || y)
+                actor.set_position(0, 0);
```
joaquimrocha: The reason for this seems a bit non-trivial (at least to me), can you explain (and if you agree it's not obvious, add a comment) why this is needed? Maybe to prevent the actor changing its position by itself when it's contained in the parent?

js/ui/codeView.js
```diff
@@ -453,6 +456,44 @@ var CodingSession = new Lang.Class({
...
+    },
+
+    _attachAppToToolbox: function() {
+        this._appParent = this.app.get_parent();
+        this._appParent.remove_child(this.app);
+        this.toolbox.insert_child_at_index(this.app, 0);
+        this.app.set_position(0, 0);
+        this._appPositionId = this.app.connect('notify::position', (actor) => {
+            const [x, y] = actor.get_position();
+            if (x || y)
```
joaquimrocha: I was trying to understand why you're checking the position here, instead of just calling `set_position` below, and I suspect it's because otherwise this would end up in an infinite call of `set_position`, as we're triggering the notify from the method we use to connect to it.

Thus, I think this is one of those cases where a comment may avoid future problems of "let me remove these lines and just call `set_position(0, 0)` directly".

js/ui/codeView.js
```diff
@@ -453,6 +456,44 @@ var CodingSession = new Lang.Class({
...
+                actor.set_position(0, 0);
+        });
+    },
+
+    _detachAppFromToolbox: function() {
+        if (!this._appParent || !this.toolbox) {
+            return;
+        }
```
joaquimrocha: Assuming the convention is to not use curly brackets in one-liner if statements (like you have in the beginning of this patch), then this should be removed.

js/ui/codeView.js
```diff
@@ -886,6 +929,7 @@ var CodingSession = new Lang.Class({
             return;
 
         let actor = window.get_compositor_private();
+
```
joaquimrocha: **Nitpick:** This may have been left-over unintentionally.

js/ui/codeView.js
danigm: We need this here because we're syncing the window position and after adding as a child we don't need to move the position of the child, so this is done to ensure that the child actor will be in the correct possition without modifying the sync function. I'll add a comment as you suggested

js/ui/codeView.js
```diff
@@ -453,6 +457,63 @@ var CodingSession = new Lang.Class({
...
+            Mainloop.source_remove(this._grabTimeoutId);
+            this._grabTimeoutId = 0;
+        }
+
+        if (!this._button)
+            return;
+
+        if (grabbed)
+            this._button.opacity = 0;
+        else {
```
joaquimrocha: I'm not sure what's the convention here (I had the impression that as long as you have a block with curly brackets, the other should have them too). So let's wait for someone else to clarify that.

danigm: I think you're right about this convention, I've configured the eslint for the shell and it tells me that I should put brackets there. I've updated it.

js/ui/codeView.js
```diff
@@ -453,6 +457,63 @@ var CodingSession = new Lang.Class({
...
+            // giving some time to finish possible window movement FXs
+            this._grabTimeoutId = Mainloop.timeout_add(300,
+                () => {
+                    this._button.opacity = 255;
+                    return GLib.SOURCE_REMOVE;
+                });
+        }
+    },
+
+    _attachAppToToolbox: function() {
```
cosimoc: This is a quite neat trick! However, from what I can see your change always sets up the app window actor to be a child of the toolbox while the toolbox is displayed.
I am a bit concerned that will break some expectation elsewhere that window actors are children of the window group.

On the other hand, if this does work reliably I think that we could remove a lot of the window position tracking/synchronization code that we have elsewhere in `codeView.js`, but that's probably besides the point of this PR.

Perhaps for now we could set up the nested hierarchy only while the window is being dragged?

js/ui/codeView.js
```diff
@@ -453,6 +457,63 @@ var CodingSession = new Lang.Class({
...
+        this._appParent = this.app.get_parent();
+        this._appParent.remove_child(this.app);
+        this.toolbox.insert_child_at_index(this.app, 0);
+        this.app.set_position(0, 0);
+        // As a child actor we need to ensure that the position is always (0, 0)
+        // to avoid the rendering outside the parent window or with an offset.
+        // We've code to sync both windows so to avoid glitches and actor
+        // movements inside the parent we force the position listening to the
+        // position change event.
+        this._appPositionHandlerId = this.app.connect('notify::position', (actor) => {
```
cosimoc: Have you considered using a `ClutterConstraint` to achieve this instead?
I'm also curious to know how this code interacts with the position tracking code we have elsewhere for coding session windows; ideally we would have a single place where we adjust positions.

js/ui/codeView.js
```diff
@@ -453,6 +457,63 @@ var CodingSession = new Lang.Class({
...
+        if (this._grabTimeoutId > 0) {
+            Mainloop.source_remove(this._grabTimeoutId);
+            this._grabTimeoutId = 0;
+        }
+
+        if (!this._button)
+            return;
+
+        if (grabbed) {
+            this._button.opacity = 0;
```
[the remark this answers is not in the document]
danigm: Done in the `_syncButtonVisibility`

js/ui/codeView.js
```diff
@@ -1142,6 +1216,11 @@ var CodingSession = new Lang.Class({
         actor.rotation_angle_y = 0;
         actor.opacity = 255;
         this._rotatingInActor = null;
+
+        if (actor === this.toolbox)
+            this._attachAppToToolbox();
+        else
+            this.toolbox.opacity = 0;
```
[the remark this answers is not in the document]
danigm: Done, now the attach is done for both windows.

js/ui/codeView.js
danigm: The first version of the path does that, but I changed to do it after the flip. I've implemented the original version again so now the attach is only done during the grab.

js/ui/codeView.js
danigm: I tried to do avoid the window sync in the existing function, but for some reason that cause some movement glitches, that's because I was using the `notify`.

I didn't know about `ClutterConstraint` before, but now I've used a `ClutterConstraint` to do this and it works like a charm.

js/ui/codeView.js
```diff
@@ -453,6 +467,67 @@ var CodingSession = new Lang.Class({
...
+
+        this._backConstraint = new InsideActorConstraint();
+        backActor.add_constraint(this._backConstraint);
+    },
+
+    _detachBackendWindow: function() {
+        const frontActor = this._actorForCurrentState();
+        const backActor = this._getOtherActor(frontActor);
+
+        if (!this._backParent || !backActor) {
```
cosimoc: Remove curlies here

js/ui/codeView.js
```diff
@@ -366,6 +366,15 @@ const SessionDestroyEvent = {
     SESSION_DESTROY_TOOLBOX_DESTROYED: 1,
 };
 
+var InsideActorConstraint = new Lang.Class({
+    Name: 'insideActorConstraint',
+    Extends: Clutter.Constraint,
+
+    vfunc_update_allocation: function(actor, actorBox) {
```
cosimoc: Is there not a stock `ClutterConstraint` subclass that we could use instead of rolling our own?

danigm: I've done this with the `ClutterPathConstraint` instead of a custom one.

GeorgesStavracas:
Please merge it. No reason to hold it.

cosimoc:
Thanks!

PR #452, opened 2019-04-17
T25336

GeorgesStavracas:
This is a huge rebase, sorry. It would be pointless to try and describe every change, since every single patch had to be rewritten, from scratch sometimes. So I'll only write the most significant changes: what was dropped.

# Dropped Commits

## Upstreamed

 * d49bc431f bluetooth: Add missing 'BluetoothHardwareAirplaneMode' to D-Bus interface
 * cc85c681b st-adjustment: Mark all properties as EXPLICIT_NOTIFY
 * 777ec777f st-box-layout: Queue relayout on adjustment changes
 * 136689190 st: Clip StEntry to allocation
 * be69aa9dc st: Make StScrollables' paint volume reflect the unconstrained view
 * 00dc5450c dash: expand and center align DashItemContainer
 * e53996829 panel: Delegate container destruction to PanelMenu.ButtonBox
 * 74614fbd5 st-bin: Destroy child in ClutterActor:destroy vfunc
 * 9d1bdc9c8 shell: Remove ShellGenericContainer
 * 96dc4890d keyboard: Stop using Shell.GenericContainer
 * 7f4ee95e2 tests: Stop using Shell.GenericContainer
 * 21c32e26b inspector: Stop using Shell.GenericContainer
 * 35f2ff02a boxPointer: Stop using Shell.GenericContainer
 * 6cf63176f boxPointer: Rename show/hide to open/close
 * 649f153f7 loginDialog: Stop using Shell.GenericContainer
 * 91fe9f963 layoutManager: Subclass GObject.Object
 * 5a4dec6b8 layoutManager: Stop using Shell.GenericContainer
 * e9d7a73f8 workspaceSwitcherPopup: Stop using Shell.GenericContainer
 * 23ac18bc4 messageTray: Drop Shell.GenericContainer usage
 * 4c094f7c0 buttonBox: Drop Shell.GenericContainer usage
 * 0a8f15ac0 appMenuButton: Rename show/hide to fadeIn/fadeOut
 * b3ebf7e81 thumbnailBox: Stop using Shell.GenericContainer
 * 49d2162c7 iconGrid: Rename skip_paint to _skipPaint
 * 1eceddc38 iconGrid: Stop using Shell.GenericContainer
 * cec97e570 iconGrid: Rename ::key-focus-in signal
 * 7449c6568 dash: Set scale and opacity on parent actor
 * d3c2bf095 baseIcon: Stop using Shell.GenericContainer
[...]

PR #453, opened 2019-04-17
T25336 (Debian)

GeorgesStavracas:
Nothing was particularly dropped except the downstream backports. Counterpart to #452 

https://phabricator.endlessm.com/T25336


wjt:
This half looks fine to me.

PR #455, opened 2019-05-03
ShellDBus: Add SHELL_APP_ADDED_EVENT

danigm:
The AddApplication method uses this varaible as an identifier to record
this event. This variable should have been removed during the 3.32
rebase but the reference is there causing a crash on every call.

This patch recovers this line from previous commits.


cosimoc:
Merged, since this is obviously correct!

PR #456, opened 2019-05-03
appDisplay: remove reference to this._views

danigm:
this._views doesn't exists in the current version so we should remove
this reference because it's undefined.

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1355,7 +1355,7 @@ var AppDisplay = class AppDisplay {
 
     selectApp(id) {
         this._showView(Views.ALL);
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -1355,7 +1355,6 @@ var AppDisplay = class AppDisplay {
 
     selectApp(id) {
         this._showView(Views.ALL);
-        this._views[Views.ALL].view.selectApp(id);
```
cosimoc: This line should not be removed, it should be changed to `this._allView.selectApp(id);`

danigm: Updated

GeorgesStavracas:
Thanks

PR #459, opened 2019-05-07
clubhouse: Reload clubhouse paths on fail

danigm:
If we install/uninstall a different version of the clubhouse and a new
clubhouse notification comes to the shell, the notification code will
not find the image because the flatpak path has changed.

This patch fixes that problem reloading the clubhouse flaptak
directories if we can't find the animation image. This will be only done
once to avoid an infinite recursion in the case of trying to find a non
existing image.

https://phabricator.endlessm.com/T26543


cosimoc:
Looks great to me, thanks!

PR #461, opened 2019-05-10
Fix more access-after-destroy GJS warnings

cosimoc:
https://phabricator.endlessm.com/T26627


GeorgesStavracas:
> Looks good, we probably should upstream the second commit.

Nevermind, it's a downstream change :slightly_smiling_face:

PR #462, opened 2019-05-10
networkAgent: Add missing import

GeorgesStavracas:
This needs to be squashed with ff636dc058 at the next rebase.

https://phabricator.endlessm.com/T26608


jprvita:
An easy way to remember to squash on the next rebase is to create it as a fixup commit (pass `--fixup ff636dc` when committing). Then `git rebase --autosquash` will put it in the right place during the next rebase.

GeorgesStavracas:
Oh I'm pretty intimate to `--fixup`, it's just that autosquashing doesn't help a lot when rebasing GNOME Shell itself (because of the big number of patches)

PR #464, opened 2019-05-13
clubhouse: Do not add actions twice during the init

danigm:
The clubhouse notification parent class calls to _addActions before we
set the pagination and this is causing duplicates actions because then
we're adding the action again.

https://phabricator.endlessm.com/T26646

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -336,15 +336,7 @@ var ClubhouseAnimator = class ClubhouseAnimator {
 var ClubhouseNotificationBanner =
 class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
     constructor(notification) {
-        let originalBody = notification.bannerBodyText;
-        // We will show the text differently, so have the parent set no text for now
-        notification.bannerBodyText = '';
-
         super(notification);
-
-        this._textPages = [];
-        this._textIdx = -1;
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -458,6 +458,10 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
     }
 
     _addActions() {
+        // Don't add actions if pagination is not initialized yet
+        if (this._textIdx === undefined || this._textPages === undefined)
+            return;
```
cosimoc: The fact that we have to special case this both here and in `this._inLastPage()` made me think that the approach we're using to set the text is actually not the best here.
Instead of saving the text body beforehand, and re-injecting it later, I wonder if we couldn't instead override e.g. the `setBody()` method and assign `this._textPages`, etc, from there. My goal would be to remove checks for `this._textPages` and `this._textIdx` from these two methods.

cosimoc: Alternatively, if that's not possible, I'd split out a `this._isPaginationReady()` method or something like that.

danigm: Done, I've overridden the `setBody` method and add the text split there with a new flag attribute `_paginationReady` to do this only the first time the method is called in the parent constructor, because the `setBody` method is called again on page change and in that case we don't want to split the text again.

PR #465, opened 2019-05-20
fixup! sideComponent: Add SideComponent back

GeorgesStavracas:
https://phabricator.endlessm.com/T26597

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -63,7 +63,8 @@ const defaultParams = {
 };
 
 var LayoutManager = GObject.registerClass({
-    Signals: { 'hot-corners-changed': {},
+    Signals: { 'background-clicked': {},
+               'hot-corners-changed': {},
                'startup-complete': {},
                'startup-prepared': {},
                'monitors-changed': {},
--- a/js/ui/sideComponent.js
+++ b/js/ui/sideComponent.js
@@ -78,6 +78,15 @@ class SideComponent extends GObject.Object {
         this._propertiesChangedId =
             this.proxy.connect('g-properties-changed', this._onPropertiesChanged.bind(this));
 
+        // Clicking the background (which calls overview.showApps) hides the component,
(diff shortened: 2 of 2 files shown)


js/ui/layout.js
```diff
@@ -63,7 +63,8 @@ const defaultParams = {
 };
 
 var LayoutManager = GObject.registerClass({
-    Signals: { 'hot-corners-changed': {},
+    Signals: { 'background-clicked': {},
```
cosimoc: This is OK, but I'm not sure that's all we need.
As mentioned on the phabricator ticket, in the eos3.5 codebase `sideComponent.js` connects to this signal.
This patch will fix the warning when the signal is emitted, but functionality-wise, on master, even after this patch I don't see anything connecting to this signal.

GeorgesStavracas: You're right; brought back the code that connected to it.

js/ui/sideComponent.js
```diff
@@ -78,6 +78,15 @@ class SideComponent extends GObject.Object {
         this._propertiesChangedId =
             this.proxy.connect('g-properties-changed', this._onPropertiesChanged.bind(this));
 
+        // Clicking the background (which calls overview.showApps) hides the component,
+        // so trying to open it again will call WindowManager._mapWindow(),
+        // which will hide the overview and animate the window.
+        // Note that this is not the case when opening the window picker.
+        this._desktopShownId = Main.layoutManager.connect('background-clicked', () => {
```
cosimoc: Is there a reason this https://github.com/endlessm/gnome-shell/blob/eos3.5/js/ui/sideComponent.js#L60 and this https://github.com/endlessm/gnome-shell/blob/eos3.5/js/ui/sideComponent.js#L110 snippets were not brought back?

GeorgesStavracas: I did not bring them back because they are already there.

cosimoc: Lol sorry!

PR #466, opened 2019-05-22
T26593

GeorgesStavracas:
Simple fix that I managed to figure out while working on another issue.

https://phabricator.endlessm.com/T26593


andrunko:
The changes look fine but did you find out why the regression was introduced in the first place? Was it something missed during rebase?

GeorgesStavracas:
Yes, mistakenly dropped during the rebase too.

PR #467, opened 2019-05-22
T26532

GeorgesStavracas:
This was mistakenly dropped when rebasing. Bring it back.

https://phabricator.endlessm.com/T26532


andrunko:
lgtm - before merging, did you find out why this regression was introduced? was this something we missed during the rebase or ...?

GeorgesStavracas:
Aha! Figured out the actual root cause of it.

PR #468, opened 2019-05-23
icon-grid: improve wording when removing desktop icons

wjt:
Currently, if you right-click an app icon and click “Remove from
desktop”, you see a notification reading “$APP has been deleted”, which
is not true. If you drag an icon, the app center icon changes to a trash
can, and the label changes to “Delete”, which is inaccurate in the same
way.

Change these strings to say “remove[d]”, not “delete[d]”, per design
direction in March 2017(!).

https://phabricator.endlessm.com/T16267

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2723,7 +2723,7 @@ class AppCenterIcon extends AppIcon {
 
     handleViewDragBegin() {
         this.iconState = AppCenterIconState.EMPTY_TRASH;
-        this.replaceText(_("Delete"));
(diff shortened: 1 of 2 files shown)


andrunko:
> If you drag an icon, the app center icon changes to a trash can

I guess that is still true with the change? Is that expected? Otherwise looks good.

wjt:
It is still true. All the discussion on https://phabricator.endlessm.com/T16267 assumes that a trash can is appropriate even with “Remove” so I just took that as given.

PR #471, opened 2019-05-29
overview: Show apps again after relayout

danigm:
When the lid is closed and opened again in a desktop without any app
opened, the overview is hidden and never get shown again. This causes a
desktop with disabled buttons.

To avoid this, this patch calls to showApps just after the hide but only
if the app buttons are visibiles before the hide call.

https://phabricator.endlessm.com/T24778

### DIFF
--- a/js/ui/overview.js
+++ b/js/ui/overview.js
@@ -429,11 +429,6 @@ var Overview = class {
     }
 
     _relayout() {
-        // To avoid updating the position and size of the workspaces
-        // we just hide the overview. The positions will be updated
-        // when it is next shown.
-        this.hide();
-
         this._relayoutNoHide();
     }
 


js/ui/overview.js
```diff
@@ -432,9 +432,18 @@ var Overview = class {
...
 
         this._relayoutNoHide();
+
+        // The previous hide call will make *NotVisible* the apps icons
+        // in the desktop. If it was visible before this relayout we should
+        // show again to avoid a desktop with disabled icons and no apps opened
+        if (visible) {
+            this.visible = false;
+            this.showApps();
+        }
```
cosimoc: What causes the call to relayout() when the laptop lid is closed? Is there any way to avoid that?

I guess this also causes the apps page to be displayed if the laptop lid is closed while showing the window picker? Perhaps we should try to redisplay whichever page was visible, if we can't avoid the call to relayout at all.

js/ui/overview.js
```diff
@@ -432,9 +432,18 @@ var Overview = class {
...
+        const visible = this.visible;
         this.hide();
 
         this._relayoutNoHide();
+
+        // The previous hide call will make *NotVisible* the apps icons
+        // in the desktop. If it was visible before this relayout we should
+        // show again to avoid a desktop with disabled icons and no apps opened
+        if (visible) {
+            this.visible = false;
```
cosimoc: Why do we need to set this to `false` here?

GeorgesStavracas:
On a more general perspective, last time I dig into this issue, I reached a point where the ClutterClickAction that is added to the clone (and that is responsible for toggling the app icons) would sometimes simply stop working -- which might be a signal of broken input delivery -- or randomly be removed from the clone -- which might indicate it's being recreated.

Even if it turns out to not be related to it, I would be curious to know what's happening with the click action.

js/ui/overview.js
danigm: The relayout is because the lid is closed. This callback is triggered by the `monitors-changed` signal, emitted in the `LayoutManager` class and that signal is directly connected to the mutter `MonitorManager`.

js/ui/overview.js
danigm: This is set to false here because the `showApps` does nothing if `visible` is true.

danigm:
> On a more general perspective, last time I dig into this issue, I reached a point where the ClutterClickAction that is added to the clone (and that is responsible for toggling the app icons) would sometimes simply stop working -- which might be a signal of broken input delivery -- or randomly be removed from the clone -- which might indicate it's being recreated.
> 
> Even if it turns out to not be related to it, I would be curious to know what's happening with the click action.

About this, I've noticed that the screen lock is related to this issue, if you've the screen lock disabled, the overview is grayed out, but you can activate with a click anywhere.

cosimoc:
What we discussed with @danigm today is that there seem to be two related issues here:
- when the laptop lid is closed with no app window open, upon resume the desktop appears unfocused
- in that situation clicking on the desktop *also* does not bring it to focus, and has no effect

@GeorgesStavracas asked about looking into the second piece of this, which I agree would be great to understand, but fixing that problem is still not optimal, because ideally we would like the desktop not to lose focus entirely.

I've asked @danigm to look into whether we could remove the call to hide the overview upon relayout [1], or whether we can avoid receiving `monitors-changed` when the lid opens and closes, but if that does not find any other ways to accomplish this, I think the proposed patch is quite reasonable.

[1] https://github.com/endlessm/gnome-shell/blob/master/js/ui/overview.js#L432

danigm:
@cosimoc I've been testing removing the hide and it's working correctly for me with every test case that I've tried, closing and opening the laptop lid with and without apps and plugin a second monitor with and without apps.

This line is to avoid the redraw of workspaces, so I think that it's okay for us to remove this meantime we don't have workspaces.

PR #472, opened 2019-05-30
texture-cache: Remove also scaled keys from the cache

danigm:
We're storing in the texture cache images and scaled images appending
the scaling factor to the key. When a file changes the cache key
corresponding to that file is removed, but not the keys for the scaled
ones so that images in the cache are never reloaded.

This patch removes all keys from the cache related to the file that
changes, including those with the scaling factor.

https://phabricator.endlessm.com/T26505


GeorgesStavracas:
Let's go with it for 3.6, and continue moving this upstream.

PR #473, opened 2019-06-03
shellDBus: Add property to HackAppsManager to get whitelisted apps

cfoch:
https://phabricator.endlessm.com/T26924

### DIFF
--- a/data/dbus-interfaces/com.endlessm.HackableAppsManager.xml
+++ b/data/dbus-interfaces/com.endlessm.HackableAppsManager.xml
@@ -1,5 +1,6 @@
 <node>
   <interface name="com.endlessm.HackableAppsManager">
     <property name="CurrentlyHackableApps" type="ao" access="read"/>
+    <property name="WhitelistedAppIds" type="as" access="read"/>
   </interface>
 </node>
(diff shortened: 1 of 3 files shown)


cosimoc:
Before I look at the code, why does this need to be exposed? I don't think the sound server should do any other checks; it can assume that all the apps that end up in `HackableApps` have already gone past the whitelist. And conversely, if `FocusedApp` is not part of `HackableApps` it is not a hackable one.

cfoch:
The reason I was using this was because, if for example an app X gets focused, I would have to check if this is a hackable app or not. I could just look at the CurrentlyHackableApps property (which involves to create a proxy for each hackable app) but the case may be that the proxies for each object in CurrentlyHackableApps object paths are not ready yet. So we have to wait for these proxies to be ready. However, if the focused application is not a hackable app, we would wait forever.

cosimoc:
@cfoch Why would the proxies not be ready by the time the app X gets focused? You can create them synchronously in the sound server.
Either way yes, I think we should use the values from `CurrentlyHackableApps` for this purpose, so I don't think this PR is required.
Tangentially, in the long term we're looking to remove the whitelist altogether and use each app's desktop file to know whether it's hackable or not.

cosimoc:
I'm closing this since I'm pretty sure we don't want this change, but feel free to reopen if you disagree.

PR #474, opened 2019-06-04
X-Endless-Hackable

ptomato:
https://phabricator.endlessm.com/T26928


ptomato:
Unfortunately it seems prohibitively difficult to test this during the interval where my base system is EOS 3.5.x and gnome-shell master is EOS 3.6.x. I'd appreciate help from someone who has a 3.6.x system installed.

PR #476, opened 2019-06-04
windowManager: don't use ES6 `template strings`

wjt:
I spotted this while working on https://github.com/endlessm/gnome-shell/pull/475. It makes xgettext sad:

    $ ninja -C obj-x86_64-linux-gnu/ meson-gnome-shell-update-po
    ninja: Entering directory `obj-x86_64-linux-gnu/'
    [0/1] Running external command gnome-shell-update-po.
    xgettext: warning: file 'data/50-gnome-shell-system.xml' extension 'xml' is unknown; will try C
    js/ui/windowManager.js:1146: warning: unterminated string

(The first warning is an upstream bug.)

As it happens, this problem has no effect on the generated .pot file, but I think it would prevent xgettext from seeing translatable strings later in the file.

https://phabricator.endlessm.com/T25388

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1136,13 +1136,13 @@ var ShellWindowManagerAnimatableSurface = GObject.registerClass({
         if (effects.length === 0) {
             throw new GLib.Error(AnimationsDbus.error_quark(),
                                  AnimationsDbus.Error.UNSUPPORTED_EVENT_FOR_ANIMATION_SURFACE,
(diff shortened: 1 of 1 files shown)


ptomato:
Alternatively, you could update to gettext 0.20. https://savannah.gnu.org/bugs/index.php?50920

wjt:
Great news! Unfortunately it's not in Debian yet https://packages.debian.org/search?keywords=gettext but soon, soon.

PR #477, opened 2019-06-05
eos-payg: Add prefix and sufix characters for code insert

mazen-asef:
Add a prefix and sufix characters for codes inserted on the
payg-unlock-screen, making it more user-friendly.

This is blocked on endlessm/eos-payg#44 and endlessm/eos-payg-nonfree#18

https://phabricator.endlessm.com/T25987

### DIFF
--- a/js/misc/paygManager.js
+++ b/js/misc/paygManager.js
@@ -43,6 +43,8 @@ const EOS_PAYG_IFACE = '<node> \
 <property name="Enabled" type="b" access="read"/> \
 <property name="RateLimitEndTime" type="t" access="read"/> \
 <property name="CodeFormat" type="s" access="read"/> \
+<property name="CodeFormatPrefix" type="s" access="read"/> \
+<property name="CodeFormatSuffix" type="s" access="read"/> \
 </interface> \
 </node>';
 
@@ -206,6 +208,7 @@ var PaygManager = GObject.registerClass({
             this._codeFormatRegex = new GLib.Regex(this._codeFormat,
                                                    GLib.RegexCompileFlags.DOLLAR_ENDONLY,
                                                    GLib.RegexMatchFlags.PARTIAL);
+
         } catch (e) {
             logError(e, 'Error compiling CodeFormat regex: %s'.format(this._codeFormat));
             this._codeFormatRegex = null;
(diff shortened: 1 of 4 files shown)


mazen-asef:
@wjt I have opened https://github.com/endlessm/eos-payg/pull/44 and https://github.com/endlessm/eos-payg-nonfree/pull/18 to reflect the changes required in the daemon :)

js/ui/payg.js
```diff
@@ -191,7 +195,13 @@ var PaygUnlockUi = GObject.registerClass({
     }
 
     validateCurrentCode() {
-        return Main.paygManager.validateCode(this.entryCode);
+        if (Main.paygManager.validateCode(Main.paygManager.codeFormatPrefix
```
wjt: I think you should move the logic that adds the prefix (and suffix, depending on the second parameter) into `validateCode()` rather than changing every call site.

js/ui/paygUnlockDialog.js
```diff
@@ -373,6 +377,34 @@ var PaygUnlockDialog = GObject.registerClass({
         return buttonsBox;
     }
 
+    _createEntryArea() {
+        let head, tail = null;
```
wjt: I don't think this does what you expect it to do – it assigns `null` to `b` and leaves `a` as `uninitialized`. I think these variables should just be declared where they are used.

js/misc/paygManager.js
```diff
@@ -206,6 +210,13 @@ var PaygManager = GObject.registerClass({
             this._codeFormatRegex = new GLib.Regex(this._codeFormat,
                                                    GLib.RegexCompileFlags.DOLLAR_ENDONLY,
                                                    GLib.RegexMatchFlags.PARTIAL);
+
+            // Create a regular expression that matches common identifiers used on regex - if
```
wjt: This comment is now out of date.

js/misc/paygManager.js
```diff
@@ -318,6 +337,18 @@ var PaygManager = GObject.registerClass({
         return this._initialized;
     }
 
+    get codeFormatRegex() {
+        return this._codeFormatRegex;
```
wjt: I don't think this accessor is used.

js/misc/paygManager.js
```diff
@@ -43,6 +43,8 @@ const EOS_PAYG_IFACE = '<node> \
 <property name="Enabled" type="b" access="read"/> \
 <property name="RateLimitEndTime" type="t" access="read"/> \
 <property name="CodeFormat" type="s" access="read"/> \
+<property name="CodeFormatPrefix" type="s" access="read"/> \
+<property name="CodeFormatSufix" type="s" access="read"/> \
```
wjt: suffix

js/misc/paygManager.js
```diff
@@ -309,15 +325,30 @@ var PaygManager = GObject.registerClass({
...
-        let [is_match, match_info] = this._codeFormatRegex.match(code, 0);
         return is_match || (partial && match_info.is_partial_match());
     }
 
     get initialized() {
         return this._initialized;
     }
 
+    get codeFormatPrefix() {
+        return this._codeFormatPrefix;
```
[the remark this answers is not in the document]
mazen-asef: The idea was just to follow convention, because the paygmanager never returns info directly from the proxy, it always save on a local variable (e.g. ratelimitendtime and enabled). Of course, those two emits signals anytime they are changed.

If that's okay, I can just return them and get rid of the the setters too.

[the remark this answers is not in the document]
wjt: Fine by me either way. (I didn't write the original code.)

mazen-asef:
Decided to take @wjt comment in the name of simplicity - also, I removed a loop in the validateCode method where I pulled references from `Main.paygManager` instead of `this`.

PR #479, opened 2019-06-12
iconGridLayout: Protect against invalid JSON

GeorgesStavracas:
Passing an invalid JSON file may cause GNOME Shell
to have undefined behavior, due to the JSON loading
routines not being protected against broken files.

Fix that by wrapping the call in a try / catch.

https://phabricator.endlessm.com/T20843

### DIFF
--- a/js/ui/iconGridLayout.js
+++ b/js/ui/iconGridLayout.js
@@ -174,13 +174,19 @@ var IconGridLayout = GObject.registerClass({
     }
 
     _getDefaultIcons() {
-        let mergedJson = this._mergeJsonStrings(
-            this._loadConfigJsonString(DEFAULT_CONFIGS_DIR, DEFAULT_CONFIG_NAME_BASE),
-            this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, OVERRIDE_CONFIG_NAME_BASE),
-            this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, PREPEND_CONFIG_NAME_BASE),
-            this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, APPEND_CONFIG_NAME_BASE)
-        );
-        let iconTree = Json.gvariant_deserialize_data(mergedJson, -1, 'a{sas}');
+        let iconTree = null;
+
+        try {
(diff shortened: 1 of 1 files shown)


js/ui/iconGridLayout.js
```diff
@@ -174,13 +174,19 @@ var IconGridLayout = GObject.registerClass({
...
+        try {
+            let mergedJson = this._mergeJsonStrings(
+                this._loadConfigJsonString(DEFAULT_CONFIGS_DIR, DEFAULT_CONFIG_NAME_BASE),
+                this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, OVERRIDE_CONFIG_NAME_BASE),
+                this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, PREPEND_CONFIG_NAME_BASE),
+                this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, APPEND_CONFIG_NAME_BASE)
+            );
+            iconTree = Json.gvariant_deserialize_data(mergedJson, -1, 'a{sas}');
+        } catch (e) {
+            logError(e, 'Failed to read JSON config');
```
wjt: As written, this will cause the icon grid to be empty if there is a syntax error in any of the JSON files that get read, and the error that gets logged to the journal will not contain the filename of the invalid file.

I would personally have moved the call to `JSON.parse()` from `_mergeJsonStrings` into `_loadConfigJsonString` (which already has a suitable `try`/`catch`), and removed `String` from both function names.

On balance, I think it is probably OK like this – the empty icon grid will make it abundantly clear to anyone testing a custom image that something is wrong with the icon grid JSON, rather than silently falling back to the built-in grid which might not be spotted. Just curious if you considered the other approach.

GeorgesStavracas: > the empty icon grid will make it abundantly clear to anyone testing a custom image that something is wrong with the icon grid JSON

This is precisely the reason I went for full failure instead of skipping just what is broken.

> Just curious if you considered the other approach.

I did. In fact, moving the JSON calls to `_loadConfigJsonString` would be easier and reduce the code in size. But I thought it would not prevent us from shipping slightly broken configs.

PR #482, opened 2019-06-19
js: Make Shell responsive against app-fitler changes for current user

uajain:
Hook up to "app-filter-changed" signal from malcontent and if the app
filter is changed for the currently logged in user, enforce the new
app filter in the shell. This is one of the main pieces for our feature work
in T26893.

https://phabricator.endlessm.com/T26969

### DIFF
--- a/js/misc/parentalControlsManager.js
+++ b/js/misc/parentalControlsManager.js
@@ -60,8 +60,24 @@ var ParentalControlsManager = class {
                 logError(e, 'Failed to get parental controls settings');
             }
         }
+
+        this._manager.connect('app-filter-changed', (manager, uid) => {
+            let current_uid = Shell.util_get_uid();
+            // Emit 'changed' signal only if app-filter is changed for currently logged-in user.
+            if (current_uid == uid)
+                this._manager.get_app_filter_async(current_uid, Malcontent.GetAppFilterFlags.NONE,
+                                                   null, this._onAppFilterChanged.bind(this));
+        });
     }
 
+    _onAppFilterChanged(object, res) {
+        try {
+            this._appFilter = this._manager.get_app_filter_finish(res);
+            this.emit('changed');
+        } catch (e) {
+            logError(e, 'Failed to get new MctAppFilter for uid ' + Shell.util_get_uid() + ' on app-filter-changed');
+        }
+    }
     // Calculate whether the given app (a Gio.DesktopAppInfo) should be shown
     // on the desktop, in search results, etc. The app should be shown if:
     //  - The .desktop file doesn’t say it should be hidden.
@@ -88,3 +104,4 @@ var ParentalControlsManager = class {
         return this._appFilter.is_appinfo_allowed(appInfo);
     }
 };
+Signals.addSignalMethods(ParentalControlsManager.prototype);
--- a/js/ui/search.js
+++ b/js/ui/search.js
@@ -414,6 +414,10 @@ var SearchResults = class {
         Util.blockClickEventsOnActor(this.actor);
 
         this._parentalControlsManager = ParentalControlsManager.getDefault();
+        this._parentalControlsManager.connect('changed', () => {
+            this._reloadInternetProviders();
+            this._reloadRemoteProviders();
+        });
 
         let closeIcon = new St.Icon({ icon_name: 'window-close-symbolic' });
         let closeButton = new St.Button({ name: 'searchResultsCloseButton',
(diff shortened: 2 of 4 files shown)


js/ui/search.js
```diff
@@ -414,6 +414,13 @@ var SearchResults = class {
         Util.blockClickEventsOnActor(this.actor);
 
         this._parentalControlsManager = ParentalControlsManager.getDefault();
+        this._parentalControlsManager.connect('changed', () => {
+            this._reloadRemoteProviders();
+
+            // Reload internet search providers too
+            this._unregisterProvider(this._internetProvider);
+            this._registerProvider(this._internetProvider);
```
andrunko: I believe you also need to reload the internetProvider here. What it does today is that it loads the internetProvider on constructor using `InternetSearch.getInternetSearchProvider` and that always return the current default browser, but if the default browser is disabled but we have other browsers enabled we end up with no internet search provider.

This should be fine once we support disabling all browsers at once but I think this code could still ignore this assumption and check all available browsers (starting with the default one) and whether they are blacklisted, something like (pseudocode):
```

reloadInternetProvider() {
  if (this._internetProvider)
    this._unregisterProvider(this._internetProvider);
  internetProviders = InternetSearch.getInternetSearchProviders();
  for (let i = 0; i < internetProviders.length; i++) {
    let internetProvider = internetProviders[i];
    if (this._registerProvider(internetProvider))
      this._internetProvider = internetProvider;
  }
}
```

Also change _registerProvider to return false if the provider is blacklisted and true otherwise and invoke reloadInternetProvider() from the constructor.

You will also need to change or add a new function to get all available internet search providers (right now InternetSearch.getInternetSearchProvider() only returns the default browser). To get all browsers use Gio.get_all_for_type('x-scheme-handler/http') - see also js/misc/utils.js:getBrowserApp|Id methods.

andrunko: Note that the default browser should still be considered first here, so for example an impl of `InternetSearch.getInternetSearchProviders()` as mentioned above should return the default browser as first on the list.

andrunko: I am not sure if the order is important here (if it changes the way the results are displayed), but if so you also need to reload the internet search provider before reloading the remote providers (as done in the constructor).

js/misc/parentalControlsManager.js
```diff
@@ -60,8 +60,24 @@ var ParentalControlsManager = class {
                 logError(e, 'Failed to get parental controls settings');
             }
         }
+
+        this._manager.connect('app-filter-changed', (manager, uid) => {
+            let current_uid = Shell.util_get_uid();
+            // Emit 'changed' signal only if app-filter is changed for currently logged-in user.
+	    if (current_uid == uid)
+                this._manager.get_app_filter_async(current_uid, Malcontent.GetAppFilterFlags.NONE,
+                                                   null, this._onAppFilterChanged.bind(this));
```
andrunko: Any reason to make it async? We already do it synchronously above.

uajain: Ideally `_async` should be the default go-to approach. Here, we are just relaying the `app-filter-changed` signal (via another `changed` signal) to relevant entities that require an update if the app-filter is changed. Note, we only fire `changed` signal from `ParentalControlsManager` only when this _async call is finished. so, probably it seems fine to me. I didn't notice any problems using `_async` here.

 We do synchronously above because it's quite difficult to make it `_async` work properly. Note, this is the case only for construction time code-paths. Read https://phabricator.endlessm.com/T26969#729762 maybe?

andrunko: I see, yes, the code seems fine and should work (as you said we only fire `changed` when the async call finishes) but I was actually in doubt because of the comment you pointed out in the ticket heh - I misread it as "lets use sync for now everywhere", but I understand now you were referring to async calls on constructors only.

js/ui/search.js
uajain: I confirm with @wjt and this is not the approach we are taking here.

GNOME settings UI will expose all-or-nothing deal for browsers and hence disable everything for `x-scheme-handler/http` (including the default InternetProvider). This makes the shell's bit(this PR) easy to process.

wjt:
There is a "fitler" typo in the commit message subject.

js/ui/search.js
```diff
@@ -497,6 +501,19 @@ var SearchResults = class {
         this._reloadRemoteProviders();
     }
 
+    _reloadInternetProviders() {
+        if (this._internetProvider)
+            this._unregisterProvider(this._internetProvider);
+        this._internetProvider = null;
```
andrunko: nitpick: this is not needed as you assign it again in the next line.

uajain: Addressed.

PR #483, opened 2019-06-21
shellEntry: Update shell entry if browsers are disabled

uajain:
Update the shell's search-entry hint text if the default
internet provider is enabled or disabled via parental controls.

https://phabricator.endlessm.com/T26991

### DIFF
--- a/js/ui/shellEntry.js
+++ b/js/ui/shellEntry.js
@@ -4,10 +4,12 @@ const { Clutter, Gio, GObject, Gtk, Shell, St } = imports.gi;
 
 const Animation = imports.ui.animation;
 const BoxPointer = imports.ui.boxpointer;
+const InternetSearch = imports.ui.internetSearch;
 const Main = imports.ui.main;
 const Panel = imports.ui.panel;
 const Params = imports.misc.params;
 const PopupMenu = imports.ui.popupMenu;
+const ParentalControlsManager = imports.misc.parentalControlsManager;
 
 const Tweener = imports.ui.tweener;
 const Util = imports.misc.util;
@@ -145,16 +147,12 @@ var OverviewEntry = GObject.registerClass({
         this._spinnerAnimation = new Animation.AnimatedIcon(iconFile, SPINNER_ICON_SIZE);
         this._spinnerAnimation.actor.hide();
 
-        // Set the search entry's text based on the current search engine
-        let entryText;
-        let searchEngine = Util.getSearchEngineName();
+        this._parentalControlsManager = ParentalControlsManager.getDefault();
+        this._parentalControlsManager.connect('changed', () => {
+            this._updateSearchEntryText();
+        });
 
-        if (searchEngine != null)
-            entryText = _("Search %s and more…").format(searchEngine);
-        else
-            entryText = _("Search the internet and more…");
-
-        let hintActor = new St.Label({ text: entryText,
+        let hintActor = new St.Label({ text: '',
                                        style_class: 'search-entry-text-hint' });
 
         super._init({ name: 'searchEntry',
@@ -167,6 +165,8 @@ var OverviewEntry = GObject.registerClass({
                       secondary_icon: this._spinnerAnimation.actor,
                       x_align: Clutter.ActorAlign.CENTER,
                       y_align: Clutter.ActorAlign.CENTER });
+
+        this._updateSearchEntryText();
 
         this._blinkBrightnessEffect = new Clutter.BrightnessContrastEffect({
             enabled: false,
@@ -182,6 +182,27 @@ var OverviewEntry = GObject.registerClass({
         this.clutter_text.connect('key-press-event', this._onKeyPress.bind(this));
         this.clutter_text.connect('text-changed', this._onTextChanged.bind(this));
         global.stage.connect('notify::key-focus', this._onStageKeyFocusChanged.bind(this));
+    }
+
+    _updateSearchEntryText() {
+        let entryText;
+        let internetProvider = InternetSearch.getInternetSearchProvider();
+
+        if (internetProvider &&
+            internetProvider.appInfo &&
+            this._parentalControlsManager.shouldShowApp(internetProvider.appInfo)) {
+            // Set the search entry's text based on the current search engine
+            let searchEngine = Util.getSearchEngineName();
+
+            if (searchEngine != null)
+                entryText = _("Search %s and more…").format(searchEngine);
+            else
+                entryText = _("Search the internet and more…");
+        } else {
+            entryText = _("Search…");
+        };
+
+        this.hint_actor.set_text(entryText);
     }
 
     _isActivated() {


js/ui/shellEntry.js
```diff
@@ -184,6 +179,28 @@ var OverviewEntry = GObject.registerClass({
         global.stage.connect('notify::key-focus', this._onStageKeyFocusChanged.bind(this));
     }
 
+    _updateSearchEntryText() {
+
+        let entryText;
+
+        if (this._parentalControlsManager.isContentTypeAllowed("x-scheme-handler/http")) {
```
andrunko: I think it would probably be better to check whether we have an internet search provider here given that is what decides whether we can search the internet.

js/ui/shellEntry.js
```diff
@@ -184,6 +179,28 @@ var OverviewEntry = GObject.registerClass({
...
+            else
+                entryText = _("Search the internet and more…");
+        } else {
+            entryText = _("Search…");
+        };
+
+        let hintActor = new St.Label({ text: entryText,
+                                       style_class: 'search-entry-text-hint' });
+
+	this.hint_actor = hintActor;
```
andrunko: nitpick: missing space here.

js/ui/shellEntry.js
uajain: I think the shell can have the internet provider but it can be blocked via parental controls. I am also not sure what plumbing might be required here to actually decide if the entry text needs to be updated.

What is wrong with the current approach i.e. querying `x-scheme-handler/http` to decide?

js/ui/shellEntry.js
uajain: Addressed.

js/misc/parentalControlsManager.js
```diff
@@ -103,5 +103,19 @@ var ParentalControlsManager = class {
 
         return this._appFilter.is_appinfo_allowed(appInfo);
     }
+
+    isContentTypeAllowed (contentType) {
```
wjt:
```suggestion
    isContentTypeAllowed(contentType) {
```

js/misc/parentalControlsManager.js
```diff
@@ -103,5 +103,19 @@ var ParentalControlsManager = class {
...
+        // Are parental controls disabled at runtime?
+        if (this._disabled)
+            return true;
+
+        // Have we finished initialising yet?
+        if (!this._appFilter) {
+            log('Warning: Hiding app because parental controls not yet initialised: ' + appInfo.get_id());
```
wjt: `appInfo` is not defined. This is a copy-paste error from the block above.

js/ui/shellEntry.js
```diff
@@ -184,6 +179,28 @@ var OverviewEntry = GObject.registerClass({
         global.stage.connect('notify::key-focus', this._onStageKeyFocusChanged.bind(this));
     }
 
+    _updateSearchEntryText() {
+
```
wjt:
```suggestion
```

js/ui/shellEntry.js
```diff
@@ -184,6 +179,28 @@ var OverviewEntry = GObject.registerClass({
...
+            else
+                entryText = _("Search the internet and more…");
+        } else {
+            entryText = _("Search…");
+        };
+
+        let hintActor = new St.Label({ text: entryText,
+                                       style_class: 'search-entry-text-hint' });
+
+        this.hint_actor = hintActor;
```
wjt: Why replace the hint actor rather than calling `.set_text()` on the existing one?

js/misc/parentalControlsManager.js
uajain: Addressed! Thanks

js/ui/shellEntry.js
uajain: Addressed! Thanks

js/ui/shellEntry.js
uajain: Oh yea! that's right. Maybe I should spend some time reading documentation/API on `St` .
Re-iterated the PR to use this approach.

js/misc/parentalControlsManager.js
uajain: Addressed! Thanks

js/ui/shellEntry.js
andrunko: You are correct that we can have an internet search provider although it can be blocked, but I think the problem here is that we are making assumptions again on how browsers are blocked via parental controls (which is done by a different code/component).

It would be the same as only creating an Internet search provider if the content type is not blocked, which is not what we do, we create it and ask the MctAppFilter whether it is blocked via appinfo (which also checks executable path and flatpak app ID apart from content types to decide whether something is blacklisted).

To clarify, the code is not wrong, it will work with the current plans, but it is making assumptions that I dont think should be here.

andrunko: My suggestion is that we check if the internet search provider is blocked (not registered) and change the text accordingly.

andrunko: See also related discussions on #parental-controls@slack.

js/ui/shellEntry.js
```diff
@@ -184,6 +184,26 @@ var OverviewEntry = GObject.registerClass({
         global.stage.connect('notify::key-focus', this._onStageKeyFocusChanged.bind(this));
     }
 
+    _updateSearchEntryText() {
+        let entryText;
+        let internetProvider = InternetSearch.getInternetSearchProvider();
```
andrunko: Was checking this code and looks like the call to `InternetSearch.getInternetSearchProvider()` will always return a new object, wonder if that could be an issue or waste of resources (specially considering https://phabricator.endlessm.com/T27082). I wonder if we could access the Search.SearchResults here and check whether the cached internet provider (after reloading it when the parental controls settings change) is valid?

js/ui/shellEntry.js
```diff
@@ -184,6 +184,26 @@ var OverviewEntry = GObject.registerClass({
...
+
+        if (internetProvider &&
+            this._parentalControlsManager.shouldShowApp(internetProvider.appInfo)) {
```
andrunko: To be 100% safe here we should probably check if `internetProvider.appInfo` is valid as done by `Search._registerProvider` - although I think it is always valid for internet providers.

js/ui/shellEntry.js
andrunko: the `ViewsDisplay` class from (`viewSelector.js`) seems to have access to both the `SearchResults` and the `OverviewEntry` objects so it may be a good place to add the plumbing.

uajain: Isn't this getting a bit complicated for read / maintain the code? 
I mean,  I get your point, but this is something that might hit back during next rebase etc. 
It's exposing elements (e.g. shell entry hint text, cached providers) to other classes, because to read/write their values based on relevant `if .. else` cases.
Meanwhile I think, the `ViewsDisplay` class shouldn't be getting touched here at all.

andrunko: Yes, I agree this may be getting a bit complicated for the use case, but at the same time I am not sure about the extra resources used by this impl. Maybe we should check what others think? @pwithnall and/or @wjt what do you think? The changes look good, but see comments above.

wjt: I don't think there is a performance problem here. This function will only be called when the entry is constructed (once) and when parental controls change (rarely), and the returned object will only live until the next GC pass.

js/ui/shellEntry.js
wjt: Is there logic in the internet search provider which checks whether it should be active or not? (If not, what prevents internet search results being shown anyway?) If so, this seems to duplicate it.

andrunko: The logic is handled in `Search._registerProvider()` which gets called when parental controls change (the search providers are unregistered and registered again). The `_registerProvider()` code checks the `parentalControlsManager` object if the provider should be active. Maybe this logic could be moved to the search providers but probably done in a separate task.

js/ui/shellEntry.js
andrunko: Ok, fair enough, lets go with it then. This looks good apart from the small comments below.

js/ui/shellEntry.js
uajain: Addressed.

PR #486, opened 2019-06-28
eos-payg: Add the character enforcement for payg-tray

mazen-asef:
As we have set up on the screen unlock, set up the Pay As You Go
tray applet to enforce the correct characters entered in the field.

https://phabricator.endlessm.com/T25996

Aside from the task, I have moved the PaygUnlockCodeEntry class to `js/ui/payg.js` because all auxiliary funtions already reside there but the one mentioned, so it made more sense to move it there and generalize a bit better.


mazen-asef:
Oh, this is blocked on https://github.com/endlessm/gnome-shell/pull/485

PR #487, opened 2019-07-01
Add hack launcher to the desktop app grid

danigm:
This patch adds a new permanent launcher to the app grid to
enable/disable the HACK mode.

Right now the hack mode enables/disables the codeview and flip to hack
button and the clubhouse component.

https://phabricator.endlessm.com/T27140


js/ui/sessionMode.js
```diff
@@ -111,11 +111,11 @@ const _modes = {
                     ['networkAgent', 'polkitAgent', 'telepathyClient',
                      'keyring', 'autorunManager', 'automountManager',
                      'appStore', 'discoveryFeed', 'trayArea',
-                     'updaterManager', 'clubhouse'] :
+                     'updaterManager'] :
```
cfoch: why did you remove this? This makes the Clubhouse button to be hidden always.

danigm: I removed that because the clubhouse will be launched by the new desktop icon and not by the gnome-shell. But now I think that removing this breaks the clubhouse notifications and more, so I think I need to work on that a bit more to fix that overrides.

js/ui/appDisplay.js
```diff
@@ -2743,3 +2771,64 @@ class AppCenterIcon extends AppIcon {
...
+        this._iconContainer.add_child(this.icon.actor);
+
+        this.actor.set_child(this._iconContainer);
+
+        this.actor.connect('destroy', this._onDestroy.bind(this));
+    }
+
+    _createIcon(iconSize) {
+        let iconUri = 'resource:///org/gnome/shell/theme/hack-button.svg';
+        if (this._iconState === HackAppIconState.DEACTIVATED)
```
cfoch: `this.iconState` instead of `this._iconState`

danigm: It's the same, because the `iconState` is a getter for the `_iconState`, but yes, we should use the getter instead the attribute itself, I'll change that.

js/ui/sessionMode.js
danigm: I've left the clubhouse component and removed all the button code because we won't need that anymore, clubhouse button is not needed anymore.

js/ui/appDisplay.js
```diff
@@ -626,6 +633,10 @@ var AllView = class AllView extends BaseAppView {
     getLayoutIds() {
         let layoutIds = super.getLayoutIds();
 
+        // Only show the hack icon if the clubhouse app is in the system
+        if (_getClubhouse())
+            layoutIds.splice(0, 0, HACK_APP_ID);
```
erikos: This could cause side effects. If an element is not found an empty array is returned. No items on the desktop? :package: 
I think we need to guard this.

erikos: At least when installing/uninstalling the Clubhouse I have seen traces for this in the log and empty desktop grids.

danigm: I think that the problem was not there. I've been reviewing this and I've detected that the `hackAppIcon` variable was not set to null on redraw, causing a crash. I've fixed that and tested installing and uninstalling the clubhouse and now it seems to work correctly.

erikos: Great!

PR #488, opened 2019-07-02
T27162 first hack changes

joaquimrocha:
https://phabricator.endlessm.com/T27162


joaquimrocha:
I have noticed that I should have used a different key for the hack mode. I will update the PR.

joaquimrocha:
> I have noticed that I should have used a different key for the hack mode. I will update the PR.

@xjuan had a good idea about using a more Hack related key than the `enable-code-view`. So I guess this PR is fine to be reviewed (let's agree on the fine tuning during the review).

joaquimrocha:
Closing because we'll use the Clubhouse as a single window for now.

PR #489, opened 2019-07-03
codeView: Add highlight pulse animation for FtH

danigm:
https://phabricator.endlessm.com/T27141


js/ui/shellDBus.js
```diff
@@ -610,6 +610,12 @@ var HackableAppsManager = class {
             paths.push(session.hackableApp.objectPath);
         return paths;
     }
+
+    PulseFtH(appid, enable) {
+        let session = this._codeViewManager.sessions.find((s) => s.appId === appid);
```
cfoch: Maybe it would be better to add this method on the HackableApp class?

danigm: Yes, you're right, I'll move to there.

danigm: Done.

danigm:
I've rebased this on top of hack2.

data/dbus-interfaces/com.endlessm.HackableApp.xml
```diff
@@ -1,5 +1,8 @@
 <node>
   <interface name="com.endlessm.HackableApp">
+    <method name="PulseFtH">
```
manuq: I have other suggestions for the method name: "PulseFlipButton" or "PulseFlipToHackButton". Since "FtH" is a shortcut for the "lfip to hack" feature, but this method is about pulsating the button. In case you accept my suggestions, the clubhouse part should be adapted too.

danigm: I agree, it's better to be more explicit, I've done that change and I'll update the clubhouse part too.

PR #494, opened 2019-07-08
Upstream cherry-picks

GeorgesStavracas:
A light series of trivial cherry-picks, synchronizing our fork with `gnome-3-32` and adding https://github.com/endlessm/gnome-shell/commit/2715aadc9cda5168696ef3d023bca06d582dcc25 on top.

https://phabricator.endlessm.com/T27210

PR #497, opened 2019-07-10
Add parental controls filtering to favorite apps/taskbar pins

andrunko:
Some notes on the changes:
- Apart from the parental controls support for taskbar favorite apps, these changes includes fixes for a couple issues that are also valid for 3.6:
  - crash when pinning/unpinning app - see https://phabricator.endlessm.com/T27191#735008 for more details on how to reproduce
  - the taskbar app icon highlight state sometimes would not get updated when the app was closed (i.e. the button would stay highlighted)
- If an app is blacklisted while already running the app will stay running and the taskbar button for it will stay visible in the taskbar until the app is closed
- The favorites positions are respected when disabling/enabling the apps via parental controls

https://phabricator.endlessm.com/T27191
https://phabricator.endlessm.com/T26981

### DIFF
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -13,6 +13,7 @@ const BoxPointer = imports.ui.boxpointer;
 const Main = imports.ui.main;
 const Panel = imports.ui.panel;
 const PanelMenu = imports.ui.panelMenu;
+const ParentalControlsManager = imports.misc.parentalControlsManager;
 const PopupMenu = imports.ui.popupMenu;
 
 const MAX_OPACITY = 255;
@@ -599,6 +600,9 @@ const ScrolledIconList = GObject.registerClass({
         this._container.connect('style-changed', this._updateStyleConstants.bind(this));
 
         let appSys = Shell.AppSystem.get_default();
+
+        this._parentalControlsManager = ParentalControlsManager.getDefault();
+
         this._taskbarApps = new Map();
 
         // Update for any apps running before the system started
@@ -631,6 +635,15 @@ const ScrolledIconList = GObject.registerClass({
 
         appSys.connect('installed-changed', this._onInstalledChanged.bind(this));
         appSys.connect('app-state-changed', this._onAppStateChanged.bind(this));
+
+        this._parentalControlsManager.connect('changed', () => {
+            for (let [app, appButton] of this._taskbarApps) {
+                let shouldShow = this._parentalControlsManager.shouldShowApp(app.get_app_info());
+                let stopped = (app.state == Shell.AppState.STOPPED);
+
+                appButton.visible = !stopped || shouldShow;
+            }
+        });
     }
 
     setActiveApp(app) {
@@ -644,6 +657,13 @@ const ScrolledIconList = GObject.registerClass({
 
     getNumAppButtons() {
         return this._taskbarApps.size;
+    }
+
+    getNumVisibleAppButtons() {
+        let buttons = [...this._taskbarApps.values()];
+        return buttons.reduce((counter, appButton) => {
+            return appButton.visible ? counter : counter + 1;
+        }, 0);
     }
 
     activateNthApp(index) {
@@ -663,7 +683,7 @@ const ScrolledIconList = GObject.registerClass({
 
     _updatePage() {
         // Clip the values of the iconOffset
-        let lastIconOffset = this._taskbarApps.size - 1;
+        let lastIconOffset = this.getNumVisibleAppButtons() - 1;
         let movableIconsPerPage = this._appsPerPage - 1;
         let iconOffset = Math.max(0, this._iconOffset);
         iconOffset = Math.min(lastIconOffset - movableIconsPerPage, iconOffset);
@@ -707,7 +727,7 @@ const ScrolledIconList = GObject.registerClass({
     }
 
     isForwardAllowed() {
-        return this._iconOffset < this._taskbarApps.size - this._appsPerPage;
+        return this._iconOffset < this.getNumVisibleAppButtons() - this._appsPerPage;
     }
 
     calculateNaturalSize(forWidth) {
@@ -812,6 +832,11 @@ const ScrolledIconList = GObject.registerClass({
             this._container.replace_child(oldChild, newChild);
         else
             this._container.add_actor(newChild);
+
+        if (app.state == Shell.AppState.STOPPED &&
+            !this._parentalControlsManager.shouldShowApp(app.get_app_info())) {
+            newChild.hide();
+        }
     }
 
     _addButton(app) {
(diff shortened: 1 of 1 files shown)


andrunko:
See also https://phabricator.endlessm.com/T27239.

andrunko:
Rebased against latest master which already included https://github.com/endlessm/gnome-shell/commit/a61bfa742d21fa57ac11b09f07aed3f55923abf2 as part of https://github.com/endlessm/gnome-shell/pull/494.

js/ui/appIconBar.js
```diff
@@ -641,10 +641,22 @@ const ScrolledIconList = GObject.registerClass({
 
     setActiveApp(app) {
         this._taskbarApps.forEach((appButton, taskbarApp) => {
-            if (app == taskbarApp)
-                appButton.add_style_pseudo_class('highlighted');
-            else
-                appButton.remove_style_pseudo_class('highlighted');
+            var changed = false;
```
wjt: Does this change fix https://phabricator.endlessm.com/T26981?

andrunko: Yes, looks like it from a quick test. The initial problem I saw was that closing the app would not update the indicator but with this change it also seems to fix the other issues pointed in the task above.

andrunko: Yes, confirmed it fixes this issue - updated the task accordingly.

wjt: Please update the commit message and PR summary with the URL to that ticket in that case!

andrunko: Done.

js/ui/appIconBar.js
```diff
@@ -631,21 +641,52 @@ const ScrolledIconList = GObject.registerClass({
...
         });
     }
 
     getNumAppButtons() {
         return this._taskbarApps.size;
     }
 
+    getNumVisibleAppButtons() {
+        var count = 0;
+        this._taskbarApps.forEach((appButton, app) => {
```
GeorgesStavracas: This function can be simplified to:

```javascript
return this._taskbarApps.reduce((counter, appButton, index, array) => {
    return appButton.visible ? counter : counter + 1;
}, 0);
```

js/ui/appIconBar.js
```diff
@@ -631,21 +641,52 @@ const ScrolledIconList = GObject.registerClass({
...
         appSys.connect('app-state-changed', this._onAppStateChanged.bind(this));
+
+        this._parentalControlsManager.connect('changed', () => {
+            for (let [app, appButton] of this._taskbarApps) {
+                if (app.state == Shell.AppState.STOPPED &&
+                    !this._parentalControlsManager.shouldShowApp(app.get_app_info()))
+                    appButton.hide();
+                else
+                    appButton.show();
+            }
```
GeorgesStavracas: This is tidier if written as:

```javascript
let shouldShow = this._parentalControlsManager.shouldShowApp(app.get_app_info());
let stopped = app.state == Shell.AppState.Stopped;

appButton.visible = !stopped || shouldShow;
```

js/ui/appIconBar.js
```diff
@@ -631,21 +641,52 @@ const ScrolledIconList = GObject.registerClass({
 
         appSys.connect('installed-changed', this._onInstalledChanged.bind(this));
         appSys.connect('app-state-changed', this._onAppStateChanged.bind(this));
+
+        this._parentalControlsManager.connect('changed', () => {
+            for (let [app, appButton] of this._taskbarApps) {
+                if (app.state == Shell.AppState.STOPPED &&
+                    !this._parentalControlsManager.shouldShowApp(app.get_app_info()))
```
GeorgesStavracas: How does this interact with pinned buttons? Are they filtered as well?

andrunko: Yes,  actually pinned buttons are the only ones that should be filtered here as normal apps that have state `STOPPED` should not appear in `this._taskbarApps`. No need to check if the app is pinned cause if it happens (shouldn't) that an app is `STOPPED` and in `_taskbarApps` we want to hide it anyway.

js/ui/appIconBar.js
```diff
@@ -142,11 +143,17 @@ const AppIconMenu = class extends PopupMenu.PopupMenu {
...
-        parentActor.connect('notify::mapped', () => {
+        this._parentActorMappedId = parentActor.connect('notify::mapped', () => {
             if (!parentActor.mapped)
                 this.close();
         });
-        parentActor.connect('destroy', () => { this.actor.destroy(); });
+        parentActor.connect('destroy', () => {
+            if (this._parentActorMappedId > 0) {
+                parentActor.disconnect(this._parentActorMappedId);
+                delete this._parentActorMappedId;
```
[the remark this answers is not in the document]
andrunko: hmm, I had it initially like this (and without `delete`), but saw several other places in the shell where they call `delete sourceId`. I guess it has the same effect?

js/ui/appIconBar.js
```diff
@@ -142,11 +143,17 @@ const AppIconMenu = class extends PopupMenu.PopupMenu {
...
+            }
+            this.actor.destroy();
+        });
```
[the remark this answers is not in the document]
andrunko: This is already in a separate commit https://github.com/endlessm/gnome-shell/pull/497/commits/2b928d1f7586bcdffdd9ba20061d08dff981f57e. The reason this is here is to fix the crash I mentioned in https://phabricator.endlessm.com/T27191#735731, which is partially fix in master but still crashes without this change.

js/ui/appIconBar.js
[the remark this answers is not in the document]
andrunko: For example in [here](https://github.com/endlessm/gnome-shell/blob/master/js/ui/boxpointer.js#L479).

js/ui/appIconBar.js
```diff
@@ -631,21 +641,52 @@ const ScrolledIconList = GObject.registerClass({
...
+                    changed = true;
+                }
+            } else {
+                if (isHighlighted) {
+                    appButton.remove_style_pseudo_class('highlighted');
+                    changed = true;
+                }
+            }
+            if (changed)
+                appButton.queue_redraw();
```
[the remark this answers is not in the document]
andrunko: I will check this upstream change, but only calling add/remove style was not working here, hence why I forced `queue_redraw()`.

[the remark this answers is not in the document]
andrunko: Ok, I removed this change from this branch and will handle it as part of https://phabricator.endlessm.com/T26981 (where it belongs) - the changes to force `queue_redraw()` from JS fixes the issue in that task but the upstream fix you pointed does not seem to fix it, needs more investigation.

js/ui/appIconBar.js
andrunko: Done - had to do it a bit differently as `this._taskbarApps` is a map (not an array), but still used `Array.reduce`.

js/ui/appIconBar.js
andrunko: Done.

js/ui/appIconBar.js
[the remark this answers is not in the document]
andrunko: Updated anyway.

js/ui/appIconBar.js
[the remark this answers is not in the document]
andrunko: Left as is as explained above.

andrunko:
@GeorgesStavracas updated after comments, please check again.

js/ui/appIconBar.js
```diff
@@ -812,6 +853,10 @@ const ScrolledIconList = GObject.registerClass({
             this._container.replace_child(oldChild, newChild);
         else
             this._container.add_actor(newChild);
+
+        if (app.state == Shell.AppState.STOPPED &&
+            !this._parentalControlsManager.shouldShowApp(app.get_app_info()))
+            newChild.hide();
```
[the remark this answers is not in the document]
andrunko: Done.

PR #499, opened 2019-07-12
discoveryFeed: Add hightlight method to the button

danigm:
https://phabricator.endlessm.com/T27256


cfoch:
Nice! But make it an unlighted icon by default, please.

danigm:
> Nice! But make it an unlighted icon by default, please.

What do you mean by unlighted by default? It should not be highlighted until the clubhouse property changes, in the constructor the animation icon is created but not set as a child until the setHighlight method is called.

PR #502, opened 2019-07-16
AppIconMenu: Disconnect from parent actor signals when the actor is destroyed

andrunko:
~~No specific task but being handled at https://phabricator.endlessm.com/T27191.~~

~~See https://phabricator.endlessm.com/T27191#735731 for details on how to reproduce the crash.~~

https://phabricator.endlessm.com/T27279

Signed-off-by: Andre Moreira Magalhaes <andre@endlessm.com>


andrunko:
Force pushed after rebase against latest master.

PR #507, opened 2019-07-24
workspace: Protect against NULL apps

GeorgesStavracas:
Under some (unknown) circumstances, Shell.WindowTracker's
get_window_app() method may return NULL. The code under
workspace.js is not robust against that. This is reproducible
with Minetest.

Check for a NULL app, and return an empty title when that's
the case, before trying to retrieve the app title.

https://phabricator.endlessm.com/T27263

### DIFF
--- a/js/ui/workspace.js
+++ b/js/ui/workspace.js
@@ -594,6 +594,8 @@ var WindowOverlay = class {
 
         let tracker = Shell.WindowTracker.get_default();
         let app = tracker.get_window_app(metaWindow);
+        if (!app)
+            return '';
         return app.get_name();
     }
 


js/ui/workspace.js
```diff
@@ -594,6 +594,8 @@ var WindowOverlay = class {
 
         let tracker = Shell.WindowTracker.get_default();
         let app = tracker.get_window_app(metaWindow);
```
ptomato: I don't know enough about how gnome-shell keeps track of windows but it seems that the only case where this would return null, is if the window is not present in the `window_to_app` hash table. And it seems the only reason why that would happen, is if [`get_app_for_window()`](https://github.com/endlessm/gnome-shell/blob/master/src/shell-window-tracker.c#L431-L541) fails.

Most of this function falls back to `_shell_app_new_for_window()` except the test marked "Side components don't have an associated app". Could it be that minetest is creating a side component window, outside of gnome-shell's knowledge?

GeorgesStavracas: Without digging into that, I don't think that's the case. The only 2 side components I'm aware of are the Clubhouse, and the old app store. That's really not the role window that Minetest sets; reading the source code, Minetest doesn't seem to set any role.

ptomato: Then it seems like Minetest is creating a window that somehow avoids being tracked in `track_window()`?

GeorgesStavracas: That, or maybe a race condition. However, I mentioned in the Daily that I can't spend more time on this issue without compromising the more important and urgent DnD in upstream.

Your comment made me pretty confident, though, that `get_app_for_window()` **can** return `NULL` now, so this patch is correct as it is. It is not the root of the problem, but protecting against `NULL` here is still desired.

PR #508, opened 2019-07-25
T27347 keep background

cfoch:
https://phabricator.endlessm.com/T27347


danigm:
I think it could be better to add a gschema to the clubhouse to store this keys instead of storing it in the gnome-shell, the background is only needed in the clubhouse because there will be the only place where we'll change it, so maybe it's better to add this configuration keys to the clubhouse and then we don't need to modify the shell.

danigm:
This keys are moved to the clubhouse here: https://github.com/endlessm/clubhouse/pull/646

PR #511, opened 2019-07-25
appDisplay: Use AppIcon as HackAppIcon base class

danigm:
The HackAppIcon is only for launch the clubhouse, so we can use the
AppIcon base class and only override the icon itself.

This should fix the rare behaviour with the show/hide of the window
because it calls the app activation directly.

Unfortunately, this will bring back the spinning cursor problem with the
launch, but that problem is a minor issue and the show/hide problem is a
bigger problem right now. The solution was a workaround in the meantime the
cursor problem is solved, so we should wait until we solve the real
cursor problem.

https://phabricator.endlessm.com/T27357

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2787,17 +2787,16 @@ const HackAppIconState = {
 };
 
 var HackAppIcon = GObject.registerClass(
-class HackAppIcon extends ViewIcon {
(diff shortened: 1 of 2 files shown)


danigm:
This is not needed, the problem is in the clubhouse window, so I'm closing this PR.

PR #519, opened 2019-08-14
Hack2 deploy: rename hack dbus interfaces

erikos:
Rename hack dbus interfaces as well.

https://phabricator.endlessm.com/T27400


erikos:
The build is working now again and I tested it locally here with the rename of the corresponding part in the Clubhouse.

PR #520, opened 2019-08-28
Don't crash on startup if saved notification data is corrupt

wjt:
An Endless OS system was found in the wild with a malformed `.local/share/gnome-shell/notifications`. When deserialized in Python, after passing `trusted=True` to `g_variant_new_from_bytes()`, the first element of the first struct in the array looks like this:

    In [41]: _38.get_child_value(0).get_child_value(0)
    Out[41]: GLib.Variant('s',
    '\Uffffffff\Uffffffff\Uffffffff\Uffffffff\Uffffffff')

When deserialised in GJS, we get:

    gjs> v.get_child_value(0).get_child_value(0)
    [object variant of type "s"]
    gjs> v.get_child_value(0).get_child_value(0).get_string()
    typein:43:1 malformed UTF-8 character sequence at offset 0
     @typein:43:1
     @<stdin>:1:34

This branch addresses this problem on two fronts:

* Pass `trusted=FALSE` to `g_variant_new_from_bytes()`, as recommended by `g_variant_new_from_data()`'s documentation
* Handle exceptions from `_loadNotifications()`

I have tested this branch by copying `.local/share/gnome-shell/notifications` from the affected customer machine to my development VM. With just the changes to _loadNotifications(), I see the backtrace in the journal, but the shell comes up OK. With the additional change to pass `trusted=FALSE` to `g_variant_new_from_bytes()`, the journal is blissfully silent.

I don't have an explanation for how this file ended up being malformed. I also don't have an explanation for when this started crashing: my guess is that recent GJS became stricter about validating UTF-8 but I could be wrong!

https://phabricator.endlessm.com/T27153

### DIFF
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -730,7 +730,11 @@ var GtkNotificationDaemon = class GtkNotificationDaemon {
     constructor() {
         this._sources = {};
 
-        this._loadNotifications();
+        try {
+            this._loadNotifications();
+        } catch (e) {
+            logError(e, "Failed to load saved notifications");
+        }
 
         this._dbusImpl = Gio.DBusExportedObject.wrapJSObject(GtkNotificationsIface, this);
         this._dbusImpl.export(Gio.DBus.session, '/org/gtk/Notifications');
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -761,29 +761,31 @@ var GtkNotificationDaemon = class GtkNotificationDaemon {
     _loadNotifications() {
         this._isLoading = true;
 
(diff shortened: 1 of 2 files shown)


wjt:
Submitted upstream: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/694

wjt:
Verbal review from @pwithnall, who looked at the upstream version of this branch: “ship it”. (I am keen to get this into a 3.6.3 beta ASAP since users are hitting this in the wild.)

ptomato:
At least at some point in the past, SpiderMonkey did not validate UTF-8 when converting it to a string, as there are comments in the GJS source code to that effect. It does seem to do that now: https://searchfox.org/mozilla-central/source/js/src/vm/CharacterEncoding.cpp#291

PR #521, opened 2019-09-03
codeView: Use old or new HackToolbox depending on ID

danigm:
Old hack apps with the com.endlessm.X ID uses the old HackToolbox that
has the id com.endlessm.HackToolbox. The new hack apps and distribution
are using the new ID prefix com.hack_computer. but we can keep backward
compatibility in the shell to avoid break current hack-computer
installations.

This patch uses one toolbox or another depending on the app ID. This is
one step forward to the posibility of different toolboxes apps for
different apps, but right now there's only these two that are the same
but old and new versions.

https://phabricator.endlessm.com/T27673

### DIFF
--- a/js/ui/codeView.js
+++ b/js/ui/codeView.js
@@ -442,6 +442,7 @@ var CodingSession = GObject.registerClass({
 
         this._state = CodingSessionStateEnum.APP;
         this._toolboxActionGroup = null;
+        this._toolboxAppActionGroup = null;
 
         this._hackableProxy = null;
         this._hackablePropsChangedId = 0;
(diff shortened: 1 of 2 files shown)


js/ui/codeView.js
```diff
@@ -1484,7 +1498,7 @@ var CodeViewManager = GObject.registerClass({
         // HackUnlock and HackToolbox are inside the com.hack_computer.Clubhouse flatpak
         // and have the Clubhouse appInfo, so we should ignore those cases here
         // to be able to show the HackUnlock and HackToolbox windows
-        } else if (gtkId !== 'com.hack_computer.HackUnlock' && gtkId !== 'com.hack_computer.HackToolbox') {
+        } else if (gtkId !== 'com.hack_computer.HackUnlock' && !gtkId.endsWith('HackToolbox')) {
```
GeorgesStavracas: This seems rather fragile. What about checking for both the old and the new ids?

js/ui/codeView.js
```diff
@@ -514,6 +509,16 @@ var CodingSession = GObject.registerClass({
         return this._toolbox;
     }
 
+    get toolboxId() {
+        // FIXME: this should be extended to make it possible to launch
+        // arbitrary toolboxes in the future, depending on the application
+        if (this.appId.startsWith('com.endlessm.')) {
+            return 'com.endlessm.HackToolbox';
+        }
```
GeorgesStavracas: Style: single line `if`s don't need curly braces

Also, allow me to suggest the following:

```javascript
let prefix = 'com.hack_computer';

if (this.appId.startsWith('com.endlessm.'))
    prefix = 'com.endlessm';

return `${prefix}.HackToolbox`;
```

danigm: Okay, I've just follow your suggestion here and use the prefix code.

js/ui/codeView.js
danigm: Okay, I've updated this to check for both ids.

danigm:
Ok, let's merge this.

PR #522, opened 2019-09-11
windowManager: Transition splash to always to the screen center

cfoch:
https://phabricator.endlessm.com/T27474


danigm:
I'm closing this PR because I think we don't need it.

PR #523, opened 2019-09-11
Hack pulse bounce

danigm:
https://phabricator.endlessm.com/T27573


js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+            }
+            this.pulse();
+        });
+
         this.canDrop = false;
+
+        this.pulse();
+    }
+
+    pulse() {
```
GeorgesStavracas: Change this to a pair of `_startPulse()` and `_stopPulse()`

js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+        if (!this._pulse)
+            return;
+
+        const params = { time: 0.1, transition: 'easeOutQuad' };
+        this._tweenIcon({...params, scale_x: 1.1, scale_y: 1.1})
+            .then(this._tweenIcon.bind(this, {...params, scale_x: 0.9, scale_y: 0.9}))
+            .then(this._tweenIcon.bind(this, {...params, scale_x: 1.1, scale_y: 1.1}))
+            .then(this._tweenIcon.bind(this, {...params, scale_x: 0.9, scale_y: 0.9}))
+            .then(this._tweenIcon.bind(this, {...params, scale_x: 1, scale_y: 1}))
```
GeorgesStavracas: This is a bit different from the rest of the code base (though, it's an interesting approach!) I wonder if using promises won't conflict with the GLib mainloop.

Anyway, a minor nitpick: `1.0` instead of `1`.

js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+            .then(() => {
+                this._pulseWaitId = Mainloop.timeout_add(1000, () => {
```
GeorgesStavracas: Please use `GLib.timeout_add(GLib.PRIORITY_DEFAULT, ...` here

js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+                    this._pulseWaitId = 0;
+                    this.pulse();
+                    return GLib.SOURCE_REMOVE;
+                });
+            });
+    }
+
+    _tweenIcon(tweenParams) {
+        // Getting the icon without the label
+        const actor = this.icon._iconBin.first_child;
```
GeorgesStavracas: This is pretty much a no-go... can't you use `this.icon.icon` here?

js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+        actor.set_pivot_point(0.5, 0.5);
```
GeorgesStavracas: Please set this pivot point in the icon itself.

js/ui/appDisplay.js
```diff
@@ -2807,7 +2810,50 @@ class HackAppIcon extends AppIcon {
...
+        return new Promise((resolve) => {
+            if (!this._pulse) {
+                resolve(this);
+            }
```
GeorgesStavracas:
```javascript
if (!this._pulse)
    resolve(this);
```

js/ui/appDisplay.js
```diff
@@ -2837,6 +2888,10 @@ class HackAppIcon extends AppIcon {
     _onDestroy() {
         if (this._hackModeId)
             global.settings.disconnect(this._hackModeId);
+        if (this._hackPulseId)
+            global.settings.disconnect(this._hackPulseId);
+        if (this._pulseWaitId)
+            Mainloop.source_remove(this._pulseWaitId);
```
GeorgesStavracas: `GLib.source_remove`

js/ui/appDisplay.js
danigm: I think that there's no problem with Promises and mainloop, but maybe @ptomato will know better.

js/ui/appDisplay.js
danigm: Yes sure, I was looking into the `BaseIcon` constructor and I didn't found the `icon` param at first so I was using the path that I found using looking glass.

js/ui/appDisplay.js
```diff
@@ -2818,8 +2858,15 @@ class HackAppIcon extends AppIcon {
         let iconFile = Gio.File.new_for_uri(iconUri);
         let gicon = new Gio.FileIcon({ file: iconFile });
 
-        return new St.Icon({ gicon: gicon,
-                             icon_size: iconSize });
+        let icon = new St.Icon({ gicon: gicon,
+                                 icon_size: iconSize });
```
GeorgesStavracas: Nitpick:

```javascript
let icon = new St.Icon({
    gicon: gicon,
    icon_size: iconSize,
});
```

js/ui/appDisplay.js
```diff
@@ -2818,8 +2858,15 @@ class HackAppIcon extends AppIcon {
...
+        icon.set_pivot_point(0.5, 0.5);
```
GeorgesStavracas: I think you can pass the pivot point to the `St.Icon` constructor, like this:

```javascript
let icon = new St.Icon({
    gicon: gicon,
    icon_size: iconSize,
    pivot_point: new Clutter.Point({ x: 0.5, y: 0.5 }),
});
```

danigm: Yes, that works.

js/ui/appDisplay.js
ptomato: There should be no conflict, a promise's callback will be called in the next idle iteration after it resolves.

PR #525, opened 2019-09-13
Hack: Enable com.endlessm.Clubhouse for old users

danigm:
The migration to the new hack experience won't happen at once, there are
hack-computer users that will receive the EOS update and the old system
should continue working meantime the new hack experience is not
released.

This patch detects if the system is an old hack or a new one and uses
the old or new ids depending on that.

The old clubhouse shell button will be visible only if we detect
installed com.endlessm.Clubhouse and not com.hack_computer.Clubhouse. In
other case the new hack icon is shown, but only if the
com.hack_computer.Clubhouse app is installed.

If we detect an old hack environment we use:

 * com.endlessm.HackSoundServer
 * com.endlessm.Clubhouse
 * com.endlessm.HackComponents

https://phabricator.endlessm.com/T27778


js/ui/components/clubhouse.js
```diff
@@ -60,7 +57,7 @@ const ClubhouseIface =
   </interface> \
 </node>';
 
-function _clipToMonitor(actor) {
+function clipToMonitor(actor) {
```
cfoch: ~~Is there any advantage by renaming it? I would keep it as it was before.~~

js/ui/components/clubhouse.js
```diff
@@ -75,96 +72,10 @@ function _clipToMonitor(actor) {
     actor.set_clip(0, 0, clip, actor.height);
 }
 
-function getClubhouseApp() {
-    return Shell.AppSystem.get_default().lookup_app(CLUBHOUSE_ID + '.desktop');
+function getClubhouseApp(clubhouseId = CLUBHOUSE_ID) {
```
cfoch: I don't agree with having the clubhouse id as argument, because you can pass an id that is not the Clubhouse. Maybe you can add an argument, hackVersion argument, defaulting it to 2? Or an argument `isOldClubhouse` defaulting it to false.

js/ui/components/clubhouse.js
```diff
@@ -678,13 +590,17 @@ class ClubhouseNotificationSource extends NotificationDaemon.GtkNotificationDaem
 
 var Component = GObject.registerClass({
 }, class ClubhouseComponent extends SideComponent.SideComponent {
-    _init() {
-        super._init(ClubhouseIface, CLUBHOUSE_ID, CLUBHOUSE_DBUS_OBJ_PATH);
+    _init(clubhouseIface, clubhouseId, clubhousePath) {
```
cfoch: Would it be easier to have just an argument isOldClubhouse? If so, I would prefer that.

js/ui/components/clubhouse.js
```diff
@@ -708,14 +624,18 @@ var Component = GObject.registerClass({
     }
 
     get _useClubhouse() {
-        return this._imageUsesClubhouse() && !!getClubhouseApp();
+        return this._imageUsesClubhouse() && !!this.getClubhouseApp();
+    }
+
+    getClubhouseApp() {
```
cfoch: I think this method is not necessary. There is already a getClubhouseApp function!

js/ui/components/clubhouse.js
```diff
@@ -207,9 +118,10 @@ var ClubhouseAnimation = class ClubhouseAnimation extends Animation {
 };
 
 var ClubhouseAnimator = class ClubhouseAnimator {
-    constructor(proxy) {
+    constructor(proxy, clubhouseId) {
         this._proxy = proxy;
         this._animations = {};
+        this._clubhouseId = clubhouseId;
```
cfoch: I also think that you can use here just an argument isOldClubhouse. Or something similar.

js/ui/components/clubhouse.js
```diff
@@ -836,7 +756,7 @@ var Component = GObject.registerClass({
             let [appId, notificationId, notification] = params;
 
             // If the app sending the notification is the Clubhouse, then use our own source
-            if (appId == CLUBHOUSE_ID) {
+            if (appId === this._clubhouseId) {
```
cfoch: :+1:

js/ui/components/clubhouse.js
```diff
@@ -852,7 +772,7 @@ var Component = GObject.registerClass({
             let [appId, notificationId] = params;
 
             // If the app sending the notification is the Clubhouse, then use our own source
-            if (appId == CLUBHOUSE_ID) {
+            if (appId === this._clubhouseId) {
```
cfoch: :+1:

js/ui/components/clubhouse.js
```diff
@@ -678,13 +590,17 @@ class ClubhouseNotificationSource extends NotificationDaemon.GtkNotificationDaem
 
 var Component = GObject.registerClass({
 }, class ClubhouseComponent extends SideComponent.SideComponent {
-    _init() {
-        super._init(ClubhouseIface, CLUBHOUSE_ID, CLUBHOUSE_DBUS_OBJ_PATH);
+    _init(clubhouseIface, clubhouseId, clubhousePath) {
+        this._clubhouseId = clubhouseId || CLUBHOUSE_ID;
+        this._clubhouseIface = clubhouseIface || CLUBHOUSE_IFACE;
+        this._clubhousePath = clubhousePath || CLUBHOUSE_DBUS_OBJ_PATH;
+
```
cfoch: Even with that solely isOldClubhouse argument, I think you can keep this attributes/variables. But decide the value depending on the isOldClubhouse argument.

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+        // mode will be enabled by default for hack 1.0 users
+        const activated = global.settings.get_boolean('hack-mode-enabled');
+        if (!activated && this.getClubhouseApp())
+            global.settings.set_boolean('hack-mode-enabled', true);
+
+        super._init(CLUBHOUSE_IFACE, CLUBHOUSE_ID, CLUBHOUSE_DBUS_OBJ_PATH);
+        this._clubhouseButtonManager = null;
+
+        let keybindingSettings = new Gio.Settings({ schema: SHELL_KEYBINDINGS_SCHEMA });
+        Main.wm.addKeybinding('toggle-clubhouse',
```
cfoch: Could you add the keybinding as well as the `_updateCloseButtonPosition` function and `_visibleOpenButton` variable/attribute as part of a new commit?

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+        if (this._enabled)
+            super.toggle();
+    }
+
+    get _useClubhouse() {
+        return this._imageUsesClubhouse() && !!this.getClubhouseApp();
+    }
+
+    enable() {
+        super.enable();
```
cfoch: It should be only enabled if Clubhouse from Hack1 is installed.

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+
+        // Make sure the proxy didn't fail to initialize
+        if (this.proxy.SuggestingOpen !== undefined && this._clubhouseButtonManager)
+            this._clubhouseButtonManager.setSuggestOpen(!!this.proxy.SuggestingOpen);
+    }
+
+    _syncVisibility() {
+        super._syncVisibility();
+        if (this._clubhouseButtonManager)
+            this._clubhouseButtonManager.setVisible(this._enabled);
```
cfoch: In old code I see:
```
        if (this._questBanner)
            this._questBanner.actor.visible = this._enabled;

        if (this._itemBanner)
            this._itemBanner.actor.visible = this._enabled;
```
Is it not necessary now?

cfoch: I will reply myself. It is okay with the new changes, since that code is in the parent class.

js/ui/components/clubhouse.js
danigm: it's used in oldClubhouse.

js/ui/components/clubhouse.js
danigm: This method is here to make the code generic to be used in the oldClubhouse class that inherits this one, so the component code doesn't know if it's the new or the old one and this.getClubhouseApp will return the component app. This can be replaced with `getClubhouseApp(this._clubhouseId)`, but this is a shortcut.

js/ui/components/oldClubhouse.js
danigm: This is done in the parent class, checking with `this._useClubhouse`

danigm:
I've applied some of the suggestions, but not all.

I've created the oldClubhouse component to try to separate all the code to make it easier to remove in the future, that's the reason to try to avoid any reference to oldClubhouse or the old clubhouse id in the clubhouse.js component and try to move all code to the oldClubhouse.js file and, what's needed in the codeView.js.

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+
+        Main.overview.connect('showing', () => {
+            this.emit('window-changed');
+        });
+        Main.overview.connect('hidden', () => {
+            this.emit('window-changed');
+        });
+        Main.sessionMode.connect('updated', () => {
+            this.emit('window-changed');
+        });
```
GeorgesStavracas: This can be simplified to:

```javascript
Main.sessionMode.connect('...', () => this.emit('window-changed'));
```

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+
+        this._windowActor = actor;
+
+        this.emit('window-changed');
+
+        // Track Clubhouse's window actor to make the closeButton always show on top of
+        // the Clubhouse's window edge
+        this._notifyHandler = actor.connect('notify::x', () => {
+            this.emit('window-changed');
+        });
```
GeorgesStavracas: Here too

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+    let clubhouseWindowTracker;
+    return function() {
+        if (!clubhouseWindowTracker)
+            clubhouseWindowTracker = new ClubhouseWindowTracker();
+        return clubhouseWindowTracker;
+    };
+}());
+
+var ClubhouseOpenButton = GObject.registerClass({
+}, class ClubhouseOpenButton extends Soundable.Button {
```
GeorgesStavracas: Nitpick: 

```javascript
var ClubhouseOpenButton = GObject.registerClass(
class ClubhouseOpenButton extends Soundable.Button {
```

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+    destroy() {
+        if (this._updateVisibilityId)
+            getClubhouseWindowTracker().disconnect(this._updateVisibilityId);
+        this._closeButton.destroy();
+        this._openButton.destroy();
+    }
+});
+
+var Component = GObject.registerClass({
+}, class OldClubhouseComponent extends Clubhouse.Component {
```
GeorgesStavracas: Here too

js/ui/components/oldClubhouse.js
```diff
@@ -0,0 +1,339 @@
...
+      <arg type="u" direction="in" name="timestamp"/> \
+    </method> \
+    <method name="getAnimationMetadata"> \
+      <arg type="s" direction="in" name="path"/> \
+      <arg type="v" direction="out" name="metadata"/> \
+    </method> \
+    <property name="Visible" type="b" access="read"/> \
+    <property name="SuggestingOpen" type="b" access="read"/> \
+  </interface> \
+</node>';
```
GeorgesStavracas: You could move this to `data/dbus-interfaces/com.endlessm.Clubhouse.xml` and add:

```javascript
const { loadInterfaceXML } = imports.misc.fileUtils;

...

const ClubhouseIface = loadInterfaceXML('com.endlessm.Clubhouse');
```

js/ui/components/oldClubhouse.js
danigm: Done

js/ui/components/oldClubhouse.js
danigm: Done

js/ui/components/oldClubhouse.js
danigm: Done

js/ui/components/oldClubhouse.js
danigm: Done

js/ui/components/oldClubhouse.js
danigm: I've done this in the oldClubhouse and in the clubhouse.js too.

PR #526, opened 2019-09-17
background: Adjust to gnome-desktop API break

andrunko:
gnome-desktop broke API in commit ca5d61cf24, as it didn't *add* a property
as incorrectly stated in the commit message, but *replaced* an existing one.

https://gitlab.gnome.org/GNOME/gnome-shell/issues/1457

https://phabricator.endlessm.com/T27822


GeorgesStavracas:
retest this please

PR #530, opened 2019-09-19
Update changelog to 3.32.2

GeorgesStavracas:
This will be the version we'll be based on for a
few days.

https://phabricator.endlessm.com/T27794


wjt:
The CI is failing because the `Version_3.32.2[_debian]` tags are missing.

GeorgesStavracas:
retest this please

PR #533, opened 2019-10-01
T27749 hack new colors

manuq:
https://phabricator.endlessm.com/T27845


manuq:
I fixed the layout manager issue. This is now ready for final review.

manuq:
@danigm with this last commit I think this PR is mergeable! All except the line spacing was addressed (which should be addressed upstream in PangoText). Please review.

danigm:
I've just rebased this PR on top on the hack rebase branch

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -813,7 +817,12 @@ $clubhouse-image-height: 187px;
   }
   .notification-button {
     color: white;
-    background-color: #CE5611;
+
+    // This is the default for buttons after the 3rd one (unlikely).
+    background-gradient-start: $clubhouse_button_color_d;
```
ptomato: If the start and end of the gradient is the same, then it might be better to use `background-color` here?

js/ui/components/clubhouse.js
```diff
@@ -367,6 +367,9 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
 
         button.set_x_expand(false);
         super.addButton(button, callback);
+
+        // @todo: This is a workaround for the missing CSS selector, :nth-child
```
ptomato: nitpick: What's the upstream policy on whether `:nth-child` is likely to be implemented? If it is, then it would be convenient to add a link to the feature request in this comment.

js/ui/components/clubhouse.js
```diff
@@ -369,13 +375,17 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
     // Override this method because we don't want the button
     // horizontally expanded:
     addButton(button, callback) {
-        if (button.label === '>') {
+        if (Object.keys(CLUBHOUSE_ICONS).includes(button.label)) {
```
ptomato: I can see the previous code already worked by checking if the button label was a particular string. Now that we have three possibilities for icon buttons, is it now time to find a more robust way of specificing it? I would try to avoid (undocumented?) magic values. In particular this seems sensitive to typos and accidental extra whitespace.

js/ui/components/clubhouse.js
```diff
@@ -369,6 +369,10 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
         if (!this._inLastPage())
             return;
 
+        // If we paginated, remove the next button before adding others.
```
ptomato: I was wondering if I should comment on the `child-*` code to suggest updating it whenever a child is added or removed, but I thought it could always be added later if it was necessary. However, after reading this, it seems necessary :smile: In particular, because it seems likely we will continue to accumulate workarounds to make sure all the buttons are added in the right order, if we need to add any features to the buttons later.

js/ui/components/clubhouse.js
manuq: If I remember correclty @GeorgesStavracas said "very unlikely", but I think is a good idea to open a conversation upstream in any case.

js/ui/components/clubhouse.js
manuq: Yes, this is a workaround for the missing `nth-child` workaround :(

js/ui/components/clubhouse.js
manuq: Indeed this is not robust at all. Unfortunately `Gio.Notification` doesn't have a way to pass icons to buttons. I'm thinking if there is a more robust way to do it. If not, would documenting this work? The `>` is usually added by the Clubhouse code, but the thumbs-up, thumbs-down comes directly from the dialogues spreadsheet, sanitized on import but always prone to typos.

manuq:
I have addressed the comments in fixup commits:
- Added a feature request upstream for the :nth-child CSS pseudo class.
- Simplified CSS when gradients aren't needed.
- Added comment for icons replacing some emojis.

Also I added a new icon for the "previous arrow" emoji.

Still not ready for a second review pass, since I found a rebase regression in the buttons & character position:
 
dialogue-bug

js/ui/components/clubhouse.js
ptomato: We discussed on Slack that since these strings come from the dialogue spreadsheet, having the magic value is probably the best option, better than having separate columns for text and image for each button.

js/ui/components/clubhouse.js
ptomato: But can we give the workaround a bit more fidelity, by making it update the CSS classes when buttons are added/removed so that we don't have this requirement of never adding/removing buttons once the CSS classes are assigned?

manuq:
Thanks for the review @ptomato . I appended fixup commits to recalculate the CSS classes of buttons and to fix the regression depicted above.

danigm:
I've just fixed an small Issue with multiline notifications and the button and rebased against master merging the fixup commits with the corresponding one. I think this is ready to be merged. I've been testing and it works correctly.

manuq:
I have added a fixup commit since the latest rebase regressed the buttons and character position again:
dialogue-again


Now it's back to normal:
dialogue-again2

ptomato:
All good, please autosquash before merging.

manuq:
Thanks! Autosquashed and merged.

PR #534, opened 2019-10-02
clubhouse: Remove clip on notification after slideIn

danigm:
https://phabricator.endlessm.com/T27972


js/ui/components/clubhouse.js
```diff
@@ -505,6 +505,7 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
                                onUpdate: () => _clipToMonitor(this.actor),
                                onComplete: () => {
                                    // Ensure it only slides in once
+                                   /this.actor.remove_clip();
```
GeorgesStavracas: This won't run due to the extra `/`

danigm: comment/uncomment problem. Now it should be fixed.

PR #536, opened 2019-10-03
T27795 (Debian)

GeorgesStavracas:
https://phabricator.endlessm.com/T27795


debian/control
```diff
@@ -70,6 +70,7 @@ Vcs-Browser: https://salsa.debian.org/gnome-team/gnome-shell
 Package: gnome-shell
 Architecture: linux-any
 Depends: adwaita-icon-theme,
+         dconf-cli,
```
ptomato: Can this be pushed up to Debian, presumably that's where the postinst script comes from?

debian/control
```diff
@@ -14,6 +14,11 @@ Build-Depends: asciidoc-base,
                dh-sequence-gnome,
                eos-metrics-0-dev,
                gir1.2-accountsservice-1.0 <!nocheck>,
+               gir1.2-animations-dbus-0 <!nocheck>,
+               gir1.2-animation-glib-0 <!nocheck>,
+               gir1.2-eosmetrics-0 <!nocheck>,
+               gir1.2-eos-parental-controls-0 <!nocheck>,
```
ptomato: This should be `gir1.2-malcontent-0`, I think? (also below)

ptomato:
There's also a bash completion file that should be accounted for:
```
dh_missing: usr/share/bash-completion/completions/gnome-extensions exists in debian/tmp but is not installed to anywhere
	The following debhelper tools have reported what they installed (with files per package)
	 * dh_install: gnome-shell (7), gnome-shell-common (8)
	 * dh_installdocs: gnome-shell (2), gnome-shell-common (0)
	 * dh_installman: gnome-shell (0), gnome-shell-common (0)
	If the missing files are installed by another tool, please file a bug against it.
	When filing the report, if the tool is not part of debhelper itself, please reference the
	"Logging helpers and dh_missing" section from the "PROGRAMMING" guide for debhelper (10.6.3+).
	  (in the debhelper package: /usr/share/doc/debhelper/PROGRAMMING.gz)
	Be sure to test with dpkg-buildpackage -A/-B as the results may vary when only a subset is built
	For a short-term work-around: Add the files to debian/not-installed
dh_missing: missing files, aborting
```

debian/control
GeorgesStavracas: Yes

debian/control
ptomato: I guess you removed this instead?

GeorgesStavracas: No, the commit is still in the branch.

ptomato: I see, it was in a collapsed section.

PR #540, opened 2019-10-09
fixup! iconGridLayout: Add the Endless-specific class IconGridLayout

GeorgesStavracas:
When calling either the destroyCallback or the undoCallback,
Overview doesn't pass any arguments. However, the callbacks
passed by IconGridLayout were expecting an 'info' argument
that never came!

Remove these arguments and simplify the callbacks.

https://phabricator.endlessm.com/T28051


GeorgesStavracas:
I couldn't reproduce this bug, so I can't say for sure this is the proper fix. However, this is a blatant bug anyway, that appeared as the result of the previous rebase's switch from `Lang.bind()` to `() => {}`. Regardless of whether it actually fixes T28051 or not, this should be merged.

PR #541, opened 2019-10-09
clubhouse: Autostart migration quest for hack1 users

danigm:
We're removing the clubhouse button from right edge of the screen, to
guide users to the new clubhouse we need to launch the new clubhouse and
the migration quest will guide those users to the new experience.

https://phabricator.endlessm.com/T27985


danigm:
I've just rebased this PR on top on the hack rebase branch

js/ui/components/clubhouse.js
```diff
@@ -839,6 +839,25 @@ var Component = GObject.registerClass({
         this._hasForegroundQuest = false;
     }
 
+    _onProxyConstructed(object, res) {
+        super._onProxyConstructed(object, res);
+
+        // Check if the old HackComponents flatpak is installed, in that case,
+        // we've and old hack computer so we can launch the migration Quest
```
wjt:
```suggestion
        // this is an old hack computer so we can launch the migration Quest.
```

PR #542, opened 2019-10-10
Hack: Fix clubhouse import

danigm:
The use of CLUBHOUSE_ID outside the clubhouse.js file was causing some
problems due to the javascript import and importing cycles, when the
getClubhouseApp function is called in the codeView.js file, the first
time, the CLUBHOUSE_ID is not set yet, so it's undefined and because of
that it's unable to check if the clubhouse is installed.

To fix this problem, this patch changes the use of CLUBHOUSE_ID with a
literal string. Another solution could be to move this id to a new file
to remove the importing cycles, but as we're using it two times, it's
easier to use the literal here.

https://phabricator.endlessm.com/T27778

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -38,7 +38,7 @@ const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
 
 const CLUBHOUSE_BANNER_ANIMATION_TIME = 0.2;
 
-var CLUBHOUSE_ID = 'com.hack_computer.Clubhouse';
+const CLUBHOUSE_ID = 'com.hack_computer.Clubhouse';
 const CLUBHOUSE_DBUS_OBJ_PATH = '/com/hack_computer/Clubhouse';
 const ClubhouseIface = loadInterfaceXML('com.hack_computer.Clubhouse');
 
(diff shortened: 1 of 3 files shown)


js/ui/components/clubhouse.js
```diff
@@ -38,7 +38,7 @@ const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
 
 const CLUBHOUSE_BANNER_ANIMATION_TIME = 0.2;
 
-var CLUBHOUSE_ID = 'com.hack_computer.Clubhouse';
+const CLUBHOUSE_ID = 'com.hack_computer.Clubhouse';
```
ptomato: I'd just remove this if you're replacing all the occurrences with a literal.

erikos: Isn't that still used in https://github.com/endlessm/gnome-shell/blob/5f5ca0ca83116c0c88365fa59e6812228adae093/js/ui/components/clubhouse.js#L580 ?

erikos: I added a new commit cleaning that up. We can squash this before merging but I wanted it to be a new commit to be looked at.

ptomato:
Build failed, but that is due to some translation files, unrelated to this.

PR #558, opened 2019-10-23
Fix dragging icon past end of icon grid

ptomato:
https://phabricator.endlessm.com/T28156


ptomato:
First commit is also sent upstream: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/777

PR #559, opened 2019-10-23
Hack: Add final hack button assets

danigm:
https://phabricator.endlessm.com/T27847


erikos:
The Build fail does not seem to be relevant. Merging as it is only an asset swap.

PR #564, opened 2019-10-23
fixup! taskbar: Add the application icons taskbar and related style c…

GeorgesStavracas:
…hanges

 * Replace Tweener by Clutter implicit animations
 * Redraw on style changes

https://phabricator.endlessm.com/T28150

### DIFF
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -4,7 +4,6 @@ const { Clutter, Gdk, Gio, GObject,
         Gtk, Meta, Shell, St } = imports.gi;
 
 const Signals = imports.signals;
-const Tweener = imports.ui.tweener;
 
 const AppActivation = imports.ui.appActivation;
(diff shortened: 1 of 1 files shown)


js/ui/appIconBar.js
```diff
@@ -657,6 +662,8 @@ const ScrolledIconList = GObject.registerClass({
                 appButton.add_style_pseudo_class('highlighted');
             else
                 appButton.remove_style_pseudo_class('highlighted');
+
+            appButton.queue_redraw();
```
andrunko: As discussed on Slack with @GeorgesStavracas I was actually trying to fix this issue on `st_widget_recompute_style()` based on [this comment](https://github.com/endlessm/gnome-shell/pull/497#discussion_r303653166), but @GeorgesStavracas suggested we go with this as we won't have a solution for `st_widget_recompute_style()` for now.

This is fine to me and the changes work fine on local testing, so merging.

PR #565, opened 2019-10-23
fixup! taskbar: Add the application icons taskbar and related style c…

GeorgesStavracas:
…hanges

MetaShapedTexture became a GObject subclass that implements
ClutterContent in GNOME 3.34. The preview code assumed that
MetaShapedTexture is a ClutterActor, breaking window previews.

We can actually remove the ClutterClone usage from WindowMenuItem
and make the window preview a plain ClutterActor with a
ClutterContent attached to it -- win / win! This theoretically
does not account for client subsurfaces, but that's fine since
X11 doesn't support subsurfaces anyway.

For GNOME 3.36, we'll have MetaWindowContent which will account
not only for subsurfaces, but also effects applied to the window
actors.

https://phabricator.endlessm.com/T28331

### DIFF
--- a/js/ui/appIconBar.js
+++ b/js/ui/appIconBar.js
@@ -59,7 +59,10 @@ class WindowMenuItem extends PopupMenu.PopupBaseMenuItem {
         let maxH = (ratio > 1) ?
             PANEL_WINDOW_MENU_THUMBNAIL_SIZE / ratio : PANEL_WINDOW_MENU_THUMBNAIL_SIZE;
 
-        let clone = new Clutter.Clone({ source: windowActor.get_texture() });
+        let clone = new Clutter.Actor({
+            content: windowActor.get_texture(),
+            request_mode: Clutter.RequestMode.CONTENT_SIZE,
+        });
         let cloneW = clone.width;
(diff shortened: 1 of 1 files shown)


andrunko:
The changes look good but comparing to 3.7 the previews are not updated automatically unless I move the mouse over the preview. Can we do something about it?

I tested by opening 2 terminal windows, with one running `ls -laR /` and clicking the panel icon I can see the previews but they are not updated until I move the mouse over the preview.

GeorgesStavracas:
> Can we do something about it?

We probably can, but I'd suggest looking in this particular bug in another ticket. I took a brief look at it and tried a few hypothesis, and MetaShapedTexture is indeed invalidating its contents correct, but somehow this is not being propagated to the WindowMenuItem's particular actor.

The alternative would be use a ClutterClone of the window actor itself, but I would strongly oppose to it; ClutterClone must die, and even if it superficially fixes an issue, it's a can of worms for future regressions (see the broken icon grid dessaturated FBO madness.)

I suggest we go ahead with this PR, which at least **shows** a preview, and open another ticket for the lack of content updates in the window previews.

PR #571, opened 2019-10-28
iconGrid: Adjust internal alignment of icon components in icon container

ptomato:
Previously, icons were vertically centered within their bin layout.
However, the icon label may take up one or two lines, so that would mean
that icons were staggered vertically when icons with two-line labels
were next to icons with one-line labels.

This can be squashed with 36d607fc3fc1814a19cfab762a41481e74c9d439 on
the next rebase. It's not relevant to upstream since upstream icons only
have one line of text (which ellipsizes a lot less frequently since the
icons are larger.)

https://phabricator.endlessm.com/T28168

### DIFF
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -469,6 +469,7 @@ popup-separator-menu-item {
 }
 
 .overview-icon-label {
+    height: 46px;
     text-shadow: black 0px 2px 2px;
     border: none;
     background-color: transparent;


ptomato:
Before:
Screenshot from 2019-10-28 15-17-58

After:
Screenshot from 2019-10-28 15-05-36

Sadly, the hover highlight doesn't look good on the single-line icons now. Maybe we should increase the margins so there is more space left over on the top, at least. I'd recommend asking Robin to at least look at this and give an opinion on which is worse, before merging this PR.

andrunko:
> Sadly, the hover highlight doesn't look good on the single-line icons now. Maybe we should increase the margins so there is more space left over on the top, at least. I'd recommend asking Robin to at least look at this and give an opinion on which is worse, before merging this PR.

It would be interesting to know how that looked with the old Shell. Unfortunately I don't have a VM to check it atm.

ptomato:
Good point. Here it is:
Screenshot from 2019-10-28 16-36-17

It looks like the small margin at the top is actually not so objectionable since it was the old status quo :smile: On the other hand, here it looks like the label always requests two lines of space, and a one-line label is centered within those two lines of space. I'll go back to the code and implement that.

ptomato:
OK, I deleted that one line of code and wrote a different line of code. Now it looks much more like 3.6 did:
Screenshot from 2019-10-28 17-43-43

PR #572, opened 2019-10-29
Hack: Set the custom icon on drag to HackAppIcon

danigm:
This patch sets the custom icon that shows the hack state as drag actor,
by default it was using the Clubhouse icon that doesn't show the on/off
status.

https://phabricator.endlessm.com/T28375

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2748,6 +2748,10 @@ class HackAppIcon extends AppIcon {
         });
     }
 
+    getDragActor() {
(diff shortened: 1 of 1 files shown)


danigm:
> I trust that the change is fine. If you think this should be squashed into another commit on the next rebase, I suggest either committing with `--fixup=e187162cf` or noting in the commit message “Squash with '$patch_title' in next rebase”. Up to you, as the de-facto maintainer of the Hack shell customizations :-)

Okay, thanks, I'll do the fixup for this kind of changes.

PR #573, opened 2019-10-29
appDisplay: Avoid icon movement beyond limits

danigm:
Moving an icon behind the AppCenterIcon causes a shell error because

visibleItems[newPosition] is undefined

This patch fixes that problem and also avoid app movement before the
HackAppIcon.


ptomato:
I thought I fixed the shell error in https://github.com/endlessm/gnome-shell/pull/558, can you still reproduce it on master?

The other commit looks good to me.

danigm:
> I thought I fixed the shell error in #558, can you still reproduce it on master?
> 
> The other commit looks good to me.

I've just tested with the current master and I've the same shell error when trying to move an icon after the AppCenter icon.

js/ui/appDisplay.js
```diff
@@ -261,8 +261,18 @@ class BaseAppView {
 
     moveItem(item, newPosition) {
         let visibleItems = this._allItems.filter(item => item.actor.visible);
+
+        // Avoid overflow, this happens when trying to move an icon behind the
+        // AppCenterIcon
```
GeorgesStavracas: This comment can be reduced to simply `Avoid overflow`

js/ui/appDisplay.js
```diff
@@ -261,8 +261,18 @@ class BaseAppView {
...
+        if (newPosition >= visibleItems.length)
+            return -1;
+
         let targetId = visibleItems[newPosition].id;
 
+        // Do not allow to move icons before the HackAppIcon
+        if (targetId === Clubhouse.getClubhouseApp().get_id())
+            return -1;
```
GeorgesStavracas: Hm, I think this is the wrong place to put this check. One problem with the check being here is that there would be a visual feedback for dropping the icon (the Hack icon would "nudge"), but dropping would be ignored. Perhaps adding a similar check in `AppDisplay.handleDragOver()`.

js/ui/appDisplay.js
danigm: done

js/ui/appDisplay.js
danigm: Thank you, I've done that in the `handleDragOver` in the `HackAppIcon` class and it's better to have the code in this class instead of the parent class.

js/ui/appDisplay.js
```diff
@@ -2786,4 +2791,8 @@ class HackAppIcon extends AppIcon {
             GLib.source_remove(this._pulseWaitId);
         super._onDestroy();
     }
+
+    handleDragOver(_source, _actor, _x, _y) {
```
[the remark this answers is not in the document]
danigm: javascript!

PR #574, opened 2019-10-29
Fix YouTube MPRIS notifications

ptomato:
https://phabricator.endlessm.com/T28408


ptomato:
Note, accepted upstream already: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/791

PR #575, opened 2019-10-30
clubhouse: Improve styling of item notifications

manuq:
- Add CSS class to item notifications
- CSS: Add margin to the right to avoid colliding with the x button
- CSS: Set square size to item icon

https://phabricator.endlessm.com/T28392


danigm:
> One question: is this supposed to be squashed with another commit during the next rebase? If so, I'd suggest using `git commit --fixup=<commit hash>`, and adding a commit message, instead.

I think that we can squash with this commit:

d2fa8f6743df22cc91febb037689d9419dfcf1d0

manuq:
It will be more clear if this is a separate commit, IMO. Because it affects a different kind of notifications.

erikos:
I followed the final comment from Manuel to keep it as a separate commit.

PR #577, opened 2019-10-31
clubhouse: use priority field to mark a dialogue cancellable

danigm:
https://phabricator.endlessm.com/T28384


erikos:
Looks good indeed. I adjusted the commit message to make clearer what we are doing.

PR #578, opened 2019-10-31
T28152

GeorgesStavracas:
From the first commit:

```
The problem is complicated, so please bear with us: during the 3.34 cycle,
Clutter received an optimization where hidden actors are not allocated
anymore. This caused a few problems in the app grid, which uses
ClutterClones for running the spring animation, so a fix was committed
to ClutterClone that makes it allocate the preferred size of the
source actor when necessary.

Fast-foward to Endless OS 3.7; we introduce the ViewsClone class which
contains a bunch of real actors, and a ClutterClone for the IconGrid.
The ViewsClone is shown when the real icon grid is hidden, thus hitting
that code path that allocates the preferred size of the icon grid. Turns
out, most of the time the preferred size won't match the allocation made
by ViewsDisplayLayout; in fact, it will be much bigger.

Enter ClutterOffscreenEffect. The dessaturation effect applied by the
ViewsClone class has one unfortunate side-effect: the actor pixels are
cached in the GPU as a texture. ClutterOffscreenEffect itself has an
optimization where, if the actor didn't itself redraw, it would use the
cached framebuffer to draw instead (this is the optimization path taken
when using offscreen-redirect).

Now connecting the dots: we're drawing the GPU-cached, offscreen-redirected
icon grid using ClutterClone's preferred size -- which is much larger than
what it *should* be.

In essence, this commit restructures ViewsClone so that it doesn't use an
actual ClutterClone internally. Instead, use a real icon grid, and hide it
from picking entirely.
```

https://phabricator.endlessm.com/T28152

### DIFF
--- a/js/ui/viewSelector.js
+++ b/js/ui/viewSelector.js
@@ -594,25 +594,6 @@ class ViewsClone extends St.Widget {
                 );
             }
         });
-
-        let settings = Clutter.Settings.get_default();
-        settings.connect('notify::font-dpi', () => {
-            let overviewVisible = Main.layoutManager.overviewGroup.visible;
-            let saturationEnabled = this._saturation.enabled;
-
-            // Maybe because of the already known issue with FBO and ClutterClones,
-            // simply redrawing the overview group without assuring it is visible
-            // won't work. Clutter was supposed to do that, but it doesn't. The
-            // FBO, in this case, is introduced through the saturation effect.
-            this._saturation.enabled = false;
-            Main.layoutManager.overviewGroup.visible = true;
-
-            Main.layoutManager.overviewGroup.queue_redraw();
-
-            // Restore the previous states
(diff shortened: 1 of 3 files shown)


js/ui/viewSelector.js
```diff
@@ -598,25 +598,31 @@ class ViewsClone extends St.Widget {
...
+        if (!this._firstRelayout &&
+            this.visible &&
+            this.opacity != 0 &&
+            this._viewSelector.getActivePage() == ViewPage.APPS) {
 
-            // Restore the previous states
-            Main.layoutManager.overviewGroup.visible = overviewVisible;
-            this._saturation.enabled = saturationEnabled;
-        });
+            log('Queueing redraw');
```
ptomato: Remove this

GeorgesStavracas: Oops! Thanks

PR #580, opened 2019-11-04
fixup! Hack: Make hack icon removable

danigm:
https://phabricator.endlessm.com/T28387


danigm:
> If the purpose of this, and the previous merge requests, is to allow the Hack icon to move in the main app grid _but not to folders_, then it looks good.

The purpose to enable drag & drop for the Hack icon is only to make it removable from the grid dropping it in the app center, to keep it consistent with other icons, but it shouldn't be possible to move this icon in the desktop to other place, it should be fixed the same way the App Center icon is fixed.

PR #582, opened 2019-11-05
clubhouse: Dialogues should not be ellipsized

cfoch:
https://phabricator.endlessm.com/T28470


danigm:
I've been testing and this fix the problem with the text ellide, so I think it's safe to merge.

manuq:
It was concerning to me to remove the ellipsizing, but I grepped the code and saw the same in other notifications.

PR #589, opened 2019-11-07
power: Handle "100% but charging" case

ptomato:
I've observed that UPower can occasionally report a charge level of 100%
while the state is still "charging". This usually doesn't last very long
but it is noticeable because the power icon changes to a "missing icon"
icon. This will handle that rare case correctly.

https://phabricator.endlessm.com/T28321

### DIFF
--- a/js/ui/status/power.js
+++ b/js/ui/status/power.js
@@ -111,9 +111,12 @@ var Indicator = class extends PanelMenu.SystemIndicator {
         let chargingState = this._proxy.State == UPower.DeviceState.CHARGING
             ? '-charging' : '';
         let fillLevel = 10 * Math.floor(this._proxy.Percentage / 10);
-        let icon = this._proxy.State == UPower.DeviceState.FULLY_CHARGED
-            ? 'battery-level-100-charged-symbolic'
-            : `battery-level-${fillLevel}${chargingState}-symbolic`;
+        let icon;
+        if (this._proxy.State == UPower.DeviceState.FULLY_CHARGED ||
+            fillLevel === 100)
+            icon = 'battery-level-100-charged-symbolic';
+        else
+            icon = `battery-level-${fillLevel}${chargingState}-symbolic`;
 
         // Make sure we fall back to fallback-icon-name and not GThemedIcon's
         // default fallbacks


ptomato:
Upstream merge request: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/814

js/ui/status/power.js
```diff
@@ -111,9 +111,12 @@ var Indicator = class extends PanelMenu.SystemIndicator {
         let chargingState = this._proxy.State == UPower.DeviceState.CHARGING
             ? '-charging' : '';
         let fillLevel = 10 * Math.floor(this._proxy.Percentage / 10);
-        let icon = this._proxy.State == UPower.DeviceState.FULLY_CHARGED
-            ? 'battery-level-100-charged-symbolic'
-            : `battery-level-${fillLevel}${chargingState}-symbolic`;
+        let icon;
+        if (this._proxy.State == UPower.DeviceState.FULLY_CHARGED ||
+            fillLevel === 100)
```
andrunko: If `this._proxy.Percentage` for some reason returns > 100 we may end up with the same icon issue if `State != FULLY_CHANGED` and `fillLevel > 100`. Given we are already covering for issues with UPower, we may as well check if `fillLevel >= 100` here? The same also applies to the percentageLabel which may display a value > 100 in that case.

andrunko: if `this._proxy.Percentage` >= 110 I mean for the icon, because of the Math.floor :).

ptomato: Good point, although I'd say it's _really_ broken then, and maybe the icon _should_ be obviously broken in that case :stuck_out_tongue_closed_eyes:

andrunko: Good counterpoint :) - you are probably right and in that case it should make it easier to identify that the reported percentage is bogus.

Alternatively we could maybe log a warning if percentage > 100 but not break the UI (for both label and icon).

I will leave it for you to decide which approach you prefer. Approving this PR as is.

ptomato: In the meantime the upstream pull request was merged, and this is pretty minor, so I think let's leave it as is, to make the rebase simpler next time.

PR #592, opened 2019-11-12
T28577

GeorgesStavracas:
This PR cherry-picks an upstream fix (commit 7280409) and fixes T28577.

From the last commit:

```
ModalDialog is a subclass of St.Widget now, so using show() and hide() would
shadow the actual ClutterActor methods.
```

https://phabricator.endlessm.com/T28577

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1499,6 +1499,15 @@ class ViewIcon extends GObject.Object {
     }
 
     _onDestroy() {
+        if (this._itemDragBeginId) {
+            Main.overview.disconnect(this._itemDragBeginId);
+            delete this._itemDragBeginId;
+        }
+        if (this._itemDragEndId) {
+            Main.overview.disconnect(this._itemDragEndId);
+            delete this._itemDragEndId;
+        }
+
         this.actor._delegate = null;
     }
 
@@ -1597,9 +1606,6 @@ var FolderIcon = GObject.registerClass({
     }
 
     onDestroy() {
-        Main.overview.disconnect(this._itemDragBeginId);
-        Main.overview.disconnect(this._itemDragEndId);
-
         this.view.actor.destroy();
 
         if (this._spaceReadySignalId) {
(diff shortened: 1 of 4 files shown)


js/ui/appDisplay.js
```diff
@@ -2406,7 +2406,7 @@ var AppIcon = GObject.registerClass({
         let view = _getViewFromIcon(source);
 
         return source != this &&
-               (source instanceof AppIcon) &&
```
ptomato: Note that the class expression is named `AppDisplayIcon` and not `AppIcon` here. Class declarations and class expressions do have a special case for accessing the class name inside the class body. Sadly this is broken for GObject classes because of the `GObject.registerClass()` call wrapping the class expression.

You could work around it by moving the definition of `_canAccept` outside of the class body, as is done [here](https://gitlab.gnome.org/GNOME/gjs/blob/master/installed-tests/js/testGtk.js#L51-76):
```js
AppIcon.prototype._canAccept = function (source) {
    ...
    if (source instanceof AppIcon) ...
};
```
I bring this up because in theory the `this.constructor` solution will break if you have two different subclasses of AppIcon ("A" and "B") and you drag an A icon onto a B icon. However, I don't think AppIcon is actually subclassed in the current codebase, so you might just be able to ignore this.

GeorgesStavracas: Turns out we have a subclass of `AppIcon`: the `HackIcon`.

GeorgesStavracas: I simply removed this commit; things won't necessarily break without it.

ptomato: Oh, interesting, I grepped for `extends AppIcon` but maybe I didn't check thoroughly enough or maybe I checked on an upstream branch... Yes, that would break if dragging the Hack icon onto the app center icon :smile:

andrunko:
Pushed a couple commits on top of @GeorgesStavracas changes and it seems to work fine now and I could not detect any issue dragging and dropping icons back and forth.

@GeorgesStavracas feel free to update/change the commits.

andrunko:
Force pushed to fix a commit msg, no actual change.

PR #595, opened 2019-11-19
fixup! viewsSelector: New ViewsClone class for desaturated versions o…

GeorgesStavracas:
…f the desktop

The ViewsClone is not properly really applying the same set of style properties
compared to the App Grid, which ends up removing pixels from the top padding of
the entry, which misaligns everything.

Fix that by applying the appropriate style classes.

https://phabricator.endlessm.com/T28547


ptomato:
I'm just a bit too late with my review but I came here to post that this works for me as well :smile:

PR #597, opened 2019-11-20
fixup! clubhouse: Enable hack-mode for hack1 users

danigm:
https://phabricator.endlessm.com/T28649

This commit place the flag file inside the user config dir instead of the default path.

This fixup is a follow up of this PR:
https://github.com/endlessm/gnome-shell/pull/594#pullrequestreview-319803683

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -653,7 +653,8 @@ function _migrateHack1() {
 
     // Only enable the first time, to allow the user to disable the hack mode
     try {
-        Gio.File.new_for_path('.hack1-migrated').create(Gio.FileCreateFlags.NONE, null);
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -653,7 +653,8 @@ function _migrateHack1() {
 
     // Only enable the first time, to allow the user to disable the hack mode
     try {
-        Gio.File.new_for_path('.hack1-migrated').create(Gio.FileCreateFlags.NONE, null);
+        const filePath = `${GLib.get_user_config_dir()}/.hack1-migrated`;
```
erikos: let path = GLib.build_filenamev([GLib.get_user_config_dir(), '.hack1-migrated']);

Would be an option too, looking at other places in the shell.

danigm: Yes, I think that's a better option, let's use that instead.

PR #600, opened 2019-11-26
clubhouse: Button to move dialogue up / down

manuq:
https://phabricator.endlessm.com/T28800

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -39,6 +39,10 @@ const GtkNotificationDaemon = NotificationDaemon.GtkNotificationDaemon;
 
 const CLUBHOUSE_BANNER_ANIMATION_TIME = 200;
 
+// This margin is used to position the banner relative to the top-left
+// or bottom-left corners of the screen.
+const CLUBHOUSE_BANNER_MARGIN = 30;
+
 const CLUBHOUSE_DBUS_OBJ_PATH = '/com/hack_computer/Clubhouse';
 const ClubhouseIface = loadInterfaceXML('com.hack_computer.Clubhouse');
 
@@ -225,6 +229,24 @@ var ClubhouseAnimator = class ClubhouseAnimator {
     }
 };
 
+var QuestBannerPosition = class QuestBannerPosition {
+    constructor() {
+        this._atBottom = false;
+    }
+
+    get atBottom() {
+        return this._atBottom;
+    }
+
+    set atBottom(bool) {
+        this._atBottom = bool;
+    }
+
+    toggle() {
+        this.atBottom = !this.atBottom;
+    }
+};
+
 var ClubhouseNotificationBanner =
 class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
     constructor(notification) {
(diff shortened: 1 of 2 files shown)


js/ui/components/clubhouse.js
```diff
@@ -225,6 +225,24 @@ var ClubhouseAnimator = class ClubhouseAnimator {
     }
 };
 
+var QuestBannerPosition = class QuestBannerPosition {
+    constructor() {
+        this._atBottom = false;
```
cfoch: nitpick: I'd rename this to position, and let the value to be "like an enum" like in here https://github.com/endlessm/gnome-shell/blob/master/js/misc/soundServer.js#L36-L40

js/ui/components/clubhouse.js
```diff
@@ -460,7 +482,11 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
             this.actor.x = endX;
         }
 
-        this.actor.y = margin;
+        if (this._position.atBottom) {
+            this.actor.y = monitor.y + monitor.height - this.actor.height - margin - taskbarHeight;
+        } else {
+            this.actor.y = margin;
+        }
```
cfoch: maybe you can create a helper function to calculate the value depending on the position. Or something else to avoid code duplication. I see the same code below.

js/ui/components/clubhouse.js
manuq: Great idea, thanks!

js/ui/components/clubhouse.js
```diff
@@ -425,6 +445,8 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
             return;
 
         let margin = 30;
+        // FIXME obtain taskbar height?
+        let taskbarHeight = 40;
```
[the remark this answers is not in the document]
manuq: Good one, I will try it!

js/ui/components/clubhouse.js
manuq: Yes the code duplication should be fixed before the final review. There is more code duplication in the current code that needs refactor, see `_slideOut()`.

js/ui/components/clubhouse.js
manuq: After a second thought, I really prefer a boolean here. Because there are only two cases, at bottom or not (ie at top). An enum-like variable will make the check complicated, with a switch/case statement.

manuq:
@cfoch, @danigm or @GeorgesStavracas, can you take a look at this PR? Hopefully it gets merged prior to the shell freeze.

js/ui/components/clubhouse.js
```diff
@@ -577,6 +611,83 @@ class ClubhouseQuestBanner extends ClubhouseNotificationBanner {
...
+
+        if (this._position.atBottom)
+            this._upDownIcon.icon_name = 'go-top-symbolic';
+        else
+            this._upDownIcon.icon_name = 'go-bottom-symbolic';
+    }
+
+    _setupUpDownButton() {
+        this._upDownIcon = new St.Icon({ icon_name: 'go-bottom-symbolic',
+                                         icon_size: 16 });
```
GeorgesStavracas: Nitpick: 

```
this._upDownIcon = new St.Icon({
    icon_name: 'go-bottom-symbolic',
    icon_size: 16,
});
```

manuq: Good catch @GeorgesStavracas . I just updated the PR.

manuq:
I just found an outstanding bug introduced by mistake while cleaning up the code. It's in the fixup commit appended.

PR #603, opened 2019-11-28
Consistently label App Center icon as “More Apps”

wjt:
https://phabricator.endlessm.com/T28817

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2687,7 +2687,7 @@ class AppCenterIcon extends AppIcon {
     _onDragEnd() {
         super._onDragEnd();
 
-        this.icon.label.text = this.app.get_generic_name();
+        this.icon.label.text = this.name;
         this.icon.update();
     }
 
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2671,9 +2671,10 @@ class AppCenterIcon extends AppIcon {
         let app = appSys.lookup_app(EOS_APP_CENTER_ID);
 
         super._init(app, viewIconParams, iconParams);
+    }
 
-        this._id = EOS_APP_CENTER_ID;
-        this._name = this.app.get_generic_name();
+    get name() {
+        return _("More Apps");
     }
 
     _onDragBegin() {


js/ui/appDisplay.js
```diff
@@ -2671,9 +2671,10 @@ class AppCenterIcon extends AppIcon {
         let app = appSys.lookup_app(EOS_APP_CENTER_ID);
 
         super._init(app, viewIconParams, iconParams);
+    }
 
-        this._id = EOS_APP_CENTER_ID;
-        this._name = this.app.get_generic_name();
+    get name() {
+        return _("More Apps");
```
wjt: Doing this this way means that someone (hi) will need to go and copy over all the translations from gnome-software, so I'm prepared to be talked out of it, but I think the decision to show “More Apps” is more a property of our shell than it is of our app center package.

GeorgesStavracas: Alternatively, you can just move the call to `super._init()` to below these lines

wjt:
No, it doesn't work. `AppIcon._init()` is what `super._init()` calls, and it sets `this._name` to the app's non-generic name.

```
diff --git a/js/ui/appDisplay.js b/js/ui/appDisplay.js
index 0a7a5b0a1..b828e7464 100644
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1484,6 +1484,7 @@ class ViewIcon extends GObject.Object {
     }
 
     get name() {
+        log(`ViewIcon.get name: this._name = ${this._name}`);
         return this._name;
     }
 
@@ -2190,6 +2191,7 @@ var AppIcon = GObject.registerClass({
         this.app = app;
         this._id = app.get_id();
         this._name = app.get_name();
+        log(`AppIcon._init: this._name = ${this._name}`);
 
         let buttonParams = { button_mask: St.ButtonMask.ONE | St.ButtonMask.TWO };
         iconParams = Params.parse(iconParams, {
@@ -2670,10 +2672,11 @@ class AppCenterIcon extends AppIcon {
         let appSys = Shell.AppSystem.get_default();
         let app = appSys.lookup_app(EOS_APP_CENTER_ID);
 
-        super._init(app, viewIconParams, iconParams);
-
         this._id = EOS_APP_CENTER_ID;
-        this._name = this.app.get_generic_name();
+        this._name = app.get_generic_name();
+        log(`AppCenterIcon._init: this._name = ${this._name}`);
+
+        super._init(app, viewIconParams, iconParams);
     }
 
     _onDragBegin() {
```

Journal output:

```
Nov 29 16:52:56 camille gnome-shell[5395]: AppCenterIcon._init: this._name = More Apps
Nov 29 16:52:56 camille gnome-shell[5395]: AppIcon._init: this._name = App Center
Nov 29 16:52:56 camille gnome-shell[5395]: ViewIcon.get name: this._name = App Center
```

GeorgesStavracas:
Hm, right... I wonder if simply calling `this.icon.label.text = this.name;` would after the `this._name = this.app.get_generic_name();` line would work in that case.

wjt:
It does work. I have opened #605 with that one-line change. Personally I think it is clearer to override the `name` property, but I'm not too familiar with this code.

wjt:
@GeorgesStavracas preferred #605.

wjt:
@GeorgesStavracas preferred #605.

PR #606, opened 2019-12-04
T28153

GeorgesStavracas:
With the move to Clutter implicit animations, the duration of animations
is now measured in milisseconds rather than Tweener's seconds. That means
EXTRA_SPACE_ANIMATION_TIME is now 250 (ms) instead of 0.25 (s).

We're using EXTRA_SPACE_ANIMATION_TIME to call GLib.timeout_add_seconds(),
which means every time a folder popup is closed, we wait 250 seconds!

Fix that by using GLib.timeout_add() instead, which takes milisseconds as
input.

https://phabricator.endlessm.com/T28153


ptomato:
Good catch.

PR #610, opened 2020-01-03
Hack: Improve tooltip of Hack button

manuq:
Remove hardcoded newlines from the tooltip label, and instead A. set a
fixed width to the tooltip box and B. enable wrapping of text in the
tooltip description label.

https://phabricator.endlessm.com/T28841
https://phabricator.endlessm.com/T29075


wjt:
Landing this since it seems to be a CR pass.

PR #611, opened 2020-01-08
clubhouse: Scale notification icon for HiDPI

danigm:
https://phabricator.endlessm.com/T29017


wjt:
@GeorgesStavracas is this the correct way to scale assets in shell? It seems to me that this will cause the icon to be the wrong size if the dpi changes while it's displayed?

js/ui/components/clubhouse.js
```diff
@@ -299,6 +299,10 @@ class ClubhouseNotificationBanner extends MessageTray.NotificationBanner {
     }
 
     setIcon(actor) {
+        const scaleFactor = St.ThemeContext.get_for_stage(global.stage).scale_factor;
+        actor.width = actor.width * scaleFactor;
+        actor.height = actor.height * scaleFactor;
+
```
GeorgesStavracas: It doesn't look correct. Usually the scale factor is used to load higher resolution image assets, or translate coordinates correctly. Also, Mutter treats the scale as per monitor, not per stage, which makes this patch technically incorrect (though this is not a problem for us right now since on X11 there's only one "fake" device).

Also, this poses another problem: if this method it called with the same `actor` parameter multiple times, the sizes are scaled incorrectly.

I wonder why do you need to do that for Clubhouse notifications, but not for other Shell notifications.

Since I can't think of a better way to implement this right now, and I'm about to go PTO for the next couple of weeks, I suggest using `actor.set_scale(scaleFactor, scaleFactor);` for now.

danigm: Using `actor.set_scale(scaleFactor, scaleFactor)` doesn't solve the problem, it does the icon bigger but the positioning is not correct.

The layout for clubhouse notifications is different from the default notifications, I've detected that the icon is not correctly placed because the actor size.

The actor is an `Animation` and it looks like the image is scaled but not the actor size, so the image doesn't fit in the actor space and overflows. Maybe we should add this scale factor to the animation size, to fix this issue.

danigm:
I've updated the PR to modify the `Animation` class instead the clubhouse. If this is a good solution I'll move this code to upstream, but I've noticed that the master version in GNOME has changed and now inherits from `St.Bin` directly, so we'll need to solve this on master or update during the next release.

wjt:
@danigm was this resolved on master?

danigm:
> @danigm was this resolved on master?

Yes, this was merged upstream 3 months ago: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/937/diffs

PR #612, opened 2020-01-08
appDisplay: Temporarily disable "Remove from desktop" action for folders

andrunko:
Right now removing a folder from desktop works but there is no way to
undo the action as is done for non-folder icons.
Given that individual icons can still be removed from a folder and
that the folder gets deleted once the last icon is removed (and
recreated if we undo the removal of the last icon) lets remove this
option for folders for now to avoid confusing user trying to undo
the operation.

https://phabricator.endlessm.com/T29052

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1988,13 +1988,6 @@ var RenameFolderMenu = class RenameFolderMenu extends PopupMenu.PopupMenu {
         let separator = new PopupMenu.PopupSeparatorMenuItem();
         this.addMenuItem(separator);
 
-        // Add the "Remove from desktop" menu item at the end.
-        let item = new PopupMenu.PopupMenuItem(_("Remove from desktop"));
-        this.addMenuItem(item);
-        item.connect('activate', () => {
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -1989,11 +1989,13 @@ var RenameFolderMenu = class RenameFolderMenu extends PopupMenu.PopupMenu {
         this.addMenuItem(separator);
 
         // Add the "Remove from desktop" menu item at the end.
-        let item = new PopupMenu.PopupMenuItem(_("Remove from desktop"));
-        this.addMenuItem(item);
-        item.connect('activate', () => {
-            this._iconGridLayout.removeIcon(source.id, true);
-        });
+        // FIXME: re-enable menu below once the undo action for removing
```
wjt: Do we actually intend to do this? I suggest that we should just remove the code. If we're leaving it in place we should have a ticket for this pending work, and refer to this here.

andrunko: Fair enough, updated to remove the code.

PR #613, opened 2020-01-10
passwordReset: Make password reset codes last across a reboot

andrunko:
The new generated reset code will last until the user logs in again,
which happens automatically after the user changes the password once
it gets the unlock code and enters it in GDM or if the user remembers
the old password and logs in again.

https://phabricator.endlessm.com/T27921
https://phabricator.endlessm.com/T27290

### DIFF
--- a/js/gdm/authPrompt.js
+++ b/js/gdm/authPrompt.js
@@ -39,6 +39,17 @@ var BeginRequestType = {
     DONT_PROVIDE_USERNAME: 1
 };
 
+function _getMachineId() {
+    let machineId;
+    try {
+        machineId = Shell.get_file_contents_utf8_sync('/etc/machine-id');
+    } catch (e) {
+        logError(e, "Failed to get contents for file '/etc/machine-id'");
+        machineId = '00000000000000000000000000000000';
+    }
+    return machineId;
+}
+
 var AuthPrompt = class {
     constructor(gdmClient, mode) {
         this.verificationStatus = AuthPromptStatus.NOT_VERIFYING;
@@ -583,6 +594,12 @@ var AuthPrompt = class {
         this.emit('cancelled');
     }
 
+    _getUserLastLoginTime() {
+        let userManager = AccountsService.UserManager.get_default();
+        let user = userManager.get_user(this._username);
+        return user.get_login_time();
+    }
+
     _generateResetCode() {
         // Note: These are not secure random numbers. Doesn't matter. The
         // mechanism to convert a reset code to unlock code is well-known, so
@@ -592,8 +609,15 @@ var AuthPrompt = class {
         // version had one less digit in the code).
         let resetCode = Main.customerSupport.passwordResetSalt ? '1' : '';
 
-        for (let n = 0; n < _RESET_CODE_LENGTH; n++)
-            resetCode = '%s%d'.format(resetCode, GLib.random_int_range(0, 10));
+        let machineId = _getMachineId();
+        let lastLoginTime = this._getUserLastLoginTime();
+        let input = machineId + this._username + lastLoginTime;
+        let checksum = GLib.compute_checksum_for_data(GLib.ChecksumType.SHA256, input);
+        checksum = checksum.replace(/\D/g, '');
+
+        let hashCode = `${parseInt(checksum) % (10 ** _RESET_CODE_LENGTH)}`;
+        resetCode = `${resetCode}${hashCode.padStart(_RESET_CODE_LENGTH, '0')}`;
+
         return resetCode;
     }
 
@@ -629,14 +653,11 @@ var AuthPrompt = class {
         this._entry.clutter_text.set_password_char('');
         this._passwordResetCode = this._generateResetCode();
 
-        // FIXME: This string is too long. It is ellipsized, even in English.
-        // It must be shortened after the Endless 3.2 release, once there is
-        // time to translate the shortened version.
         // Translators: During a password reset, prompt for the "secret code" provided by customer support.
-        this.setQuestion(_("Enter unlock code provided by customer support:"));
+        this.setQuestion(_("Enter unlock code:"));
         this.setMessage(
             // Translators: Password reset. The first %s is a verification code and the second is an email.
-            _("Please inform customer support of your verification code %s by emailing %s. The code will remain valid until you click Cancel or turn off your computer.").format(
+            _("Please inform customer support of your verification code %s by emailing %s. Customer support will use the verification code to provide you with an unlock code, which you can enter here.").format(
                 this._passwordResetCode,
                 Main.customerSupport.customerSupportEmail));
 


js/gdm/authPrompt.js
```diff
@@ -583,6 +583,33 @@ var AuthPrompt = class {
...
+    _getFileContents(filename) {
+        let contents = null;
+        try {
+            let [ok_, bytes] = GLib.file_get_contents(filename);
+            if (bytes instanceof Uint8Array)
+                contents = imports.byteArray.toString(bytes);
+            else
+                contents = bytes.toString();
+        } catch (e) {
+        }
```
GeorgesStavracas: It would be interesting to at least `logError()` here, so we can retrieve eventual failures from the journal

js/gdm/authPrompt.js
```diff
@@ -583,6 +583,33 @@ var AuthPrompt = class {
...
+        return contents;
+    }
+
+    _getMachineId() {
+        let machineId = this._getFileContents('/etc/machine-id') ||
+                        this._getFileContents('/var/lib/dbus/machine-id');
+        if (machineId == null)
```
GeorgesStavracas: Nitpick: `===` (as per eslint)

js/gdm/authPrompt.js
```diff
@@ -583,6 +583,33 @@ var AuthPrompt = class {
         this.emit('cancelled');
     }
 
+    _getFileContents(filename) {
```
GeorgesStavracas: This method doesn't access `this` anywhere. It's better to move it to `js/mist/fileUtils.js` as a standalone function.

js/gdm/authPrompt.js
```diff
@@ -583,6 +583,33 @@ var AuthPrompt = class {
...
+            if (bytes instanceof Uint8Array)
+                contents = imports.byteArray.toString(bytes);
+            else
+                contents = bytes.toString();
+        } catch (e) {
+        }
+        return contents;
+    }
+
+    _getMachineId() {
```
GeorgesStavracas: Ditto about making it a standalone function (though it should be moved to the top of **this** file).

js/gdm/authPrompt.js
```diff
@@ -592,8 +619,14 @@ var AuthPrompt = class {
...
 
-        for (let n = 0; n < _RESET_CODE_LENGTH; n++)
-            resetCode = '%s%d'.format(resetCode, GLib.random_int_range(0, 10));
+        let machineId = this._getMachineId();
+        let lastLoginTime = this._getUserLastLoginTime();
+        let input = machineId + this._username + lastLoginTime;
+        let checksum = GLib.compute_checksum_for_data(GLib.ChecksumType.SHA256, input);
+        checksum = checksum.replace(/\D/g, '');
+
+        resetCode = '%s%d'.format(resetCode, parseInt(checksum) % (10 ** _RESET_CODE_LENGTH));
```
GeorgesStavracas: Nitpick: can you use template strings here?

```javascript
        resetCode = `${resetCode}${parseInt(checksum) % (10 ** _RESET_CODE_LENGTH)}`;
```

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done and actually fixed a small issue with padding the code to make sure it has the proper length.

andrunko:
Updated branch, please check again.

js/misc/fileUtils.js
```diff
@@ -97,3 +97,17 @@ function loadInterfaceXML(iface) {
...
 }
+
+function getFileContents(filename) {
+    let contents = null;
+    try {
+        let [ok_, bytes] = GLib.file_get_contents(filename);
+        if (bytes instanceof Uint8Array)
+            contents = imports.byteArray.toString(bytes);
+        else
+            contents = bytes.toString();
```
wjt: I don't understand why this instanceof check is necessary. Surely given we only support a single version of gjs at a time, the types returned by a particular function will always be the same.

In any case I think this entire function is unnecessary: use `Shell.get_file_contents_utf8_sync`.

Screenshot from 2020-01-14 12-18-09

js/gdm/authPrompt.js
```diff
@@ -39,6 +41,14 @@ var BeginRequestType = {
     DONT_PROVIDE_USERNAME: 1
 };
 
+function _getMachineId() {
+    let machineId = getFileContents('/etc/machine-id') ||
```
wjt: If you adjust this to use `Shell.get_file_contents_utf8_sync` you'll need to use `try { } catch { }` for robustness.

js/gdm/authPrompt.js
```diff
@@ -39,6 +41,14 @@ var BeginRequestType = {
...
+                    getFileContents('/var/lib/dbus/machine-id');
```
wjt: There is no need to fall back to this path. `/etc/machine-id` is guaranteed to be created very early in boot on systemd-managed systems. `/var/lib/dbus/machine-id` is a symlink to it so if the former can't be read, falling back to this path won't help.

js/gdm/authPrompt.js
```diff
@@ -636,7 +659,7 @@ var AuthPrompt = class {
         this.setQuestion(_("Enter unlock code provided by customer support:"));
```
wjt: While we're changing the other string here, please consider addressing https://phabricator.endlessm.com/T27290. I suggest we change this string to:

> Enter unlock code:

and clarify what this means in the longer string.

js/gdm/authPrompt.js
```diff
@@ -636,7 +659,7 @@ var AuthPrompt = class {
...
         this.setMessage(
             // Translators: Password reset. The first %s is a verification code and the second is an email.
-            _("Please inform customer support of your verification code %s by emailing %s. The code will remain valid until you click Cancel or turn off your computer.").format(
+            _("Please inform customer support of your verification code %s by emailing %s. The code will remain valid until your next login.").format(
```
wjt:
```suggestion
            _("Please inform customer support of your verification code %s by emailing %s. They will provide you with an unlock code.").format(
```

I don't think we need to mention how long the unlock code remains valid if it's (for all intents and purposes) "as long as necessary".

js/misc/fileUtils.js
andrunko: Nice catch, I failed to find an existing function in js/ for that. As for the instanceof check, this was needed otherwise gjs would complain about using toString directly or something like that - this pattern is used in several other places in the shell code.

Anyway, updated to use `Shell.get_file_contents_utf8_sync`.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
andrunko: Done.

js/gdm/authPrompt.js
```diff
@@ -633,10 +657,10 @@ var AuthPrompt = class {
         // It must be shortened after the Endless 3.2 release, once there is
         // time to translate the shortened version.
         // Translators: During a password reset, prompt for the "secret code" provided by customer support.
-        this.setQuestion(_("Enter unlock code provided by customer support:"));
+        this.setQuestion(_("Enter unlock code:"));
```
wjt: You can delete the FIXME above, having made this change.

js/gdm/authPrompt.js
```diff
@@ -592,8 +609,15 @@ var AuthPrompt = class {
...
 
-        for (let n = 0; n < _RESET_CODE_LENGTH; n++)
-            resetCode = '%s%d'.format(resetCode, GLib.random_int_range(0, 10));
+        let machineId = _getMachineId();
+        let lastLoginTime = this._getUserLastLoginTime();
+        let input = machineId + this._username + lastLoginTime;
+        let checksum = GLib.compute_checksum_for_data(GLib.ChecksumType.SHA256, input);
+        checksum = checksum.replace(/\D/g, '');
+
+        let hashCode = `${parseInt(checksum) % (10 ** _RESET_CODE_LENGTH)}`;
```
wjt: At some point in the past it was the case that backtick strings would break `xgettext` and prevent it from seeing any translatable strings later in the file. You can run `ninja meson-gnome-shell-pot` and check that the translatable strings below appear in the `po/gnome-shell.pot` file to verify whether this is still the case.

js/gdm/authPrompt.js
andrunko: Of course, sorry.

js/gdm/authPrompt.js
andrunko: Checked and it seems to work fine. Kept as is.

PR #614, opened 2020-01-13
payg: Add AccountID support for the UI

mazen-asef:
Create a notification on the screenshield that shows
the user which is the related ACCOUNT_ID tied to the computer.

https://phabricator.endlessm.com/T28225


mazen-asef:
Force-push to correct commit's phabricator task

mazen-asef:
I have added a standard padding that happens throughout gnome-shell (12px or 24px) 
```
      .unlock-dialog-payg-account-id {
          color: $osd_fg_color;
          font-size: 100%;
          text-align: left;
          margin: 12px;
      }
```

12px looked good to me.

PR #616, opened 2020-01-17
clubhouse: Fix item notification creation

danigm:
The item banner creation callback tries to use the newly created banner
and it's called directly in the constructor so the object doesn't exists
yet and cause an exception.

This patch makes the callback call async so the object will exists when
the callback is called.

There's also a reposition of the itemBanner when a new questBanner
appears and the same for the questBanner to recalculate all banners
positions when there are more than one, this fixes the item notification
position when there's quest notifications, placing that below.

https://phabricator.endlessm.com/T29166

### DIFF
--- a/js/ui/components/clubhouse.js
+++ b/js/ui/components/clubhouse.js
@@ -612,7 +612,7 @@ class ClubhouseQuestBanner extends ClubhouseNotificationBanner {
 
                 animation.play();
                 this.setIcon(animation.actor);
-                callback();
+                callback(this);
             });
         }
     }
@@ -703,7 +703,7 @@ class ClubhouseItemBanner extends ClubhouseNotificationBanner {
         super(notification);
         this.actor.add_style_class_name('clubhouse-item-notification');
         this._topBanner = null;
-        callback();
+        callback(this);
     }
 
     setTopBanner(topBanner) {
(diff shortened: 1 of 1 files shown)


js/ui/components/clubhouse.js
```diff
@@ -703,7 +703,13 @@ class ClubhouseItemBanner extends ClubhouseNotificationBanner {
         super(notification);
         this.actor.add_style_class_name('clubhouse-item-notification');
         this._topBanner = null;
-        callback();
+
+        // We can't call the callback directly because it depends on the object
+        // constructed here
+        GLib.idle_add(GLib.PRIORITY_DEFAULT, () => {
+            callback();
```
cfoch: Why do we need to pass to it a callback? I think we can remove that callback argument and move its body here. If I am correct the only place where we are using tha callback is [here](https://github.com/endlessm/gnome-shell/blob/master/js/ui/components/clubhouse.js#L1011-L1013). Or if you want to keep the callback, we can even pass `this` as the first argument of the callback.

danigm: Yes, we can remove the callback in this case and just call the code just below, but I think that this callback is there to have the same interface than the quest banner. Maybe it's better to add this as parameter, as you said, so we don't need the idle_add and will avoid future problems.

danigm: I've changed this to pass `this` as a param to the callback.

PR #617, opened 2020-01-20
js/messageTray: null-check notification to prevent double destroy

uajain:
https://phabricator.endlessm.com/T13041

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -552,6 +552,8 @@ class NotificationBanner extends Calendar.NotificationMessage {
 
         this._buttonBox.add(button);
         button.connect('clicked', () => {
+            callback();
+
             if (!this.notification.resident) {
                 // We don't hide a resident notification when the user invokes one of its actions,
                 // because it is common for such notifications to update themselves with new
@@ -560,8 +562,6 @@ class NotificationBanner extends Calendar.NotificationMessage {
                 this.emit('done-displaying');
                 this.notification.destroy();
             }
-
-            callback();
         });
 
         return button;


js/ui/messageTray.js
```diff
@@ -552,16 +552,16 @@ class NotificationBanner extends Calendar.NotificationMessage {
 
         this._buttonBox.add(button);
         button.connect('clicked', () => {
-            if (!this.notification.resident) {
+            callback();
+
+            if (this.notification && !this.notification.resident) {
```
wjt: This seems plausible but what I don't really understand is how `this.notification` can become `null`/`undefined` – I don't see any assignments to it.

uajain: I am building up on @andrunko 's investigation mentioned in the commit message: https://github.com/endlessm/gnome-shell/commit/9ba4a2b106a28d93a2bef635fe453e79ae9d253d

wjt: I wonder how this interacts with the 'done-displaying' signal emitted within this block – will this have been emitted by something higher up the call stack? Does it matter that it's not emitted?

uajain: Quick look suggests the notification should hide(actually terminate-displaying...) if the user  invokes its action by clicking of a button. That's why the `done-displaying` signal.

OTOH, if the notification has `setResident` set, the notification is *not* finished-displaying yet (because further information might be in-flight and notification is updated realtime), in that case, we don't want to emit `done-displaying` signal and destroy the notification.

andrunko: As @wjt pointed out we should make sure "done-displaying" is emitted even if the notification got destroyed in the callback (which is the case).

Before invoking the callback, save the value of "this.notification.resident" and after invoking the callback emit "done-displaying" regardless of whether the notification got destroyed in the callback?

Please also make sure the undo action still works with this change when removing icons from the desktop.

andrunko: @uajain missed your last reply but I still think we should emit "done-displaying" if the notification is not resident (even if it got destroyed in the callback), please verify and also check about the undo action...

uajain: Much deeper issue than I thought. But also reproduce-able on upstream version. 
https://gitlab.gnome.org/GNOME/gnome-shell/issues/2140

I think we should keep the revert (respecting [this](https://phabricator.endlessm.com/T13041#796575)) and fix the issue for "Undo" use-case upstream itself.

js/ui/messageTray.js
```diff
@@ -552,6 +552,8 @@ class NotificationBanner extends Calendar.NotificationMessage {
 
         this._buttonBox.add(button);
         button.connect('clicked', () => {
+            callback();
```
andrunko: Can't we just save the value of this.notification.resident here before invoking callback and use if after invoking the callback to decide whether to emit "done-displaying"?

Something like (pseudo code):
```
let resident = this.notification && this.notification.resident;
callback();
if (!resident) {
  this.emit('done-displaying');
  if (this.notification)
    this.notification.destroy();
}
```

andrunko: Ok, talked to @uajain about it and this won't work because NotificationBanner (`this` itself) gets deleted when callback is invoked, so lets merge it as is given no one is currently handling the `done-displaying` signal and follow-up with upstream on the issue.

Note that we will still get the backtrace in the journal but things should still work as expected.

PR #622, opened 2020-01-27
fixup! appDisplay: Continue loading icon grid even if an icon folder cant be loaded

andrunko:
Also remove folder from icon grid, otherwise it assumes the folder is visible and breaks state (e.g. when creating new folders by drag & drop).

The downside is that the user cannot (easily unless updating gsettings directly) restore the folder - for example by fixing wrong permissions - after restarting the shell, given the new layout will be saved without the folder.

https://phabricator.endlessm.com/T29205

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -651,6 +651,7 @@ var AllView = class AllView extends BaseAppView {
                     let item = Shell.DesktopDirInfo.new(itemId);
                     if (!item) {
                         log(`Error loading folder for ${itemId}`);
+                        this._iconGridLayout.removeIcon(itemId, false);
                         return;
                     }
                     icon = new FolderIcon(item, this);


js/ui/appDisplay.js
```diff
@@ -651,6 +651,7 @@ var AllView = class AllView extends BaseAppView {
                     let item = Shell.DesktopDirInfo.new(itemId);
                     if (!item) {
                         log(`Error loading folder for ${itemId}`);
+                        this._iconGridLayout.removeIcon(itemId, false);
```
wjt: Is it safe to modify the structure while it is being iterated?

andrunko: It is not safe, but in this case we are iterating over `items` (a copy) and removing from the icon grid.

wjt: OK, but will something unexpected happen, like (for example) re-entrantly entering the code which reads desktop files?

andrunko: It should not, also tested locally and didn't spot any issue.

PR #623, opened 2020-02-06
appDisplay: add new badge to hack icon.

xjuan:
Monitor /var/lib/flatpak/app/com.hack_computer.Clubhouse/current/active to know when
a new clubhouse version is instaled and notify the use by adding a "new" badge over
the hack icon.

https://phabricator.endlessm.com/T29254


js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
...
+
+        iconContainer.add_child(icon);
+        iconContainer.add_child(this._badge);
+        this._badge.hide();
+
+        this._checkNewRelease();
+
+        let monitor = this._flatpak_link.monitor(0, null);
+        if (monitor)
+            monitor.connect('changed', this._checkNewRelease);
```
wjt: Have you tested that this works? I'm pretty sure that `this` in the callback will not be bound to what you expect it to be bound to, and you in fact want one of the following:

```
            monitor.connect('changed', () => this._checkNewRelease());
```

or

```suggestion
            monitor.connect('changed', this._checkNewRelease.bind(this));
```

wjt: I also expect that the `monitor` object will be garbage-collected at some point since no reference to it is kept outside the scope of this function, so the callback will probably not fire.

js/ui/appDisplay.js
```diff
@@ -2747,6 +2747,7 @@ class HackAppIcon extends AppIcon {
 
         let app = Clubhouse.getClubhouseApp();
         this._activated = false;
+        this._flatpak_link = Gio.File.new_for_path('/var/lib/flatpak/app/com.hack_computer.Clubhouse/current/active');
```
wjt: You may be able to get this path out of the `app` object rather than hardcoding it?

js/ui/appDisplay.js
```diff
@@ -2747,6 +2747,10 @@ class HackAppIcon extends AppIcon {
...
+        this._monitor = this._flatpak_link.monitor(0, null);
+        if (this._monitor)
+            this._monitor.connect('changed', this._checkNewRelease.bind(this));
```
wjt: Does this need to be cleaned up somewhere? It looks to me like this creates a reference cycle.

wjt:
I see from https://phabricator.endlessm.com/T29254 that there's some design pushback on this change. Do we want to defer/close-without-merging this PR?

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
         });
     }
 
+    _checkNewRelease() {
+        this._flatpak_link.query_info_async(Gio.FILE_ATTRIBUTE_TIME_CHANGED,
+                                            0, 0, null, (file, res) => {
```
GeorgesStavracas:
```suggestion
        this._flatpak_link.query_info_async(Gio.FILE_ATTRIBUTE_TIME_CHANGED,
            0, 0, null, (file, res) => {
```

Style nitpick

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
...
+        let icon = new St.Icon({
+            layout_manager: new Clutter.BinLayout(),
+            gicon: new Gio.FileIcon({ file: iconFile }),
             icon_size: iconSize,
-            pivot_point: new Clutter.Point({ x: 0.5, y: 0.5 }),
+            pivot_point: center,
+        });
+
+        let badgeUri = 'resource:///org/gnome/shell/theme/hack-button-badge-new.svg';
+        let badgeFile = Gio.File.new_for_uri(badgeUri);
```
GeorgesStavracas: `const`

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
...
+            y_expand: true,
         });
+
+        iconContainer.add_child(icon);
+        iconContainer.add_child(this._badge);
+        this._badge.hide();
+
+        this._checkNewRelease();
+
+        let monitor = this._flatpak_link.monitor(0, null);
```
GeorgesStavracas: `const`

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
         });
     }
 
+    _checkNewRelease() {
+        this._flatpak_link.query_info_async(Gio.FILE_ATTRIBUTE_TIME_CHANGED,
+                                            0, 0, null, (file, res) => {
+            let info = file.query_info_finish(res);
+            let flatpak = info.get_attribute_uint64(Gio.FILE_ATTRIBUTE_TIME_CHANGED);
+            let last_launch = global.settings.get_int64('hack-last-launch');
```
GeorgesStavracas: Let's start using `const` instead of `let` for variables that are only assigned once.

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
...
         if (this._activated)
             iconUri = 'resource:///org/gnome/shell/theme/hack-button-on.svg';
 
         let iconFile = Gio.File.new_for_uri(iconUri);
-        let gicon = new Gio.FileIcon({ file: iconFile });
+        let center = new Clutter.Point({ x: 0.5, y: 0.5 });
 
-        return new St.Icon({
-            gicon: gicon,
+        let icon = new St.Icon({
```
GeorgesStavracas: `const`

js/ui/appDisplay.js
```diff
@@ -2822,19 +2823,60 @@ class HackAppIcon extends AppIcon {
...
+
+        this._badge = new St.Icon({
+            gicon: new Gio.FileIcon({ file: badgeFile }),
+            icon_size: iconSize / 2,
+            pivot_point: center,
+            translation_x: iconSize * 0.5,
+            translation_y: iconSize * -0.5,
+        });
+
+        let iconContainer = new St.Widget({
```
GeorgesStavracas: `const`

PR #625, opened 2020-02-11
debian: Add a endless custom session mode

uajain:
https://phabricator.endlessm.com/T28777


uajain:
We decided against going down this path : https://phabricator.endlessm.com/T28777#801566

PR #627, opened 2020-02-21
T29401

GeorgesStavracas:
This brings two Ubuntu patches that forbid session mode extensions to be overriden or updated.

The second commit was slightly modified to check for 'endless' instead of 'ubuntu'

https://phabricator.endlessm.com/T29401


uajain:
LGTM

PR #632, opened 2020-03-17
payg: Properly init accountID menu item

mazen-asef:
Trying to update the accountID every time we synced the remaining
time made the weird behaviour of not properly populating it the
first time, discarding the rest of the runs. Now, we only add the
item on  the menu just once, after paygManager has been fully
initialized.

https://phabricator.endlessm.com/T28492

### DIFF
--- a/js/ui/status/payg.js
+++ b/js/ui/status/payg.js
@@ -32,9 +32,6 @@ var Indicator = class extends PanelMenu.SystemIndicator {
         });
         this.menu.addMenuItem(this._item);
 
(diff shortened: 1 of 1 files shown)


mwleeds:
@GeorgesStavracas is currently rebasing the shell, so if we land this now we'll have to make sure it's included in the rebase branch.

mazen-asef:
Talked with @GeorgesStavracas, he'll add this in the list of commits to rebase

PR #633, opened 2020-03-19
payg/automaticUpdates: Fix banner notification

mazen-asef:
After https://phabricator.endlessm.com/T29401, the notification
banners emitted by the automaticUpdates and the payg classes didn't
work because they depended on an unexpected sessionMode. This fixes
that to the old behaviour.

https://phabricator.endlessm.com/T29582

### DIFF
--- a/js/ui/payg.js
+++ b/js/ui/payg.js
@@ -660,7 +660,8 @@ class PaygNotifier extends GObject.Object {
 
     notify(secondsLeft) {
         // Only notify when in an regular session, not in GDM or initial-setup.
-        if (Main.sessionMode.currentMode != 'user') {
+        if (Main.sessionMode.currentMode != 'user' &&
+            Main.sessionMode.currentMode != 'endless') {
             return;
         }
 
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -1429,7 +1429,8 @@ var ScreenShield = class {
         // end up loging the screen for not regular sessions (e.g. initial-setup).
         let shouldLockForPayg = Main.paygManager.isLocked &&
             (Main.sessionMode.currentMode == 'user' ||
-             Main.sessionMode.currentMode == 'user-coding');
+             Main.sessionMode.currentMode == 'user-coding' ||
+             Main.sessionMode.currentMode == 'endless');
 
         if (!this._settings.get_boolean(LOCK_ENABLED_KEY) && !shouldLockForPayg)
             return;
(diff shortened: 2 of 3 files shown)


js/ui/payg.js
```diff
@@ -660,7 +660,7 @@ class PaygNotifier extends GObject.Object {
 
     notify(secondsLeft) {
         // Only notify when in an regular session, not in GDM or initial-setup.
-        if (Main.sessionMode.currentMode != 'user') {
+        if (Main.sessionMode.currentMode != 'endless') {
```
mwleeds: Why is the session mode `endless`? Is that one of our downstream patches? Maybe this should be a fixup commit if so.

mazen-asef: This is ref. to this change : https://phabricator.endlessm.com/T29401 - i could add this as a fixup commit, but the change is part of debian-master instead of master

mwleeds: Agreed, it wouldn't make sense to make this a fixup; the change that's on master is a patch copied from Ubuntu (https://github.com/endlessm/gnome-shell/pull/627/commits/7bff1f25b5a6ba03125a1af3b6494aeda1c38361). And now I see you already link to T29401 in the commit message. So LGTM but Georges would know better.

GeorgesStavracas:
I'm okay with landing this, but do you mind checking what Ubuntu does here? Do they ship a custom patch similar to this one? We're mostly following their lead here.

mazen-asef:
@GeorgesStavracas It seems that the ubuntu fork does not have the js/ui/automaticUpdates.js, nor any instance of the 'user' sessionMode being compared anywhere. Maybe `js/ui/automaticUpdates.js` be upstreamed?

GeorgesStavracas:
So Ubuntu does patch GNOME Shell to check for the `ubuntu` session?

mazen-asef:
It doesn't seem like it - in the three projects we're the only one checking current sessionMode to assert  this, and both gnome and ubuntu's gnome-shell doesn't have automaticUpdates, at least not on master.

mazen-asef:
It seems also that automatic updates piece is not yet upstreamable since metered data isn't in a good shape upstream as of now, so that kinda explains why it's missing from GNOME/gnome-shell and ubuntu/gnome-shell.

mazen-asef:
Force-pushed with @andrunko suggestion to preserve older session and add ours - this should simplify any upstream/downstream change on `js/ui/status/automaticUpdates.js`

PR #641, opened 2020-04-01
fixup! window-tracker: associate toolbox windows with target application

danigm:
This patch removes the check to ensure that the hack mode is enabled
because that setting is not in the shell anymore. It was moved to the
extension.

We could try to use the shell extension dbus service to check if the hack
mode is enabled, but I think that it's not needed because it'll make the
code a bit complex and the performance improvement will not be great.
To do that we'll need to create a new dbus proxy and keep it alive, so it's
simpler and easier to just continue and try to create the toolbox proxy
for all windows even if the hack mode is disabled.

https://phabricator.endlessm.com/T28568


GeorgesStavracas:
LGTM.

PR #643, opened 2020-04-02
app-cache: Fix cache for folder translations

andrunko:
The app-cache code currently stores the folder translations in a hash
that can be accessed via shell_util_get_translated_folder_name().
This hash uses the filename (inc. extension) for the "desktop-directory"
as key which causes an issue when trying to find the translation
on AppDisplay._findBestFolderName() which gets categories (folder names)
from the app info which doesn't contain the ".directory" extension.

Fix that by storing the filename without extension as the hash key for
the cached folder translations.

https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1168

https://phabricator.endlessm.com/T29690

### DIFF
--- a/src/shell-app-cache.c
+++ b/src/shell-app-cache.c
@@ -109,11 +109,17 @@ load_folder (GHashTable *folders,
 
   while ((name = g_dir_read_name (dir)))
     {
+      g_autofree gchar *stripped_name = NULL;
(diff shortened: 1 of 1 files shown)


andrunko:
Updated after upstream review (only cosmetic changes).

PR #645, opened 2020-04-03
Fix icon grid icon scaling when swtiching from hidpi <-> lowdpi

andrunko:
https://phabricator.endlessm.com/T29687


andrunko:
Ah, the first commit here reapplies an upstream change that we reverted (upstream didn't revert it) and we need it to properly scale folder icons.

GeorgesStavracas:
Superseded by https://github.com/endlessm/gnome-shell/pull/648.

PR #648, opened 2020-04-03
T29687 (alternative solution)

GeorgesStavracas:
This one is future proof! It uses the solution proposed upstream, and reverts the revert instead of cherry-picking the original commit.

https://phabricator.endlessm.com/T29687

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1225,11 +1225,12 @@ class FolderView extends BaseAppView {
         });
         layout.hookup_style(icon);
         let subSize = Math.floor(FOLDER_SUBICON_FRACTION * size);
+        let scale = St.ThemeContext.get_for_stage(global.stage).scale_factor;
 
         let numItems = this._orderedItems.length;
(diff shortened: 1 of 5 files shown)


andrunko:
Ok, I discussed this with @GeorgesStavracas on Slack and the thing is that this branch is essentially the same as the original solution on #645, with the following diffs:
- this one reverts the revert instead of cherry-picking original patch as mentioned in the description
- this one doesn't redraw the icons with the correct size until logging out and in again after changing the scale settings

@GeorgesStavracas said that the fix proposed on #645 to force recreate the icon if the scale changes is not the correct one and that the issue is on the StWidget not propagating the theme changes correctly. 

I think we should either use a similar solution to #645 as workaround or fix the underlying issue on StWidget but I prefer a workaround (as the one provided by #645) than the bug).

GeorgesStavracas:
I'm fairly confident that these fixes are correct now. They're under upstream review, but I don't believe the final code will be radically different, if at all.

PR #649, opened 2020-04-03
fixup! Adjust app grid icons to Endless

andrunko:
These values now are getting overriden on _app-grid.scss, let's restore the same values used on eos3.7.

https://phabricator.endlessm.com/T28780

### DIFF
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -399,18 +399,25 @@ popup-separator-menu-item {
 
 .app-well-app {
     .overview-icon { @extend %app-well-icon; }
-    & > .overview-icon.overview-icon-with-label {
-        /* Use defaults here to "remove" the values set in _common.scss */
-        padding: 0px;
+    .overview-icon.overview-icon-with-label {
+        padding: 6px;
         spacing: 6px;
-    }
-}
+        > StBoxLayout {
+            padding: 0px;
+            spacing: 0px;
+        }
+     }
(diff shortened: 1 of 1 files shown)


andrunko:
Screenshot on master with these changes applied:
screenshot-master

Screenshot on 3.7.8:
screenshot-3 7

Note also that in both 3.7.8 and with these changes here, the app folder icons are slightly bigger than the normal icons (1-2px).

You will also notice that the icongrid is a bit shifted compared to 3.7, starting with the search entry being a bit higher on screen on master (with or without these changes) - this should be tracked separately if needed.

data/theme/gnome-shell-sass/_endless.scss
```diff
@@ -399,18 +399,25 @@ popup-separator-menu-item {
...
     .overview-icon { @extend %app-well-icon; }
-    & > .overview-icon.overview-icon-with-label {
-        /* Use defaults here to "remove" the values set in _common.scss */
-        padding: 0px;
+    .overview-icon.overview-icon-with-label {
+        padding: 6px;
         spacing: 6px;
-    }
-}
+        > StBoxLayout {
```
GeorgesStavracas, line 405: Isn't it missing a `&` here?

andrunko: Copied from https://github.com/endlessm/gnome-shell/blob/master/data/theme/gnome-shell-sass/widgets/_app-grid.scss#L29, tested and it works fine.

PR #652, opened 2020-04-07
T29698

GeorgesStavracas:
> Money, it's a crime
> Share it fairly but don't take a slice of my pie
> Money, so they say
> Is the root of all evil today
> But if you ask for a raise it's no surprise that they're
> Giving none away, away, away

This pull request is a courtesy of Nina the Dogga

nina no pc

https://phabricator.endlessm.com/T29698

### DIFF
--- a/js/misc/paygManager.js
+++ b/js/misc/paygManager.js
@@ -75,8 +75,9 @@ var PaygManager = GObject.registerClass({
         'code-expired': {},
         'code-format-changed': {},
         'enabled-changed': { param_types: [GObject.TYPE_BOOLEAN] },
-        'expiry-time-changed': { param_types: [GObject.TYPE_INT64] },
+        'expiry-time-changed': { param_types: [GObject.TYPE_UINT64] },
         'initialized': {},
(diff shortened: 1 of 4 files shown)


js/misc/paygManager.js
```diff
@@ -77,6 +77,7 @@ var PaygManager = GObject.registerClass({
         'enabled-changed': { param_types: [GObject.TYPE_BOOLEAN] },
         'expiry-time-changed': { param_types: [GObject.TYPE_INT64] },
         'initialized': {},
+        'rate-limit-end-time-changed':  { param_types: [GObject.TYPE_INT64] },
```
mwleeds, line 80: Should this be `TYPE_UINT64`?

mwleeds: And if it should, I guess `expiry-time-changed` should be too

mwleeds:
LGTM other than the one question

js/misc/paygManager.js
GeorgesStavracas: That's right, good catch!

PR #656, opened 2020-04-14
fixup! payg: Add success screen

mwleeds:
Fix "JS ERROR: this._dialog is null" in the log after a valid code is
applied using the tray applet.


mwleeds:
I didn't create a ticket for this, seems like a small thing.

andrunko:
> I didn't create a ticket for this, seems like a small thing.

@mwleeds just a reminder to cherry-pick this change to eos3.8 once tested on master.

PR #657, opened 2020-04-15
T29763

GeorgesStavracas:
We can already be in the unlock-dialog session mode when PAYG detects
the code expired, in which case we never push the unlock-dialog-payg
mode. Even if we did, because there was a dialog already, we'd never
actually create the PAYG unlock dialog.

Fix both with this commit. The first problem is fixed by properly
checking the session mode, and the second is fixed by destroying
the current dialog if PAYG needs to show its own.

https://phabricator.endlessm.com/T29763

### DIFF
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -143,7 +143,10 @@ var ScreenShield = class {
 
         this._syncInhibitor();
 
-        Main.paygManager.connect('code-expired', () => { this.lock(true) });
+        Main.paygManager.connect('code-expired', () => {
+            this._clearCurrentDialog();
+            this.lock(true);
+        });
         Main.paygManager.connect('expiry-time-changed', () => {
             let userManager = AccountsService.UserManager.get_default();
             let user = userManager.get_user(GLib.get_user_name());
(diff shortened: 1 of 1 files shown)


js/ui/screenShield.js
```diff
@@ -187,6 +191,13 @@ var ScreenShield = class {
         });
     }
 
+    _clearCurrentDialog() {
```
mwleeds, line 194: nit: a few lines above there's a block that could be replaced with `this._clearCurrentDialog()`

js/ui/screenShield.js
```diff
@@ -613,17 +624,20 @@ var ScreenShield = class {
         if (this._activationTime == 0)
             this._activationTime = GLib.get_monotonic_time();
 
-        if (Main.sessionMode.currentMode !== 'unlock-dialog' &&
+        if (Main.sessionMode.currentMode !== 'unlock-dialog' ||
```
mwleeds, line 627: Any current mode is going to be not equal to one or the other of `unlock-dialog` and `unlock-dialog-payg`, so maybe we just don't need this `if`?

js/ui/screenShield.js
GeorgesStavracas: Indeed. Done that.

js/ui/screenShield.js
GeorgesStavracas: Good point. The if is entirely useless now. Thanks for pointing that out.

PR #661, opened 2020-04-25
T29527 Fix resuming from suspend during FBE

mwleeds:
https://phabricator.endlessm.com/T29527


mwleeds:
The latter two commits here are cherry-picks from upstream.

PR #664, opened 2020-05-05
passwordMode: add keyboard layout switcher for password entries

GeorgesStavracas:
Fallout from the last rebase.

https://phabricator.endlessm.com/T29997

### DIFF
--- a/js/ui/status/keyboard.js
+++ b/js/ui/status/keyboard.js
@@ -1,7 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 /* exported InputSourceIndicator */
 
-const { Clutter, Gio, GLib, GObject, IBus, Meta, Shell, St } = imports.gi;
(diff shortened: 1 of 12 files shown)


GeorgesStavracas:
I couldn't figure out yet why sometimes the current layout name shows as a `.` in dialogs, but I guess that would be unrelated to this patch?

GeorgesStavracas:
Fixed

js/ui/status/keyboard.js
```diff
@@ -488,6 +497,45 @@ var InputSourceManager = class {
             this._updateMruSettings();
     }
 
+    get PasswordModeEnabled() {
+        return this.passwordModeEnabled;
```
andrunko, line 501: This look strange to me, why do we have 2 getters and setters for this property? And why one of them uses `this._passwordModeEnabled` while the other uses `this.passwordModeEnabled`? Same is valid for the setter below.

GeorgesStavracas: One is the D-Bus property implementation, the other is the JavaScript property.

andrunko: I see, thanks.

PR #670, opened 2020-05-21
T29763 payg unlock dialog fixes

mwleeds:
https://phabricator.endlessm.com/T29763

### DIFF
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -320,7 +320,12 @@ var ScreenShield = class {
             GLib.Source.set_name_by_id(this._lockTimeoutId, '[gnome-shell] this.lock');
         }
 
-        this._activateFade(this._longLightbox, STANDARD_FADE_TIME);
+        // Leave the screen on for the PAYG unlock code entry. It's difficult
+        // to distinguish the "lock due to credit expiration" case from the
+        // "lock due to idle timeout" one, and we don't want a blank screen in
+        // the former which includes right after the FBE.
+        if (!Main.paygManager.isLocked)
+            this._activateFade(this._longLightbox, STANDARD_FADE_TIME);
     }
 
     _activateFade(lightbox, time) {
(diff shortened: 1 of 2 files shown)


js/ui/screenShield.js
```diff
@@ -640,6 +640,9 @@ var ScreenShield = class {
             if (Main.sessionMode.currentMode !== 'unlock-dialog-payg' &&
                 Main.paygManager.isLocked)
                 Main.sessionMode.pushMode('unlock-dialog-payg');
+
+            this._clearCurrentDialog();
+            this._ensureUnlockDialog(true);
```
andrunko, line 650: A couple things:
- I think we should probably call this only if we are changing the mode in either conditionals above, otherwise every time we lock the screen in the "normal" case (without payg) we would be deleting and recreating the dialog? Also should we check whether `_ensureUnlockDialog()` fails as done above at the start of the method?
- We should also probably move the call to `_ensureUnlockDialog()` at the start of the method to the "else" clause here (when `this._isGreeter` is true).

Btw, the commit msg adding this change says `The reason is that in screenShield.animate()` and I guess you meant `screenShield.activate()` here?

mwleeds: > A couple things:
> 
>     * I think we should probably call this only if we are changing the mode in either conditionals above, otherwise every time we lock the screen in the "normal" case (without payg) we would be deleting and recreating the dialog?

You're right, I moved it up into the if block.

>Also should we check whether `_ensureUnlockDialog()` fails as done above at the start of the method?

No, because we know the unlock-dialog-payg session mode supports unlocking, so there's no chance it will fail.
> 
>     * We should also probably move the call to `_ensureUnlockDialog()` at the start of the method to the "else" clause here (when `this._isGreeter` is true).
No, I don't think so, because that would regress T29527 (initial setup authentication error). We need the early return for session modes that don't support locking.
> 
> 
> Btw, the commit msg adding this change says `The reason is that in screenShield.animate()` and I guess you meant `screenShield.activate()` here?

Yes, that was a mistake.

PR #672, opened 2020-05-29
T29994 Focus PAYG unlock code entry by default

mwleeds:
https://phabricator.endlessm.com/T29994

### DIFF
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -459,7 +459,6 @@ var ScreenShield = class {
         }
 
         this._dialog.allowCancel = allowCancel;
-        this._dialog.grab_key_focus();
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


js/ui/screenShield.js
```diff
@@ -459,7 +459,6 @@ var ScreenShield = class {
         }
 
         this._dialog.allowCancel = allowCancel;
-        this._dialog.grab_key_focus();
```
GeorgesStavracas, line 462: Doesn't this interfere with the regular unlock dialog? Specifically, typing immediately after turning the screen on.

mwleeds: Yeah I thought it might to but in practice it doesn't. I think that's because Main.pushModal() causes key focus to be grabbed for whatever is becoming modal.

mwleeds: But despite what I said in the commit message, maybe it is best to keep this patch downstream since it doesn't provide any benefit upstream.

PR #676, opened 2020-07-02
eos-link-feedback: Point to forum instead of eos-exploration-center feedback page

andrunko:
https://phabricator.endlessm.com/T30295


andrunko:
This change can be squashed with 1b87d0df19c098ee50931bf7cfff159c180e175b during next rebase or if you prefer I can change the commit to a fixup.

andrunko:
Force pushed to use a fixup commit instead.

PR #678, opened 2020-07-10
fixup! project: Add eos-link-feedback back

andrunko:
Use the default browser instead of chromium - we don't need to use chromium anymore given we don't rely on the exploration center existing anymore and instead just redirect users to the forum (which should work on all browsers).

https://phabricator.endlessm.com/T30295

### DIFF
--- a/data/eos-link-feedback.desktop.in.in
+++ b/data/eos-link-feedback.desktop.in.in
@@ -2,7 +2,7 @@
 Version=1.0
 Name=Feedback
 Type=Application
-Exec=chromium-browser https://community.endlessos.com/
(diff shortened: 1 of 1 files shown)


wjt:
Yes, and – rather than readding this after landing vanilla shell on master, can we move this file to another repo (`eos-meta/eos-tech-support` perhaps) so we don't need to reapply it every 6 months?

andrunko:
> Yes, and – rather than readding this after landing vanilla shell on master, can we move this file to another repo (`eos-meta/eos-tech-support` perhaps) so we don't need to reapply it every 6 months?

Ack, I will keep that in mind.

PR #683, opened 2020-08-04
List eos-panel@endlessm.com in Endless session

andrunko:
https://phabricator.endlessm.com/T30445


andrunko:
Depends on https://github.com/endlessm/eos-meta/pull/629.

PR #686, opened 2020-08-07
sessionMode: Enable the panel extension on gdm and lockscreen mode

andrunko:
We want the customized panel to be available on gdm and lockscreen mode as well.

https://phabricator.endlessm.com/T30445


andrunko:
Force pushed after rebasing on latest master, no actual change.

PR #688, opened 2020-08-13
Enable eos-desktop extension on lock screen as well

GeorgesStavracas:
Keep the eos-desktop extension enabled on the lock screen.
Nowadays the desktop extension can handle tearing down
fine, but the act of destroying and recreating the app
grid is a heavy one and incurs stuttering when unlocking
the screen.

https://phabricator.endlessm.com/T30623

### DIFF
--- a/js/ui/sessionMode.js
+++ b/js/ui/sessionMode.js
@@ -60,7 +60,7 @@ const _modes = {
         isLocked: true,
         unlockDialog: undefined,
         allowExtensions: true,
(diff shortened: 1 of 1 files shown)


andrunko:
I am not 100% sure but I think I saw some cases where even though the panel extension was always enabled as you did here it would still be disabled and re-enabled when the session mode changes - I think because of the order of the extensions that got enabled/disabled... In any case it shouldn't impact here, but we should make sure the extension supports disabling (also because users could disable the extension manually which I think takes precedence over the configured modes here).

PR #690, opened 2020-08-17
Re-add password reset support

andrunko:
This includes the following:
- **new** ecf06ddcf authPrompt: Properly get oVirt service name
  - New commit to fix an issue with upstream - submitted here https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1414
- **pick** 86c88d49f Update default background color for login and lock screens
  - no conflict
  - this adds the machinery for the endless style/CSS changes
- **pick** da09ad5b5 passwordReset: Implement Endless-specific password reset feature
  - small conflict mainly on build system
  - changed patch slightly to also include `Config.LOCALSTATEDIR` which is used by `customerSupport.js` (on 3.8 this config entry was added as part of a commit adding the IconGridLayout, which we won't be including anymore now that we have the desktop extension)
  - for reference this commit already adds `Config.PKGDATADIR` so I thought it was a good place to also include `Config.LOCALSTATEDIR`
- **pick** f6d900105 passwordReset: Reimplement password hint feature
  - no conflict
- **pick** 58913a7a customerSupport: Manage vendor-customer-support.ini from its own class
  - small conflict mainly on build system

Should be merged together with https://github.com/endlessm/gnome-shell/pull/689.

https://phabricator.endlessm.com/T30648

PR #691, opened 2020-08-18
Re-add PAYG support

andrunko:
Changes here depend on (are on top of) https://github.com/endlessm/gnome-shell/pull/690.

This PR includes the following:
- **pick** 7d4512ea5 payg: Initial implementation of the PAYG manager component
  - small conflict mainly on build system
- **pick** ca6825e79 payg: Add a new screen to allow unlocking PAYG machines
  - also changed the default panel layout for the "unlock-dialog-payg" session mode to not include the "powerMenu" indicator (as it was in the original commit). The reason is that upstream does not have the "powerMenu" indicator (functionality included in the "aggregateMenu" upstream which is also included in the default layout when on "unlock-dialog-payg" session mode) and we want this layout to work if the panel extension is disabled (the panel extension will set the layout as it was before, but that only works if the extension is enabled)
- **pick** 5fb80182d payg: Wait for the PaygManager before showing the screenshield's dialog
  - no conflict
- **pick** 315c9b320 payg: Make sure to initialize the users list after startup has completed
  - no conflict
- **pick** 46ff68c19 payg: Integrate the PAYG unlock screen with the login screen
  - no conflict
- **pick** 07354b7c6 payg: Support image-specific overrides for text in the unlock dialog
  - no conflict
- **pick** e7f14449a payg: Show image-specific instructions for PAYG, if available
  - no conflict
- **pick** 86b0763b1 payg: tweak code entry field for longer codes
  - no conflict
- **pick** b1e871703 js/ui/payg.js: PayGo UI utilities
- **pick** c4830ceaf payg: factor out reminder notifications
  - no conflict
- **pick** 19c26bf40 status/payg: new applet for PAYG account details and actions
- **pick** b5aee404f status/payg: add menu action to apply a code before PayGo time is expired
  - no conflict
- **pick** bdf9069be payg: Add success screen
  - no conflict
- **pick** c990764b7 eos-payg: Add prefix and sufix characters for code insert
  - no conflict
- **pick** 1015c4ec5 eos-payg: Better segregate Pay As You Go classes
  - no conflict
- **pick** 09a6fb4ee eos-payg: Add the character enforcement for payg-tray
  - no conflict
- **pick** a4e24227a paygManager: Update iface for ImpendingShutdown signal
  - no conflict
- **pick** 8faf6c84d payg: Add AccountID support for the UI
  - no conflict
- **pick** ea9edb103 payg: Add AccountID info on the applet
  - no conflict
- **pick** ddc404d63 fixup! payg: Add a new screen to allow unlocking PAYG machines
  - no conflict
  - squashed in original commit
- **pick** e3775d4dc fixup! payg: Initial implementation of the PAYG manager component
  - no conflict
  - squashed in original commit
- **pick** 969bb35eb fixup! js/ui/payg.js: PayGo UI utilities
  - no conflict
  - squashed in original commit
- **pick**  d30cb337a fixup! payg: factor out reminder notifications
  - no conflict
  - squashed in original commit
- **pick** e9d48d2d8 fixup! status/payg: add menu action to apply a code before PayGo time is expired
  - no conflict
  - squashed in original commit
- **pick** 0daaf5f88 fixup! payg: Initial implementation of the PAYG manager component
  - no conflict
  - squashed in original commit
- **pick** 8c6f20b03 fixup! payg: Add a new screen to allow unlocking PAYG machines
  - no conflict
  - squashed in original commit
- **pick**  1042dea32 fixup! payg: factor out reminder notifications
  - no conflict
  - squashed in original commit
- **pick** 150a61fff fixup! payg: Add success screen
  - no conflict
  - squashed in original commit
- **pick** 0d2f4de2f fixup! payg: Add a new screen to allow unlocking PAYG machines
  - small conflict due to upstream changes
  - squashed in original commit
- **pick** 8e1edb516 fixup! payg: Add success screen
  - no conflict
  - squashed in original commit
- **pick** 4f0403a12 fixup! payg: Add a new screen to allow unlocking PAYG machines
  - no conflict
  - squashed in original commit
- **pick** eac6b4c3a payg: Fix a couple comments
  - no conflict
- **pick** 90b73b11a payg: Properly display credit expired in tray applet
  - no conflict
- **pick** 67fde3d03 payg: Fix ngettext usage
  - no conflict
- **pick** 22d11360b payg: Don’t calculate times if PAYG is disabled
  - no conflict
- **pick** 2c6d99c92 screenShield: Properly show PaygUnlockDialog on login screen
  - no conflict
- **pick** aa5a452c0 screenShield: Lock for PAYG after FBE
  - no conflict
- **pick** 75c4361cf screenShield: Don't fade to black for payg lock
  - no conflict
- **pick** e37767c0b payg: Make Success button clickable
  - no conflict
- **pick** 6f892ecd1 screenShield: Don't grab dialog key focus
  - no conflict
- **pick** 105e5ace0 paygUnlockDialog: Focus entry by default
  - no conflict
- **pick** aef7a8ca9 screenShield: Don't try to use an undefined dialog object
  - no conflict

https://phabricator.endlessm.com/T30647

### DIFF
--- a/data/dbus-interfaces/com.endlessm.Payg1.xml
+++ b/data/dbus-interfaces/com.endlessm.Payg1.xml
@@ -1,0 +1,21 @@
+<node>
+  <interface name="com.endlessm.Payg1">
+    <method name="AddCode">
+      <arg type="s" direction="in" name="code"/>
+      <arg type="x" direction="out" name="time_added"/>
+    </method>
+    <method name="ClearCode" />
+    <signal name="Expired" />
+    <signal name="ImpendingShutdown">
+      <arg type="i" name="seconds_remaining"/>
+      <arg type="s" name="shutdown_reason"/>
+    </signal>
+    <property name="ExpiryTime" type="t" access="read"/>
+    <property name="Enabled" type="b" access="read"/>
+    <property name="RateLimitEndTime" type="t" access="read"/>
+    <property name="AccountID" type="s" access="read"/>
+    <property name="CodeFormat" type="s" access="read"/>
+    <property name="CodeFormatPrefix" type="s" access="read"/>
+    <property name="CodeFormatSuffix" type="s" access="read"/>
+  </interface>
+</node>
--- a/data/gnome-shell-dbus-interfaces.gresource.xml
+++ b/data/gnome-shell-dbus-interfaces.gresource.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <gresources>
   <gresource prefix="/org/gnome/shell/dbus-interfaces">
+    <file preprocess="xml-stripblanks">com.endlessm.Payg1.xml</file>
     <file preprocess="xml-stripblanks">net.hadess.SensorProxy.xml</file>
     <file preprocess="xml-stripblanks">org.freedesktop.Application.xml</file>
     <file preprocess="xml-stripblanks">org.freedesktop.bolt1.Device.xml</file>
--- a/data/gnome-shell-theme.gresource.xml
+++ b/data/gnome-shell-theme.gresource.xml
@@ -32,5 +32,10 @@
     <file>toggle-on.svg</file>
     <file>toggle-on-dark.svg</file>
     <file>toggle-on-hc.svg</file>
+
+    <!-- Endless-specific resources beyond this point -->
+
+    <file>payg-near-expiration-symbolic.svg</file>
+    <file>payg-normal-symbolic.svg</file>
   </gresource>
 </gresources>
--- a/data/theme/gnome-shell-sass/_endless.scss
+++ b/data/theme/gnome-shell-sass/_endless.scss
@@ -30,6 +30,109 @@
   }
 }
 
+// Pay As You Go unlock screen (based on .login-dialog)
+
+$success_time_payg_color: #02b842;
+
+.unlock-dialog-payg {
+  border: none;
+  background-color: transparent;
+
+  .modal-dialog-button-box { spacing: 3px; }
+  .modal-dialog-button {
+    padding: 3px 18px;
+    &:default {
+      @include button(normal,$c:$selected_bg_color);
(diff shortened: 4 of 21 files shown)


mwleeds:
I definitely agree the history here should be cleaned up, and the scheme you proposed looks reasonable. One patch that might be worth keeping on its own is acc2f16def44764445d39877a377f5f43bac2dd8 since it has useful info in the commit message. But up to you, we can always use the eos3.8 branch to see the history if needed.

andrunko:
Thanks both, tried splitting as suggested above, the branch should be ready for review again. 

The diff compared to the previous version can be found [here](https://github.com/endlessm/gnome-shell/compare/6c3e554c73319522f27ed5f60dade3c789d69d3e..6e58f587fe575079f3561fb1ed201535980b3131), but in summary the changes include:
- Use `const` instead of `let` where applicable
- Add some cleanup code to fix a race found during tests - the race would happen when using the notification to add new code, which would call ` PaygManager.addCode` with a callback, then ` PaygManager.addCode` would invoke `proxy.AddCodeRemote` which in turn would emit properties changed in the proxy with a new expire time and then call `clearNotification()` (destroying the notification). By the time the callback passed to `PaygManager.addCode` was invoked the notification was already destroyed and that was causing issues. I fixed this by invoking `clearNotification()` on an idle callback when the expire time changes.
  - The new `destroy()` method on PaygManager is unused but added for completeness
- Removed unused CSS from `_high-contrast-colors.scss` (all that is needed is on `_endless.scss`)
- Updated some copyright headers
- Changed how the code checks if the session is a regular session to use `SessionMode.hasOverview` instead of comparing session mode ids

Tested the updated changes again and didn't find any regression compared to 3.8.

@mwleeds I didn't keep https://github.com/endlessm/gnome-shell/commit/acc2f16def44764445d39877a377f5f43bac2dd8 separate as I think (as you said) we can refer to the 3.8 branch in the future if needed.

PR #692, opened 2020-08-19
General improvements/fixes for PAYG indicator

andrunko:
This includes the following updates:
- Always keep status in sync when session mode changes (e.g. the menu sensitivity is correctly updated after unlocking the screen if the lock code was entered while the screen was locked)
- Always show account id menu item if supported
- Properly handle destroy (The parent class SystemIndicator doesn't define '_onDestroy')
- Reuse existing payg manager instance

https://phabricator.endlessm.com/T30647

### DIFF
--- a/js/ui/status/payg.js
+++ b/js/ui/status/payg.js
@@ -35,7 +35,7 @@ class PaygIndicator extends PanelMenu.SystemIndicator {
     _init() {
         super._init();
 
-        this._paygManager = new PaygManager.PaygManager();
+        this._paygManager = Main.paygManager;
         this._indicator = this._addIndicator();
         this._item = new PopupMenu.PopupSubMenuMenuItem('', true);
         this._paygNotifier = new Payg.PaygNotifier();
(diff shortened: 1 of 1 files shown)


js/ui/status/payg.js
```diff
@@ -64,21 +70,35 @@ class PaygIndicator extends PanelMenu.SystemIndicator {
...
+        this._sessionModeUpdatedId = Main.sessionMode.connect('updated',
+            this._sync.bind(this));
+
+        this.connect('destroy', this._onDestroy.bind(this));
+
+        this._sync();
     }
 
     _onDestroy() {
-        super._onDestroy();
```
GeorgesStavracas, line 75: You still need to preserve this chainup, don't you?

andrunko: Not really, this was actually wrong and this method never invoked as the base class doesn't define `_onDestroy()` or connects to the destroy signal.

andrunko: The base class being `PanelMenu.SystemIndicator` which inherits from `St.BoxLayout`.

GeorgesStavracas: Okay

PR #696, opened 2020-08-20
Rebase passwordMode on new shell

andrunko:
This includes the following:
- **pick** 4e5f756a0 passwordMode: add keyboard layout switcher for password entries
- **pick** 64842f1e9 fixup! passwordMode: add keyboard layout switcher for password entries
  - squashed in the commit above
- Apart from that it also includes some changes to use `const` instead of `let` where applicable and a change to set the style class `prompt-dialog-password-layout` for the boxes where the keyboard layout menu is added, such that the UI adds a spacing (default is 8px for this style class) between the entry and the keyboard layout menu.

https://phabricator.endlessm.com/T30649


andrunko:
Fwiw, the debian changes to add a dep to xkbcommon are already on debian-master.

PR #697, opened 2020-08-21
Re-add metrics support to new shell

andrunko:
This PR includes:
- **pick** dfa64aada4  shell-app-system: Report an event when an application opens/closes
  - sends `SHELL_APP_IS_OPEN_EVENT` metric
- **pick** 1e349a1304  viewSelector: Record metric when user initiates a desktop search
  - reworked the patch a bit to adjust to upstream changes
  - sends  `EVENT_DESKTOP_SEARCH` metric

Tested and both events above are sent as expected.

Apart from those we also had the following metrics on 3.8:
- `SHELL_APP_ADDED_EVENT`:  when an app was added to the grid
- `SHELL_APP_REMOVED_EVENT`: as above but when removed from the grid

Given we don't support adding/removing apps to/from grid anymore I did not include those.

https://phabricator.endlessm.com/T30661

### DIFF
--- a/js/ui/viewSelector.js
+++ b/js/ui/viewSelector.js
@@ -1,7 +1,8 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 /* exported ViewSelector */
 
-const { Clutter, Gio, GObject, Meta, Shell, St } = imports.gi;
+const { EosMetrics, Clutter, Gio,
+        GLib, GObject, Meta, Shell, St } = imports.gi;
 const Signals = imports.signals;
 
 const AppDisplay = imports.ui.appDisplay;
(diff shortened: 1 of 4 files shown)


andrunko:
> > I think I can add the second commit to the eos-desktop extension. I'd rather do that instead of adding a downstream commit.
> 
> Ok, I can drop it from here or would you prefer to check if feasible to do in the desktop extension first?

On second thought I think we should stick to patching the shell for these metrics changes - I believe we still want the metrics even if the user has disabled the extensions (e.g. in case we ship a standard GNOME session as option).

wjt:
Azafea currently ignores `EVENT_DESKTOP_SEARCH` so it's rather moot. `SHELL_APP_IS_OPEN_EVENT` is more important. I wouldn't make maintenance of this harder for the benefit of an extremely non-default configuration.

andrunko:
> Azafea currently ignores `EVENT_DESKTOP_SEARCH` so it's rather moot. `SHELL_APP_IS_OPEN_EVENT` is more important. I wouldn't make maintenance of this harder for the benefit of an extremely non-default configuration.

Tbh I think the patch for EVEN_SEARCH is really simple to maintain (and the PR is ready) so I would keep it on the shell for now.

PR #699, opened 2020-08-29
eos-launch: Add launcher for 'endlessm-app://' schemes

andrunko:
Reworked changes to not depend on downstream `AppActivation` support. The `AppActivation` support was responsible for handling the integration with the speedwagon as well as to ensure the `X-Endless-Maximized` desktop key was respected. Both speedwagon and `X-Endless-Maximized` will be dropped on 3.9 so let's simplify the launcher impl.

Also fixed a small issue where `eos-launch` would try launching (and report an error) the locale specific apps even if the app requested had already launched successfully (can be easily reproduced on 3.8 by clicking on "Give Us Feedback" on Help Center and looking at the journal).

https://phabricator.endlessm.com/T30293
https://phabricator.endlessm.com/T30172

### DIFF
--- a/data/dbus-interfaces/meson.build
+++ b/data/dbus-interfaces/meson.build
@@ -1,4 +1,5 @@
 dbus_interfaces = [
+  'org.gnome.Shell.AppLauncher.xml',
   'org.gnome.Shell.Extensions.xml',
   'org.gnome.Shell.Introspect.xml',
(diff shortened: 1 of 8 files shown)


GeorgesStavracas (approved):
It's a bit disappointing that this is still done in Shell, but so be it.

PR #703, opened 2020-09-08
Remove libanimation dep

danigm:
The wobbly window effect has been removed from the shell so this
dependency is not needed anymore.

https://phabricator.endlessm.com/T30737


GeorgesStavracas:
I have another GNOME Shell rebase where I drop this dependency. Are you okay with me closing this PR, and handling it myself?

danigm:
> I have another GNOME Shell rebase where I drop this dependency. Are you okay with me closing this PR, and handling it myself?

yes, sure, we can close this in that case.

PR #709, opened 2020-09-24
status/remoteAccess: Show remote access indicator on X11

andrunko:
On Endless we only enable X11 for now, let's make sure the remote access indicator (screen sharing and recording) is displayed on X11 as well.

https://phabricator.endlessm.com/T30172


andrunko:
Superseded by https://github.com/endlessm/eos-panel-extension/pull/19.

PR #713, opened 2020-10-08
Some small fixes to objects lifecycle

andrunko:
https://phabricator.endlessm.com/T30808

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -991,6 +991,12 @@ class AppDisplay extends BaseAppView {
             GLib.source_remove(this._scrollTimeoutId);
             this._scrollTimeoutId = 0;
         }
+
+        if (this._currentDialog && this._currentDialogDestroyId) {
+            this._currentDialog.disconnect(this._currentDialogDestroyId);
+            this._currentDialog = null;
+            this._currentDialogDestroyId = 0;
+        }
     }
 
     vfunc_map() {
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -2195,7 +2195,8 @@ var AppFolderDialog = GObject.registerClass({
             coordinate: Clutter.BindCoordinate.SIZE,
         }));
 
-        this._folder.connect('changed::name', () => this._syncFolderName());
+        this._folderNameChangedId = this._folder.connect(
+            'changed::name', () => this._syncFolderName());
         this._syncFolderName();
     }
 
@@ -2359,6 +2360,11 @@ var AppFolderDialog = GObject.registerClass({
         if (this._sourceMappedId) {
             this._source.disconnect(this._sourceMappedId);
             this._sourceMappedId = 0;
+        }
+
+        if (this._folderChangedId) {
+            this._folder.disconnect(this._folderNameChangedId);
+            this._folderNameChangedId = 0;
         }
 
         this._removePopdownTimeout();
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -1117,10 +1117,13 @@ class AppDisplay extends BaseAppView {
             let icon = this._items.get(id);
             if (!icon) {
                 icon = new FolderIcon(id, path, this);
-                icon.connect('apps-changed', () => {
+                const folderAppsChangedId = icon.connect('apps-changed', () => {
```
GeorgesStavracas, line 1126: Why? The signal is disconnected when the source object is disposed.

js/ui/appDisplay.js
```diff
@@ -2198,7 +2198,8 @@ var AppFolderDialog = GObject.registerClass({
             coordinate: Clutter.BindCoordinate.SIZE,
         }));
 
-        this._folder.connect('changed::name', () => this._syncFolderName());
+        this._folderNameChangedId = this._folder.connect(
+            'changed::name', () => this._syncFolderName());
```
GeorgesStavracas, line 2202: When does the folder dialog outlives the folder settings? All their lifetimes seem tightly tied

js/ui/appDisplay.js
andrunko: But iirc we don't destroy the icons explicitly when destroying the App display?

js/ui/appDisplay.js
andrunko: As above we never destory the folder explicitly when the FolderView/Dialog/Icon is destroyed and rely on GC.

js/ui/appDisplay.js
andrunko: I am certain I did hit this issue a few times while testing the drag&drop crash but haven't tested after the gjs update.

GeorgesStavracas: Clutter recursively destroys actors. In the one case the folder icon isn't added to the grid, it is immediately destroyed. There's just no way for a folder icon to outlive the icon grid. It shouldn't happen with older GJS, it shouldn't happen with newer GJS either.

js/ui/appDisplay.js
GeorgesStavracas: The dialog is explicitly destroyed together with the icon: https://github.com/endlessm/gnome-shell/blob/master/js/ui/appDisplay.js#L1895

andrunko: Yes, the dialog is but not this._folder (the gsettings).

js/ui/appDisplay.js
andrunko: Ok, I can drop this change but I'm 100% sure I hit this while testing and not only once but multiple times.

andrunko: Dropped this one and we can re-add at a later point if we hit it again - it could be because of the old gjs playing tricks with the GC/backtrace :P.

js/ui/appDisplay.js
andrunko: Kept this one but can drop also if you prefer.

andrunko: Fwiw, upstream/FolderIcon already connects to folder changed signal and disconnects when FolderIcon is destroyed, but not for the changed:name.

andrunko: https://github.com/endlessm/gnome-shell/blob/master/js/ui/appDisplay.js#L1887-L1888

andrunko:
This doesn't seem required (we are not facing crashes anymore), but I still think they are useful fixes/improvements, please let me know what you think and we can close or merge this one.

PR #716, opened 2020-11-18
Readd X-Flatpak-Renamed-From

GeorgesStavracas:
https://phabricator.endlessm.com/T31115

### DIFF
--- a/src/shell-app-system.c
+++ b/src/shell-app-system.c
@@ -56,6 +56,7 @@ enum {
 enum {
   APP_STATE_CHANGED,
   INSTALLED_CHANGED,
+  APP_INFO_CHANGED,
   LAST_SIGNAL
 };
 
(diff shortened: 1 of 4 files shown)


src/shell-app-system.c
```diff
@@ -33,6 +33,27 @@
...
 #define SHELL_APP_IS_OPEN_EVENT "b5e11a3d-13f8-4219-84fd-c9ba0bf3d1f0"
 
+/* Additional key used to map a renamed desktop file to its previous name.
+ * For instance, org.gnome.Totem.desktop contains:
+ *
+ *   X-Endless-Alias=totem
+ *
+ * (without the .desktop suffix).
+ */
+#define X_ENDLESS_ALIAS_KEY     "X-Endless-Alias"
```
andrunko, line 43: Iiuc we should drop X-Endless-Alias and only rely on X-Flatpak-RenamedFrom - we already dropped this (X-Endless-Alias) key for most (if not all) our apps that included it.

andrunko (changes requested):
I'm missing the changes to actually use `shell_app_system_lookup_alias` as it was done before 
(eos3.8) on IconGridLayout and AppFavorites. Are they no longer needed or are you planning on adding those in another place (e.g. desktop extension) or am I missing something?

Also wonder if we need to update the desktop extension migration to handle the renames.

src/shell-app-system.c
GeorgesStavracas: This was on purpose. I didn't want to change any of these patches the day before 3.9.1 freezes.

GeorgesStavracas:
Desktop extension being taken care of at https://github.com/endlessm/eos-desktop-extension/pull/38

src/shell-app-system.c
andrunko: Ok, fair enough, we can drop it later on next rebase if we get to upstream this or some alternative solution.

andrunko (approved):
LGTM.

PR #719, opened 2021-03-13
T31561

GeorgesStavracas:
Only dropped translations, they're conflicting.


wjt:
- We lost 706386d52dceeff5cd4a8c188015c5b57c790aa5 as part of this rebase which adds a bunch of new languages. The way I normally fix the conflicts in patches like that is to reapply the `LINGUAS` bit then regenerate the `.po` files and commit the result. But we could do that separately. I'll go make a "sort out translations" ticket for switching Transifex over to the eos4 branch.
- b807b5701bd35557614f52925f489c965d279ee6 can go, per [this doc](https://docs.google.com/document/d/1MMT1HTD-rZPJ7k20TYD3zX3MslSK2X_uGPB6Hxle9Kc/)
- Sounds like everything else was either already upstream or applied cleanly?

andrunko:
>     * We lost [706386d](https://github.com/endlessm/gnome-shell/commit/706386d52dceeff5cd4a8c188015c5b57c790aa5) as part of this rebase which adds a bunch of new languages. The way I normally fix the conflicts in patches like that is to reapply the `LINGUAS` bit then regenerate the `.po` files and commit the result. But we could do that separately. I'll go make a "sort out translations" ticket for switching Transifex over to the eos4 branch.
> 

Done that.

>     * [b807b57](https://github.com/endlessm/gnome-shell/commit/b807b5701bd35557614f52925f489c965d279ee6) can go, per [this doc](https://docs.google.com/document/d/1MMT1HTD-rZPJ7k20TYD3zX3MslSK2X_uGPB6Hxle9Kc/)
> 
Reverted that change, I plan to squash it while merging.

>     * Sounds like everything else was either already upstream or applied cleanly?
Looks like the case from what I see.

andrunko (approved):
LGTM, @wjt do you want to have a final look on latest changes before merging?

js/misc/eos-launch.js
```diff
@@ -0,0 +1,57 @@
...
+        return true;
+    } catch (error) {
+        logError(error, `Failed to launch application '${appName}'`);
+        return false;
+    }
+}
+
+var uri = ARGV[0];
+if (!uri) {
+    print('Usage: eos-launch endlessm-app://<application-name>\n');
```
wjt, line 50: I had hoped that this wouldn't be needed any more. I searched `/usr` on my Endless OS system and found a single usage in eos-gates. https://phabricator.endlessm.com/T28309 was the last time we tried and failed to drop it; back then it was the exploration centre that needed it. IIRC the disco feed also used to use it.

Let's not block this PR on that but can we file an EOS-4.0 task to remove its last use and then get rid of it?

wjt (approved):
LGTM to squash and force-push. I have not reviewed #718, nor tested either.

andrunko:
Landed manually.

PR #721, opened 2021-11-15
Revert to default background color from upstream

jprvita:
This partially reverts commit 9d0025bdfb6c92fcf4a69e99309abc6300718c1d
"Update default background color for login and lock screens".

For the new bootsplash animation we are going to stick with upstream's
current background color and texture.

https://phabricator.endlessm.com/T32724


GeorgesStavracas (approved):
Looks good!

PR #724, opened 2021-12-29
41.2

GeorgesStavracas:
# Pick

ede2dd225 Add an helper to detect current desktop
9b1d089f3 Don't allow endless mode extension to update
a48657506 Disable CSS test
0e4fa1f65 passwordReset: Implement Endless-specific password reset feature
8917d9046 passwordReset: Reimplement password hint feature
506616c6d shell-util: Add helper method to get system boottime
359ef9673 payg: Add PAYG infrastructure
f5cf685ac payg: Add a new screen to allow unlocking PAYG machines
87f60dabe payg: Add new session modes
acb6534d9 payg: Integrate with screen shield and GDM screen
bf14333e9 payg: Add panel indicator for PAYG account details and actions
74072c183 payg: Integrate panel indicator
318594769 passwordMode: add keyboard layout switcher for password entries
9f6ec41a7 networkAgent: Add switch button for metered connections
10120ea73 shell-app-system: Report an event when an application opens/closes
0612d1723 eos-launch: Add launcher for 'endlessm-app://' schemes
45ca8fc01 shell-app-system: Add support for X-Endless-Alias key in desktop files
7ddd508f2 shell-app-system: obey X-Flatpak-RenamedFrom
7bac88934 appDisplay: compare possibly-changed apps more thoroughly
5ce750d94 shell-app-system: extract deep comparison for GDesktopAppInfo
77a610896 appFavorites: respect X-Endless-Alias and X-Flatpak-RenamedFrom

# Drop

92f16bfd5 sessionMode: Enable the panel extension on gdm and lockscreen mode
 * No panel extension anymore

 350f77bd1 st: Implement more CSS gradient options
 * We don't have an Endless button anymore

9d0025bdf Update default background color for login and lock screens
3e2dbc8ea Revert to default background color from upstream

691f00aa7 iconGrid: Allocate preferred sizes if bigger than child size
 * Upstream

5b3868901 Add Transifex translation infrastructure
c92151f1b Add translation support for several new languages
5be764d8c New translations from Transifex
8523a1852 New translations from Transifex

74c50ac02 app-system: Use aggregate events
4b0914f1d app-system: Use new app usage event id
 * Squash with 10120ea73

pick 4cb3b6093 shell-app-system: Add support to identify apps that are starting up
pick 70038539b shell-app-system: Add a new app-info-changed signal
 * Unused function & signal

https://phabricator.endlessm.com/T32862

### DIFF
--- a/js/ui/appFavorites.js
+++ b/js/ui/appFavorites.js
@@ -97,6 +97,11 @@ class AppFavorites {
                 updated = true;
                 return newId;
             }
+            let newApp = appSys.lookup_alias(id);
+            if (newApp && newApp.get_id() !== id) {
+                updated = true;
+                return newApp.get_id();
+            }
             return id;
         });
         // ... and write back the updated desktop file names


GeorgesStavracas:
retest this please

meson_options.txt
```diff
@@ -24,7 +24,7 @@ option('man',
 
 option('tests',
   type: 'boolean',
-  value: true,
+  value: false,
```
wjt, line 27: Breaking due to a missing dependency on `gir1.2-eosmetrics`? Is this still needed since you added that in the -debian branch?

src/shell-app-system.c
```diff
@@ -611,3 +644,9 @@ shell_app_system_get_installed (ShellAppSystem *self)
 {
   return shell_app_cache_get_all (shell_app_cache_get_default ());
 }
+
+gboolean
+shell_app_system_has_starting_apps (ShellAppSystem *self)
```
wjt, line 761: I couldn't find any uses of this function in this tree. https://github.com/endlessm/eos-desktop/blob/ee7659e61a5dcb804065b107f68ca3b81ce1f10c/js/ui/workspaceMonitor.js#L64 is the last reference I can find.

I wonder what it's for?

wjt (commented):
I only had a quick look here

src/shell-app-system.c
GeorgesStavracas: You're right that it's unused. In fact, the app-info-changed signal also is. I think we can drop both.

meson_options.txt
GeorgesStavracas: No, it's broken *because* of this dependency. One of the CSS unit tests ends up trying to open a connection to the metrics daemon, which emits a warning and fails the tests.

GeorgesStavracas:
retest this please

PR #727, opened 2022-01-26
shell-app-system: Add support to identify apps that are starting up

GeorgesStavracas:
This will be used by eos-desktop-extension to properly determine when we can close the overview.

Related to https://phabricator.endlessm.com/T33020


GeorgesStavracas:
Not today - we'll deal differently on eos-desktop-extension.

PR #728, opened 2022-01-31
payg: Port PaygUnlockDialog to new MetaIdleAdd API

GeorgesStavracas:
This causes a regression where the PAYG unlock dialog doesn't
show up, which is pretty bad considering this screenshield is
what's shown during initial setup.

This commit should be squashed on commit 96cf95a6b "payg: Add PAYG
infrastructure" during the next rebase.

https://phabricator.endlessm.com/T33021


wjt (approved):
LGTM and corresponds clearly to 08fb067d2428e10cf0f7f6ee4848e07879d821d1 upstream.

PR #729, opened 2022-02-21
GNOME Shell 41.4 cherry picks

GeorgesStavracas:
Basically cherry-picks for all patches that were included in the `gnome-41` branch after we rebased on top of it.

No conflicts.

Did this as part of my investigation of https://phabricator.endlessm.com/T33000 (none of the cherry-picks fix it, but https://github.com/endlessm/gnome-shell/commit/d7722437289d83b2c96a28e6e183cfa19e9a9b5f is somehow related, and apparently this issue is also present upstream)


wjt:
SGTM, but I wonder whether merging the relevant tag (or commit) from the gnome-41 branch would be easier to manage?

PR #732, opened 2022-04-05
payg: UI support for backends with different code lengths

jprvita:
Make use of the CodeLength D-Bus property on eos-paygd's interface to
inform the user how many characters should be entered.

This commit should be squashed on commit 96cf95a6b "payg: Add PAYG
infrastructure" during the next rebase.

https://phabricator.endlessm.com/T33325

### DIFF
--- a/js/ui/payg.js
+++ b/js/ui/payg.js
@@ -37,10 +37,6 @@ const SPINNER_ANIMATION_DELAY_MSECS = 1000;
 const SPINNER_ANIMATION_TIME_MSECS = 300;
 var SPINNER_ICON_SIZE_PIXELS = 16;
 
-const NOTIFICATION_TITLE_TEXT = _('Pay As You Go');
-const NOTIFICATION_EARLY_CODE_ENTRY_TEXT = _('Enter a 14-digit keycode to extend the time before your credit expires.');
-const NOTIFICATION_DETAILED_FORMAT_STRING = _('Subscription expires in %s.');
-
 var UnlockStatus = {
     NOT_VERIFYING: 0,
     VERIFYING: 1,
@@ -740,14 +736,18 @@ class PaygNotifier extends GObject.Object {
         Main.messageTray.add(source);
 
         // by default, this notification is for early entry of an unlock keycode
-        let messageText = NOTIFICATION_EARLY_CODE_ENTRY_TEXT;
+        let codeLength = Main.paygManager.codeLength;
+        let messageText = Gettext.ngettext(
+            'Enter a new keycode (%s character) to extend the time before your credit expires.',
+            'Enter a new keycode (%s characters) to extend the time before your credit expires.',
+            codeLength).format(codeLength);
         let urgency = MessageTray.Urgency.NORMAL;
         let userInitiated = false;
 
         // in case this is a "only X time left" warning notification
         if (secondsLeft >= 0) {
             const timeLeft = timeToString(secondsLeft);
-            messageText = NOTIFICATION_DETAILED_FORMAT_STRING.format(timeLeft);
+            messageText = _('Subscription expires in %s.').format(timeLeft);
             urgency = MessageTray.Urgency.HIGH;
         } else {
             userInitiated = true;
@@ -755,7 +755,7 @@ class PaygNotifier extends GObject.Object {
 
         this._notification = new ApplyCodeNotification(
             source,
-            NOTIFICATION_TITLE_TEXT,
+            _('Pay As You Go'),
             messageText);
 
         if (userInitiated)
--- a/js/ui/paygUnlockDialog.js
+++ b/js/ui/paygUnlockDialog.js
@@ -29,6 +29,8 @@ const Animation = imports.ui.animation;
 const Main = imports.ui.main;
 const LayoutManager = imports.ui.layout;
 
+const Gettext = imports.gettext;
+
 const MSEC_PER_SEC = 1000;
 
 // The timeout before going back automatically to the lock screen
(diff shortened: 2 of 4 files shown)


jprvita:
This depends on https://github.com/endlessm/eos-payg/pull/100 and https://github.com/endlessm/eos-payg-nonfree/pull/102.

js/ui/payg.js
```diff
@@ -740,7 +744,10 @@ class PaygNotifier extends GObject.Object {
         Main.messageTray.add(source);
 
         // by default, this notification is for early entry of an unlock keycode
-        let messageText = NOTIFICATION_EARLY_CODE_ENTRY_TEXT;
+        let codeLength = Main.paygManager.codeLength
+        let messageText = Gettext.ngettext(NOTIFICATION_EARLY_CODE_ENTRY_TEXT_1,
+                                           NOTIFICATION_EARLY_CODE_ENTRY_TEXT_N,
+                                           codeLength).format(codeLength)
```
pwithnall, line 750: Looks like semicolons might be missing at the end of the lines here.

Also, should the string format placeholder be `%s` or should it be `%u`? I’m not sure how `str.format()` works in JavaScript.

js/ui/payg.js
```diff
@@ -38,7 +38,11 @@ const SPINNER_ANIMATION_TIME_MSECS = 300;
 var SPINNER_ICON_SIZE_PIXELS = 16;
 
 const NOTIFICATION_TITLE_TEXT = _('Pay As You Go');
-const NOTIFICATION_EARLY_CODE_ENTRY_TEXT = _('Enter a 14-digit keycode to extend the time before your credit expires.');
+
+// These will be passed to ngettext to get pluralized translations
+const NOTIFICATION_EARLY_CODE_ENTRY_TEXT_1 = 'Enter a new keycode (%s character) to extend the time before your credit expires.';
+const NOTIFICATION_EARLY_CODE_ENTRY_TEXT_N = 'Enter a new keycode (%s characters) to extend the time before your credit expires.';
```
pwithnall, line 44: I’m not 100% sure, but I suspect these strings should be marked with `gettext_noop()` (in C this is `N_()`) or something like that, so that `xgettext` knows to extract them from the source code into the `.pot` file at build time.

There are two parts to using gettext: build-time extraction of the translatable strings into the `.pot` file (which translators then localise as `$lang.po` files), and runtime lookup of the translated forms of those strings from the translation catalogue. Extraction of the strings uses some basic source code parsing, looking for strings being passed to certain gettext functions, and assuming all of them are to be translated. I don’t think it’s clever enough to fold constants.

Normally the `_()` macro (equivalently the `gettext()` function) handles both parts, but in the case of defining the translatable strings separately from where they’re translated using `ngettext()`, the strings need to be explicitly marked as for extraction into a `.pot` file.

js/ui/paygUnlockDialog.js
```diff
@@ -110,9 +112,12 @@ var PaygUnlockDialog = GObject.registerClass({
         });
         paygEnterCodeBox.add_child(promptBox);
 
+        let codeLength = Main.paygManager.codeLength
```
pwithnall, line 115: Missing semicolon?

js/ui/paygUnlockDialog.js
```diff
@@ -110,9 +112,12 @@ var PaygUnlockDialog = GObject.registerClass({
...
         const promptLabel = new St.Label({
             style_class: 'unlock-dialog-payg-label',
-            text: _('Enter a new 14-digit keycode to unlock your computer:'),
+            text: Gettext.ngettext('Enter a new keycode (%s character) to unlock your computer:',
+                                   'Enter a new keycode (%s characters) to unlock your computer:',
```
pwithnall, line 119: Is `%s` the correct format placeholder?

pwithnall (commented):
I’m not sure about any of my comments, since I haven’t done JS for a long time, and have done localisation in JS even less. Still, they might be worth checking

js/ui/payg.js
jprvita: > Looks like semicolons might be missing at the end of the lines here.
> 

Good catch! It seems they are not required by the language and perhaps a coding-style preference, as the code worked without them. In any case, we should add them here.

> Also, should the string format placeholder be `%s` or should it be `%u`? I’m not sure how `str.format()` works in JavaScript.

I thought the same when I first wrote this, but `%u` gave me a runtime error. Then looking around the file I found other places using `%s` for integers, ex., `Gettext.ngettext('%s second', '%s seconds', seconds).format(Math.floor(seconds));`, so I followed that.

js/ui/paygUnlockDialog.js
jprvita: Yep

js/ui/payg.js
wjt: If you are morbidly curious about JavaScript's relationship to the noble semicolon: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#automatic_semicolon_insertion

js/ui/payg.js
jprvita: That makes sense, thanks for the explanation (I never dealt much with localization before). I skimmed through https://github.com/guillaumepotier/gettext.js/ and did not find anything equivalent to `N_()` or `gettext_noop`. I think the safe approach is to just move the definition of these strings to where they are used, together with NOTIFICATION_TITLE_TEXT and NOTIFICATION_DETAILED_FORMAT_STRING -- while it is nice to have them grouped in the top of the file, we're not gaining much with these macros, as they are only used once.

js/ui/payg.js
pwithnall: :exploding_head: Not sure I can handle living in the future.

js/ui/payg.js
pwithnall: > I think the safe approach is to just move the definition of these strings to where they are used, together with NOTIFICATION_TITLE_TEXT and NOTIFICATION_DETAILED_FORMAT_STRING -- while it is nice to have them grouped in the top of the file, we're not gaining much with these macros, as they are only used once.

Yeah, that seems pragmatic

wjt: I don't think that's the gettext module exposed by GJS. I think it's this one, which is just a thin wrapper around the corresponding functions in glib. https://gitlab.gnome.org/GNOME/gjs/-/blob/master/modules/core/_gettext.js

jprvita:
Addressed previous feedback.

pwithnall (approved):
LGTM

PR #734, opened 2022-08-17
Report metrics for application usage using aggregate timer API

wjt:
This reverts the commit that used the previous "sequence" API and backports the two commits from the `master` branch to use the new timer API.

TODO:

- [x] Land the API that this uses onto the `eos4.0` OS branch, before which this branch will fail to build!

https://phabricator.endlessm.com/T32673


wjt:
retest this please

wjt:
Tested this on an OS build that includes the other bits. It works!

GeorgesStavracas (approved):

PR #735, opened 2022-11-16
T33105

GeorgesStavracas:
image

I accidentally discovered our custom CSS has never been applied since last rebase. I wonder how much this will impact PAYG visuals.

https://phabricator.endlessm.com/T33105


jprvita (approved):
My familiarity with this codebase is still limited, but this seems to make sense.

PR #738, opened 2023-01-23
T34300 eos: Reset dynamic-workspaces key on EOS 4 → 5 transition

pwithnall:
See the comment in the new `gnome-shell-overrides-reset.sh` script for
details.

This commit is specific to EOS. It can be dropped any time after the
first checkpoint after the EOS 5 release. By that point, any users who
have upgrade from 4 → 5 (and triggered the bug this script fixes) should
have logged into EOS 5 at least once, hence run this script, and the
migration bug should be fixed for them.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>
https://phabricator.endlessm.com/T34300


GeorgesStavracas (approved):
LGTM

GeorgesStavracas:
Hmm, actually... I think https://github.com/endlessm/gnome-shell/blob/debian-master/debian/gnome-shell.install#L1 needs to be updated too?

pwithnall:
> Hmm, actually... I think https://github.com/endlessm/gnome-shell/blob/debian-master/debian/gnome-shell.install#L1 needs to be updated too?

good catch!!!

PR #741, opened 2023-02-17
welcomeDialog: Record whether the tour is launched or skipped

wjt:
Anecdotally, a sample of 1 user reported having reflexively skipped the
tour. It would be interesting to know what proportion of users takes the
tour.

The event will be recorded together with the current OS version (like
all others) and the image ID, which tells us the originally-installed OS
version, so a simple boolean is enough for us to measure what proportion
of new/upgrading/all users took or skipped the tour.

https://phabricator.endlessm.com/T34473

### DIFF
--- a/js/ui/welcomeDialog.js
+++ b/js/ui/welcomeDialog.js
@@ -1,7 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 /* exported WelcomeDialog */
 
-const { Clutter, GObject, Shell, St } = imports.gi;
+const { Clutter, EosMetrics, GLib, GObject, Shell, St } = imports.gi;
 
 const Config = imports.misc.config;
 const Dialog = imports.ui.dialog;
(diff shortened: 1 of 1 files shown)


wjt:
I've tested that this event makes it into Azafea, currently the `unknown_singular_event_v3` table:

```
azafea=> select * from unknown_singular_event_v3 where id = 14536;
  id   | os_version |               event_id               | payload_data |          occured_at           | channel_id 
-------+------------+--------------------------------------+--------------+-------------------------------+------------
 14536 | 5.0.2      | 140643be-fe47-4b4b-985b-d16f8f3973a9 | \x01006200   | 2023-03-02 15:54:23.347975+00 |          1
(1 row)
```

`\x01006200` is the correct serialization of the payload:

```
In [3]: from gi.repository import GLib

In [4]: GLib.Variant.new_from_bytes(GLib.VariantType('mv'), GLib.Bytes.new(b'\x01\x00\x62\x00'), True)
Out[4]: GLib.Variant('mv', <true>)
```

jprvita (approved):

PR #747, opened 2023-09-07
extensions: Allow desktop ID to be ‘endless’ or ‘Endless’

pwithnall:
Historically we’ve defined it as ‘endless’ (e.g.
`XDG_CURRENT_DESKTOP=endless:GNOME`). We’re now looking to standardise the desktop ID, and for that it would be better if it were capitalised like other desktop IDs.

See: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/73


pwithnall:
Fast-forwarded `eos5.0` to include this commit too.

PR #752, opened 2023-12-20
T35065 handle app renaming

dylanmccall:
At the moment, this pull request is bringing back the following patches, removing references to `X-Endless-Alias`:

- ~563b2f1 shell-app-system: Add support for X-Endless-Alias key in desktop files~ (removed, squashed with the next patch)
- 407f304 shell-app-system: obey X-Flatpak-RenamedFrom
- 4f4fc58 appDisplay: compare possibly-changed apps more thoroughly
- b04b91d shell-app-system: extract deep comparison for GDesktopAppInfo
- 26a98c3 appFavorites: respect X-Endless-Alias and X-Flatpak-RenamedFrom

It is a draft while I look around for an opportunity to better align with upstream.

https://phabricator.endlessm.com/T35065

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1062,13 +1062,33 @@ var BaseAppView = GObject.registerClass({
 
     _redisplay() {
         let oldApps = this._orderedItems.slice();
-        let oldAppIds = oldApps.map(icon => icon.id);
-
         let newApps = this._loadApps().sort(this._compareItems.bind(this));
-        let newAppIds = newApps.map(icon => icon.id);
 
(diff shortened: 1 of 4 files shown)


wjt:
I proposed https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/593 upstream 5 years ago. Florian has an [alternative approach using libappstream](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/41) which has also stalled. Personally I don't think invoking another library and reading massive XML files from the compositor is a good idea but what do I know...

manuq (approved):

PR #753, opened 2024-01-19
T35064 add payg patches

dylanmccall:
https://phabricator.endlessm.com/T35064

-----

I kept my bigger changes on top of the old patches in separate `fixup!` commits to make reviewing a bit easier - so we don't mix up old mistakes and new mistakes :b Please autosquash before merging.

### DIFF
--- a/src/shell-util.c
+++ b/src/shell-util.c
@@ -852,3 +852,27 @@ shell_util_get_translated_folder_name (const char *name)
 {
   return shell_app_cache_translate_folder (shell_app_cache_get_default (), name);
 }
+
+/**
+ * shell_util_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses `CLOCK_BOOTTIME`, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ */
+gint64
+shell_util_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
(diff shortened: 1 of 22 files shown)


dylanmccall:
This is definitely causing some breakage around the screen shield (at least in the thoroughly abused VM I was using to test everything together). So I'll switch this back to Draft while I poke at what's going on here.

wjt (approved):
Changes LGTM. I will test this locally to see how it looks.

src/shell-util.c
```diff
@@ -852,3 +852,27 @@ shell_util_get_translated_folder_name (const char *name)
...
 }
+
+/**
+ * shell_util_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses `CLOCK_BOOTTIME`, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
```
manuq, line 863: I was about to ask how much unspecified the starting point is. But maybe is enough to reference Linux CLOCK_BOOTTIME which should have that explanation.

manuq:
I took this opportunity to get familiar with the PAYG patch on top of GNOME Shell. All looks fine to me!

manuq:
> Please autosquash before merging
Ouch.

dylanmccall:
> To my eye the label is slightly too high up

Oh, I know why that is! Your account ID has a newline at the end. Are you using paygd's fallback behaviour with `/usr/local/share/eos-payg/account-id`? It expects that file to end without a newline :)

wjt:
Oh, yeah, I created that file with `echo | sudo tee`, and `echo` adds a newline. I can retest yesterday.

PR #754, opened 2024-02-01
appDisplay: Handle an error moving nonexistent app icons

dylanmccall:
With the added set of application aliases, it is possible that BaseAppGridViewLayout._redisplay will call _moveItem for an application which is no longer present in the icon grid. This requires an unusual series of events, so for simplicity, we will simply log the exception and carry on.

https://phabricator.endlessm.com/T34672


js/ui/appDisplay.js
```diff
@@ -1102,7 +1102,11 @@ var BaseAppView = GObject.registerClass({
             if (addedApps.includes(icon))
                 this._addItem(icon, page, position);
             else if (page !== -1 && position !== -1)
-                this._moveItem(icon, page, position);
+                try {
+                    this._moveItem(icon, page, position);
+                } catch (error) {
+                    log(`Error moving app icon: ${error.message}`);
+                }
```
wjt, line 1109: I'm sure this will suppress the error but it doesn't seem like the right way to handle this. If you have two apps in your app grid layout, and one becomes an alias for the other, the old name should be removed from the layout, I think.

PR #755, opened 2024-03-04
Remove eos-watermark from 'endless' session mode

wjt:
We are removing this from the OS.

This is very lightly used, not enough to justify updating it for GNOME
43 compatibility.

If we wish to bring this functionality back in future, we should instead
package https://pagure.io/background-logo-extension, which is actually
the ancestral source of the eos-watermark-extension code and which is
actively maintained by a core Shell developer for each Fedora release.

https://phabricator.endlessm.com/T35077

### DIFF
--- a/debian/endless-session-modes/endless.json
+++ b/debian/endless-session-modes/endless.json
@@ -2,7 +2,6 @@
     "parentMode": "user",
     "enabledExtensions": [
         "eos-desktop@endlessm.com",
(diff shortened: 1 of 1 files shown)


wjt:
As an aside, we should move this file to `eos-theme` or somewhere that we don't have to keep mangling into the packaging.

manuq (approved):

PR #756, opened 2024-04-10
fixup! welcomeDialog: Welcome users to Endless OS, not GNOME

manuq:


js/ui/welcomeDialog.js
```diff
@@ -32,7 +32,8 @@ class WelcomeDialog extends ModalDialog.ModalDialog {
     }
 
     _buildLayout() {
-        const title = _('Welcome to Endless OS 5');
+        const [majorVersion] = GLib.get_os_info('PRETTY_NAME').split('.');
```
wjt, line 35: This does make an assumption about the format of PRETTY_NAME but I think it's fine.

wjt (approved):

wjt:
Fast-forwarded `eos6.0` to the same commit.

PR #757, opened 2024-04-23
keyboard: Ensure a Latin layout is always available

wjt:
In past versions of Endless OS, we added the concept of a “password mode” to
GNOME Shell, which is triggered either within GNOME Shell's own password
fields, or from applications by setting a D-Bus property on the shell. In this
mode, Shell would check whether any configured input source could type the
English letters a through z, and if not, append the US English keyboard layout
to the list of available sources. The intent is to ensure, for example, that it
is possible to type Wi-Fi passwords, which typically use only English ASCII
letters even in regions where the local language uses some other script, even
if the user's keyboard layout is (say) Thai.

This approach requires coöperation from applications (which we previously
patched GNOME Initial Setup to do, for example) and does not solve the related
problem that URLs and email addresses are also typically written in English
ASCII letters throughout the world.

For Endless OS 6, we had removed this feature entirely, but I am concerned that
existing users may be relying on this feature to enter their account password
using Latin characters at GDM while only having a non-Latin input source
configured.

So, reintroduce the logic, but apply it at all times, rather than when a
password is being entered. This will, for example, allow entering URLs on a
system configured with only a Thai layout.

The logic for detecting whether a fallback method is needed is a little
simplistic and does not handle some cases correctly. Some IBus methods, such as
libzhuyin for Chinese, provide a built-in passthrough mode to enter English
text; but this logic will add a fallback US English method if libzhuyin is the
only configured method. This seems acceptable as a stop-gap.

There is ongoing work in GNOME to solve this problem more generally, including
handling the case of some IBus methods providing their own passthrough mode.
See https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/213 and
linked issues & merge requests.

https://phabricator.endlessm.com/T31812

---

I structured this PR as a cherry-pick (fixing conflicts) of the old code, followed by deleting bits of it, to make it easier to review the changes compared both to the current eos6.0 code, and to what we shipped in previous Endless OSes.

TODO:

- [x] Move “Make eosmetrics dependency optional” to a separate PR (or abandon it)
- [x] Remove more dead code
- [x] Remove `log()` messages used in development

### DIFF
--- a/js/misc/keyboardManager.js
+++ b/js/misc/keyboardManager.js
@@ -1,7 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 /* exported getKeyboardManager, holdKeyboard, releaseKeyboard */
 
-const { GLib, GnomeDesktop } = imports.gi;
+const { GLib, GnomeDesktop, Shell } = imports.gi;
 
 const Main = imports.ui.main;
 
@@ -155,6 +155,16 @@ var KeyboardManager = class {
     _buildOptionsString() {
         let options = this._xkbOptions.join(',');
         return options;
+    }
+
+    isLatinLayout(id) {
+        const info = this._layoutInfos[id];
+        if (!info)
+            return false;
+        const options = this._buildOptionsString();
+        const [, variants] = this._buildGroupStrings(info.group);
+        const [, , , layout, _variant] = this._xkbInfo.get_layout_info(id);
+        return !Shell.util_needs_secondary_layout(layout, variants, options);
     }
 
     get currentLayout() {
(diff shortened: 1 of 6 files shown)


wjt:
I believe this is ready for review.

I am open to suggestions as to whether the two patches should be squashed together (reducing churn) or left separate (improving the ability to review old vs new code).

dylanmccall:
This looks good to me. I think I would prefer squashing these: I found it was easier to understand what this was doing when I squashed the commits and compared new & old in my head. But it would be very useful to mention the newest version of the original patch (e2fbf6f8908f6ed6aa63c3b34d6c943927de116a) in the commit message.

dylanmccall (approved):

wjt:
Squashed and adjusted the commit message to cite that commit ID.

PR #758, opened 2024-04-24
Revert “eos: Reset dynamic-workspaces key on EOS 4 → 5 transition”

wjt:
This reverts commit d1d176b995ee3113065a7accc0830e5d8f68eab0. Per its description:

> It can be dropped any time after the
> first checkpoint after the EOS 5 release. By that point, any users who
> have upgrade from 4 → 5 (and triggered the bug this script fixes) should
> have logged into EOS 5 at least once, hence run this script, and the
> migration bug should be fixed for them.

At the time that I rebased this commit it was not clear whether we would have a checkpoint between EOS 5 and 6, but now we know that we will, we can remove this migration.

Actually the rationale is not completely accurate. On a multi-user system there is no guarantee that all users have logged into EOS 5 at least once; and even on a single-user system it's possible to leave the computer sitting at the login screen, without logging in, and have automatic updates run their course. But practically speaking it's true enough.

https://phabricator.endlessm.com/T34300


dylanmccall:
This looks reasonable to me. Indeed, it's unlikely that someone will run into this issue, and if they do, well, they get static workspaces. There are worse fates.

wjt:
Thanks, also pushed to eos6.0

wjt:
I pushed a quick packaging fixup to f51a68e0f.

PR #759, opened 2024-07-20
Backport two upstream fixes for startup animation

wjt:
The [first patch](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/7a5f1e5c9e96b04a41407d1ae9d9f242b31ed1cf) is part of upstream 45.0 and newer.

The second was [accepted upstream](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3422) and will be in 47.beta.

https://phabricator.endlessm.com/T34453

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -658,7 +658,7 @@ var LayoutManager = GObject.registerClass({
                 if (this.primaryMonitor) {
                     this._systemBackground.show();
                     global.stage.show();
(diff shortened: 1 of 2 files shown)


manuq (approved):
Yes, let's hope these backports improve the overlay situation.

wjt:
They don't make any difference in our case but it's a step on the way.

PR #760, opened 2024-07-22
Make startup animation use async functions throughout

wjt:
Requires corresponding changes in eos-desktop-extension:

- https://github.com/endlessm/eos-desktop-extension/pull/99

----

During session startup, we have the following chain:

    LayoutManager._prepareStartupAnimation()
    `- LayoutManager._startupAnimation()
       `- LayoutManager._startupAnimationComplete()       (a)
       | or
       `- LayoutManager._startupAnimationGreeter()        (b)
       | or
       `- LayoutManager._startupAnimationSession()        (c)
          `- Overview.runStartupAnimation()               (d)
             `- OverviewActor.runStartupAnimation()       (e)
                `- ControlsManager.runStartupAnimation()  (f)

Branch (b) calls LayoutManager._startupAnimationComplete() once the
animation is complete. Branch (c) does the same, except that
ControlsManager.runStartupAnimation() is an async function that awaits
LayoutManager.ensureAllocation().

LayoutManager._prepareStartupAnimation() is an `async` function, and its
caller catches & logs any error it raises.

Previously, ControlsManager.runStartupAnimation() – (f) in the above
diagram – was declared `async`, because it uses `await` internally, but
also accepted a callback, which was called on successful completion of
the function and the animation it triggers.

If an exception is raised during execution of the function, its callback
would never be called, and because the promise it implicitly returns is
discarded, the exception would never be logged either. And, stepping a
few levels up the call stack, the callback that
LayoutManager._startupAnimationSession() (c) provided would never be
called, and so LayoutManager._startupAnimationComplete() would never be
called, meaning the cover pane that prevents input would never be
removed and the session would be unusable.

Remove the callback parameter from ControlsManager.runStartupAnimation()
(f), and instead make it resolve in the case where the callback would
previously have been called.

Remove the callback parameter from OverviewActor.runStartupAnimation()
(e) – it is just a wrapper around ControlsManager.runStartupAnimation().

Adjust Overview.runStartupAnimation() (d) accordingly: rather than
passing a callback to OverviewActor.runStartupAnimation(), await its
return and then proceed. There is a slight behaviour change here:
previously, in the branch where this._syncGrab() is false, the callback
would be called before calling this.hide(), whereas now this.hide() is
called before the async function returns.

Adjust LayoutManager._startupAnimationSession() and its siblings to be
async, rather than each calling _startupAnimationComplete() directly.
Finally, in _prepareStartupAnimation()  await _startupAnimation() either
succeeding or failing, and in either case call
_startupAnimationComplete(), removing the cover pane that prevents all
input.

https://phabricator.endlessm.com/T34453

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -727,39 +727,46 @@ var LayoutManager = GObject.registerClass({
 
         this.emit('startup-prepared');
 
-        this._startupAnimation();
-    }
-
-    _startupAnimation() {
-        if (Meta.is_restart())
+        try {
+            await this._startupAnimation();
+        } finally {
             this._startupAnimationComplete();
-        else if (Main.sessionMode.isGreeter)
-            this._startupAnimationGreeter();
-        else
-            this._startupAnimationSession();
+        }
     }
 
-    _startupAnimationGreeter() {
-        this.panelBox.ease({
-            translation_y: 0,
-            duration: STARTUP_ANIMATION_TIME,
-            mode: Clutter.AnimationMode.EASE_OUT_QUAD,
-            onStopped: () => this._startupAnimationComplete(),
+    async _startupAnimation() {
+        if (Meta.is_restart())
+            return;
+        else if (Main.sessionMode.isGreeter)
+            await this._startupAnimationGreeter();
+        else
+            await this._startupAnimationSession();
+    }
+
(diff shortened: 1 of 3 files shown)


wjt:
Best reviewed in "hide whitespace" mode

image

PR #761, opened 2024-08-14
fixup! payg: Add panel indicator for PAYG account details and actions

wjt:
The string “Enter unlock code…” was not marked for translation since
d50cd8dc60f400882db9ce425da13003b0561965 which reworked this interface
for the newer quick settings menu.

https://phabricator.endlessm.com/T35611


dylanmccall (approved):
That looks good to me :)

PR #745, opened 2023-04-07
payg: Create PaygAddCreditDialog

jprvita:
PaygAddCreditDialog is a modal dialog for early code entry, to replace
the notification widget currently in use.

https://phabricator.endlessm.com/T31748

PR #763, opened 2025-06-26
Improve early PAYG code entry UI with a new dialog

starnight:
Currently, PAYG credit input dialog in the notification closes too quickly if the cursor is not within the window area.

[Screencast_from_2024-06-05_21-02-11.webm](https://github.com/user-attachments/assets/28ce33a9-9559-4fea-a895-039ca650e299)

So, move the input textbox from the notification to a new PaygAddCreditDialog, which is a window and will not close automatically.  Thanks to @jprvita's effort.  We can extracted out the PaygAddCreditDialog from PR #745.

https://phabricator.endlessm.com/T31748


js/ui/payg.js
```diff
@@ -395,116 +484,214 @@ var PaygUnlockWidget = GObject.registerClass({
...
+            action: () => this.close(),
+        });
 
-    _createCodeEntry() {
-        const codeEntry = new PaygUnlockCodeEntry({
-            style_class: 'notification-payg-entry',
-            x_expand: true,
-            can_focus: true,
+        this._applyButton = this.addButton({
+            label: _('Add Credit'),
```
wjt, line 500: I would give this `key: Clutter.KEY_Enter` (or if there's a way to make it the default action for hitting Enter in the field, that instead)

wjt (commented):
On Slack I gave feedback that I think the field should disappear once you enter a code successfully.

I now see that it's at least theoretically possible to enter multiple codes, one after the other. So that's why the field is cleared but is still shown.

Personally I think it's a weird flow, and that it's unusual to have multiple codes to enter. I would hide the Apply button and _codeEntryLayout once a code is entered successfully, leaving only the Close button. Next time the dialog is opened they should reapepar

js/ui/payg.js
starnight: Good suggestion.  However, **_applyButton** type of **PaygUnlockCodeEntry** has already handled it.  And, **_applyButton**'s `activate` signal has been bound with `_apply()`.

starnight:
> I suppose it's actually not catastrophic if there is a bug in this logic because the "your credit has actually expired" overlay hasn't changed (?) but we should also test that.

Do you mean will the system lock up when credit becomes expired?  If yes, then, yes!  PAYG locks system when credit becomes expired.

starnight:
> No, I mean, once that happens, can you unlock it with a code? Both during the session and after rebooting?

Yes, of course.  It unlocks with a correct code both during the session and after rebooting as usual.

